PR #36, opened 2025-10-30
fix: patch image typing in new SDK

cpacker:


src/cli/App.tsx
```diff
@@ -842,6 +842,7 @@ export default function App({
       // Start the conversation loop
       await processConversation([
         {
+          type: "message",
```
cpacker, line 845: @cpfiffer idk if this part is required

PR #141, opened 2025-12-02
feat: Add skill creation command

kl2806:
Implements a `/skill` command that optionally has a instruction. The '/skill` command adds system reminders prompts the model to construct a skill based on the skill creation tool.

### DIFF
--- a/src/headless.ts
+++ b/src/headless.ts
@@ -213,21 +213,27 @@ export async function handleHeadlessCommand(
   // Re-discover skills and update the skills memory block
   // This ensures new skills added after agent creation are available
   try {
-    const { discoverSkills, formatSkillsForMemory, SKILLS_DIR } = await import("./agent/skills");
+    const { discoverSkills, formatSkillsForMemory, SKILLS_DIR } = await import(
+      "./agent/skills"
+    );
     const { join } = await import("node:path");
-    
-    const resolvedSkillsDirectory = skillsDirectory || join(process.cwd(), SKILLS_DIR);
+
+    const resolvedSkillsDirectory =
+      skillsDirectory || join(process.cwd(), SKILLS_DIR);
     const { skills, errors } = await discoverSkills(resolvedSkillsDirectory);
-    
+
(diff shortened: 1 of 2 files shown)


src/agent/prompts/skill_creator_mode.md
```diff
@@ -0,0 +1,125 @@
...
+  - Any `assets/` (templates, fonts, icons, starter projects, etc.)
+
+Validate this design with the user before you start writing files. If something is ambiguous or high‑impact, ask a brief follow‑up question using `AskUserQuestion`.
+
+## 5. Create or update the skill files
+
+Once the design is agreed upon:
+
+1. Determine the target directory for the skill (default assumption):
+   - Use a local `.skills/<skill-id>/` directory in the current project root, unless the user specifies a different path.
```
devanshrj: `skills` block also mentions the skills directory, we should probably use [it here](https://github.com/letta-ai/letta-code/blob/edd173cf34dbc54d5b35972dc8f9d277a1ec44ad/src/agent/context.ts#L80)?

src/agent/prompts/skill_creator_mode.md
```diff
@@ -0,0 +1,125 @@
...
+
+Before you start proposing a concrete skill design, you MUST ask a small bundle of **high‑value upfront questions** using the `AskUserQuestion` tool.
+
+Keep the initial question set small (3–6 questions) and focused. Examples:
+
+1. Purpose and scope:
+   - “What is the main purpose of this skill?”
+   - “Is this skill meant for a specific project or to be reused across many projects?”
+2. Example usage:
+   - “Can you give 2–3 example requests a user might say that should trigger this skill?”
```
devanshrj: the agent can probably infer this through the conversation history?

devanshrj (approved):
Looks solid!! I have one non-blocking concern about the model asking too many questions and being annoying? It makes sense for Claude Code because skill creation is stateless, but I wonder if we can reduce friction here?

PR #149, opened 2025-12-03
feat: Add `memory_apply_patch` for non-Anthropic models

kl2806:
For non-Anthropic models, use `memory_apply_patch` instead of the `memory` tool

### DIFF
--- a/src/agent/create.ts
+++ b/src/agent/create.ts
@@ -105,12 +105,15 @@ export async function createAgent(
   if (toolNames.includes("memory_apply_patch")) {
     try {
       const resp = await client.tools.list({ name: "memory_apply_patch" });
-      const hasMemoryApplyPatch = Array.isArray(resp.items) && resp.items.length > 0;
+      const hasMemoryApplyPatch =
+        Array.isArray(resp.items) && resp.items.length > 0;
       if (!hasMemoryApplyPatch) {
         console.warn(
           "memory_apply_patch tool not found on server; falling back to 'memory' tool",
         );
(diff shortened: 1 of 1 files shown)


src/tools/toolset.ts
```diff
@@ -149,6 +149,65 @@ export async function forceToolsetSwitch(
...
+      include: ["agent.tools"],
+    });
+
+    const currentTools = agent.tools || [];
+    const mapByName = new Map(currentTools.map((t) => [t.name, t.id]));
+
+    // Determine which memory tool we want based on toolset
+    const desiredMemoryTool =
+      toolsetName === "default" ? "memory" : "memory_apply_patch";
+    const otherMemoryTool =
```
devanshrj, line 166: is this to support model swap for an existing agent?

kl2806: yes, so if a existing agent has `memory` attached, i think here we need to guard against attaching both memory tools

src/agent/create.ts
```diff
@@ -89,14 +89,44 @@ export async function createAgent(
     getServerToolName(name),
   );
 
+  const baseMemoryTool = modelHandle.startsWith("anthropic/")
```
cpacker, line 92: should be invert this? eg `modelHandle.startsWith("openai/gpt-5")`?

kl2806: Yes let's revert this for non OAI #151 
@devanshrj reporting instances of Deepseek models struggling with `memory_apply_patch`

PR #153, opened 2025-12-04
fix: Enable `--system` flag for headless mode

kl2806:
Headless mode was previously not using `--system` so it would use the default prompt even with `--system letta-codex`. Enables running our Letta Code parity Codex experiment.


devanshrj (approved):
:shipit:

PR #181, opened 2025-12-11
feat: Exponential backoff for `llm_api_error`

kl2806:


src/headless.ts
```diff
@@ -789,6 +796,39 @@ export async function handleHeadlessCommand(
...
+                reason: "llm_api_error",
+                attempt,
+                max_attempts: LLM_API_ERROR_MAX_RETRIES,
+                delay_ms: delayMs,
+                run_id: lastRunId,
+              }),
+            );
+          } else {
+            const delaySeconds = Math.round(delayMs / 1000);
+            console.error(
```
cpacker, line 821: @christinatong01 we probably want to make a follow-up PR to this to instead of using `console.error`, use the proper error lines thing so it shows up clean in the transcript?

PR #217, opened 2025-12-15
feat: cache init tool upserts

AriWebb:


src/tools/manager.ts
```diff
@@ -586,6 +587,58 @@ export async function upsertToolsToServer(client: Letta): Promise<void> {
...
+ * Compute a hash of all currently loaded tools for cache invalidation.
+ * Includes tool names and schemas to detect any changes.
+ */
+export function computeToolsHash(): string {
+  const toolData = Array.from(toolRegistry.entries())
+    .sort(([a], [b]) => a.localeCompare(b)) // deterministic order
+    .map(([name, tool]) => ({
+      name,
+      serverName: getServerToolName(name),
+      schema: tool.schema,
```
sarahwooders, line 600: I guess we dont need to include the source since the source is only needed on the client? I think this is fine for now but we might want to add a note for the future

AriWebb: sg

PR #219, opened 2025-12-16
feat: update Skill tool to support load/unload commands

cpacker:
Refactor Skill tool to support explicit load/unload and make skill blocks read-only.

**New Schema:**
- `command: 'load' | 'unload'`
- `skills: string[]` (list of skill IDs)

**Examples:**
```
  skill(command='load', skills=['data-analysis'])
  skill(command='unload', skills=['data-analysis'])
  skill(command='load', skills=['web-scraper', 'pdf'])  // batch
```

**Read-Only Blocks:**
- `skills` and `loaded_skills` blocks are now created with `read_only: true`
- Agent cannot modify these blocks via memory tools (blocked by server)
- Skill tool can still update via API (bypasses agent tool executor)

**Benefits:**
- Explicit unload via tool (no manual memory editing)
- Batch operations supported
- Cleaner skill boundary handling on unload
- No risk of agent corrupting skill blocks via memory tools
- Symmetric API: load and unload are both first-class operations

**Updated Prompts:**
- All system prompts now instruct agent to use Skill tool for unloading
- Removed references to using memory tools on `loaded_skills` block

🐱 Generated with [Letta Code](https://letta.com)

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,10 +1,11 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "@letta-ai/letta-code",
       "dependencies": {
-        "@letta-ai/letta-client": "^1.3.3",
+        "@letta-ai/letta-client": "^1.4.0",
         "glob": "^13.0.0",
         "ink-link": "^5.0.0",
         "open": "^10.2.0",
@@ -36,7 +37,7 @@
 
     "@isaacs/brace-expansion": ["@isaacs/brace-expansion@5.0.0", "", { "dependencies": { "@isaacs/balanced-match": "^4.0.1" } }, "sha512-ZT55BDLV0yv0RBm2czMiZ+SqCGO7AvmOM3G/w2xhVPH+te0aKgFjmBvGlL1dH+ql2tgGO3MVrbb3jCKyvpgnxA=="],
 
-    "@letta-ai/letta-client": ["@letta-ai/letta-client@1.3.3", "", {}, "sha512-1pSmkmeuXAN9Lq8PXDO2YSLK0q6O39zh2BHlulVhtc8P3aNxAJF7XBteiAhL2hzb49wF2zv0mi2Uikr67CTUCw=="],
+    "@letta-ai/letta-client": ["@letta-ai/letta-client@1.4.0", "", {}, "sha512-7n59bRxjlcoRsDvE/jNDO1otqdF/m//DwGLKkx3uuf6wHwqt3qzJ7/zFVASl1by8oWsiEYUvJnQ634uA1+6q/g=="],
 
     "@types/bun": ["@types/bun@1.3.1", "", { "dependencies": { "bun-types": "1.3.1" } }, "sha512-4jNMk2/K9YJtfqwoAa28c8wK+T7nvJFOjxI4h/7sORWcypRNxBpr+TPNaCfVWq70tLCJsqoFwcf0oI0JU/fvMQ=="],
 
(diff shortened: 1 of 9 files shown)


src/tools/impl/Skill.ts
```diff
@@ -44,16 +99,115 @@ function extractSkillsDir(skillsBlockValue: string): string | null {
...
+      // If bundled fallback also fails, rethrow the original error
+      throw primaryError;
+    }
+  }
+}
+
+/**
+ * Get skills directory, trying multiple sources
+ */
+async function getResolvedSkillsDir(
```
devanshrj, line 130: `index.ts` and `headless.ts` have some skill directory resolution logic too, might be worth checking for redundancy?

devanshrj (approved):
non-blocking comments about redundant skill directory resolution logic, otherwise :shipit:

cpfiffer:
I'm going to update the docs assuming we're doing this

src/tools/impl/Skill.ts
cpacker: I raise you my SLOP CARD
image


```
● Review of the Review ✅

  The reviewer's feedback is valid and correct. There is indeed redundant skill directory resolution logic
  across the three files. Here's my analysis:

  Current State

  1. Skill.ts (lines 127-156) - Has a comprehensive async function:
    async function getResolvedSkillsDir(client, agentId): Promise<string> {
      // 1. Check context
      let skillsDir = getSkillsDirectory();

      // 2. Extract from skills block if not in context
      if (!skillsDir) {
        const skillsBlock = await client.agents.blocks.retrieve("skills", { agent_id: agentId });
        skillsDir = extractSkillsDir(skillsBlock.value);
      }

      // 3. Fall back to default
      if (!skillsDir) {
        skillsDir = join(process.cwd(), SKILLS_DIR);
      }

      return skillsDir;
    }

  2. index.ts (line 731-732) - Simple inline fallback:
    const resolvedSkillsDirectory = skillsDirectory || join(process.cwd(), SKILLS_DIR);

  3. headless.ts (lines 269-270) - Same simple inline fallback:
    const resolvedSkillsDirectory = skillsDirectory || join(process.cwd(), SKILLS_DIR);

  The Issue

  1. Duplication: The fallback logic || join(process.cwd(), SKILLS_DIR) is duplicated in three places
  2. Inconsistency: Skill.ts has sophisticated multi-source resolution logic that isn't used elsewhere
  3. Missed opportunity: Both index.ts and headless.ts call setAgentContext(agent.id, client, 
     skillsDirectory) just before the resolution, but then don't use getSkillsDirectory() from context

  Recommendations

  The redundancy could be reduced by:

  1. Option A - Centralize in context module: Export the getResolvedSkillsDir() function from Skill.ts or
     move it to skills.ts as a utility, then import and use it in all three places

  2. Option B - Simplify and standardize: Since both index.ts and headless.ts already have the CLI
     skillsDirectory value AND they store it in context via setAgentContext(), they could just use:
     ```typescript
     const resolvedSkillsDirectory = skillsDirectory || join(process.cwd(), SKILLS_DIR);
     ```
     This is fine, but then Skill.ts should use the same pattern instead of the complex logic

  3. Option C - Use context consistently: If setAgentContext() is always called before skill discovery, then
     all three places could simply call getSkillsDirectory() || join(process.cwd(), SKILLS_DIR) without
     needing the async block fetching

  Severity

  This is correctly flagged as non-blocking because:
  • ✅ The code works correctly
  • ✅ No bugs or functional issues
  • ⚠️ Minor maintenance burden (3x the same fallback logic)
  • ⚠️ Slight inconsistency in approach

  It's good tech debt to address when convenient, but doesn't need to block the PR.
```

devanshrj: > It's good tech debt to address _when convenient_

Shocked_Pikachu

PR #245, opened 2025-12-16
fix: skip empty reasoning blocks in static commit

kianjones9:
## Summary
- Fixes stray "✻ Thinking…" indicators appearing after responses complete
- Filters out empty reasoning blocks before committing them to static items

## Root Cause
Some providers (especially Gemini) emit reasoning blocks that get marked as `phase: "finished"` but contain empty or whitespace-only text. These empty blocks were being committed to the static transcript, causing the "Thinking..." header to appear without any content.

## Test Plan
- [ ] Test with Gemini models that previously showed the bug
- [ ] Verify thinking indicator disappears after response completes
- [ ] Ensure valid reasoning blocks (with content) still display correctly

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -533,6 +533,11 @@ export default function App({
         continue;
       }
       if ("phase" in ln && ln.phase === "finished") {
+        // Skip empty reasoning blocks to prevent stray "Thinking..." indicators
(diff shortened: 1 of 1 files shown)


kianjones9:
fixing on backend instead

PR #248, opened 2025-12-17
refactor: use system secrets when possible

ThatOneCalculator:
- Uses the `Bun.secrets` API in order to store the API key and refresh key in the system secrets manager (Keychain on macOS, `libsecret` on Linux), with fallback in case the user is running with Node
- Migrates existing tokens to secrets if available
- Uses the XDG standard config directory for user settings `$HOME/.config/letta` rather than `$HOME/.letta`
- Built `letta.js` runs with polyglot shebang that prioritizes bun, then node as a fallback, and can be directly ran with either

PR was mostly written by agent, but a good portion was written by hand and was all hand-tested.

image


src/tests/secrets.test.ts
```diff
@@ -0,0 +1,188 @@
...
+      await deleteSecureTokens();
+    }
+  });
+
+  test("isKeychainAvailable works", async () => {
+    const available = await isKeychainAvailable();
+    expect(typeof available).toBe("boolean");
+  });
+
+  test.skipIf(!keychainSupported)(
```
4shub: I noticed all the tests skip most of these tests, could you configure a github action that can actually test these?

4shub (changes requested):

src/tests/secrets.test.ts
ThatOneCalculator: ~~I tried, but GitHub actions is super inconsistent with testing keyring stuff since you need to have a dbus session, which needs either an x11 or wayland session to be running. Running those tests on my computer in a graphical session results with all passes~~

ThatOneCalculator: I'm going to do a test push with a third party action that claims to let GNOME keyring work headlessly, and I'll revert it if it doesn't work.

ThatOneCalculator: @4shub turns out I was wrong. They work in CI now!

```
(skip) Link/Unlink Tools > (unnamed)
(skip) Link/Unlink Tools > linkToolsToAgent attaches all Letta Code tools
(skip) Link/Unlink Tools > linkToolsToAgent adds approval rules for all tools
(skip) Link/Unlink Tools > linkToolsToAgent returns success when tools already attached
(skip) Link/Unlink Tools > unlinkToolsFromAgent removes all Letta Code tools
(skip) Link/Unlink Tools > unlinkToolsFromAgent removes approval rules
(skip) Link/Unlink Tools > unlinkToolsFromAgent preserves non-Letta-Code tools
(skip) Link/Unlink Tools > unlinkToolsFromAgent preserves non-approval tool_rules
(skip) Link/Unlink Tools > (unnamed)
 462 pass
 9 skip
 0 fail
 913 expect() calls
```

ThatOneCalculator:
Re-requesting review from @4shub 
Requesting review from @cpfiffer (haiii)

4shub (approved):

cpacker:
possible solution: https://github.com/letta-ai/letta-code/pull/435

PR #410, opened 2025-12-29
fix: fix typos

immanuel-peter:


cpacker (changes requested):
Needs `bun run lint` then should be good to go

PR #435, opened 2025-12-31
feat: prefer bun runtime

cpacker:
Restores functionality from #248 (Windows regression was fixed with #434, but temp disabled the Bun secrets functionality)

```
  npm publish → letta.js with #!/usr/bin/env node
                      ↓
                npm install
                      ↓
              postinstall runs
                      ↓
      ┌───────────────┴───────────────┐
      ↓                               ↓
   Windows                     Unix + Bun available
   (no change)                 (swap to polyglot shebang)
```

### DIFF
--- a/scripts/postinstall-patches.js
+++ b/scripts/postinstall-patches.js
@@ -1,7 +1,14 @@
 // Postinstall patcher for vendoring our Ink modifications without patch-package.
 // Copies patched runtime files from ./src/vendor into node_modules.
 
-import { copyFileSync, existsSync, mkdirSync } from "node:fs";
+import { execSync } from "node:child_process";
+import {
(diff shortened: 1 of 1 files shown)


cpacker (commented):
Self-review notes:

**Verified:**
- ✅ Polyglot shebang is correct and won't double-replace on multiple postinstall runs (checks for `#!/usr/bin/env node` specifically)
- ✅ Windows correctly excluded (polyglot breaks npm wrappers)
- ✅ Graceful fallback if Bun not available (`execSync` in try/catch)
- ✅ `debugWarn` silences noise for npm users (fallback is expected behavior)
- ✅ `existsSync` check handles missing `letta.js`

Ready for external review.

PR #448, opened 2026-01-02
feat: add BYOK fallback for built-in subagents

### LINKED ISSUE #424: feat: Allow BYOK (Bring Your Own Key) configuration for subagents
### Summary

Subagents spawned via the `Task` tool should be able to use the user's BYOK (Bring Your Own Key) configuration, specifically the Anthropic OAuth credentials stored in `~/.letta/settings.json`. Currently, there's no clear way to ensure subagents use BYOK instead of Letta's default API.

### Problem Statement

Users who configure BYOK (via `/login anthropic` or similar) expect all their inference calls to use their own API subscription. However, subagents are ephemeral—they spawn, execute, and terminate immediately. This makes it unclear:

1. **Which API key do subagents use?** The parent's BYOK or Letta's default?
2. **How can users explicitly configure subagent API usage?**
3. **Is there billing transparency?** Users may unknowingly consume Letta credits for subagent calls.

### Current State

The `~/.letta/settings.json` contains BYOK configuration:

```json
{
  "anthropicOAuth": {
    "access_token": "sk-ant-oat01-...",
    "refresh_token": "sk-ant-ort01-...",
    "expires_at": 1767084164773,
    "scope": "user:inference user:profile"
  }
}
```

When the parent agent makes direct API calls, BYOK is used. But when spawning subagents:

```typescript
Task({
  subagent_type: "explore",
  prompt: "Find all Python files...",
  model: "haiku"  // Which API key is used here?
})
```

It's unclear if `haiku` uses:
- The user's Anthropic OAuth (BYOK)
- Letta's pooled API
- Some hybrid approach

#### Option 1: Inherit BYOK by Default

Subagents should automatically inherit the parent agent's BYOK configuration:

```typescript
// Internal implementation
const subagentConfig = {
  model: resolvedModel,
  apiKey: parentAgent.byokConfig || lettaDefault
};
```

#### Option 2: Explicit BYOK Parameter

Add a `use_byok` parameter to the Task tool:

```typescript
Task({
  subagent_type: "explore",
  prompt: "...",
  model: "haiku",
  use_byok: true  // Explicitly use user's API key
})
```

#### Option 3: Global Setting

Add a setting in `~/.letta/settings.json`:

```json
{
  "subagentUseBYOK": true,
  "anthropicOAuth": { ... }
}
```

### Use Cases

1. **Cost control**: Users want all inference on their own subscription for predictable billing
2. **Rate limits**: BYOK users have their own rate limits, separate from Letta's pooled limits
3. **Enterprise compliance**: Some organizations require all API calls through their own accounts
4. **Model access**: BYOK users may have access to models not available via Letta's API

#### Transparency

- Display which API is being used when spawning subagents (in verbose mode)
- Show billing attribution in usage reports

#### Documentation

- Document how BYOK interacts with subagents
- Clarify the default behavior

#### Edge Cases

- What if BYOK token is expired? Fall back to Letta API or fail?
- What if the specified model isn't available via BYOK?

### Related Issues

- #423 - Subagent model 'inherit' not recognized (related to model resolution)

### Impact

- **User segment**: All BYOK users who use subagents
- **Benefit**: Cost transparency, billing control, rate limit management

cpacker:
When the parent agent is using a BYOK model (claude-pro-max provider),
subagents will now automatically use the BYOK equivalent of their
recommended model. This ensures consistent billing for users who have
connected their Anthropic Pro/Max subscription.

Closes #424

Generated with [Letta Code](https://letta.com)

### DIFF
--- a/src/agent/subagents/manager.ts
+++ b/src/agent/subagents/manager.ts
@@ -52,6 +52,35 @@ interface ExecutionState {
 }
 
 // ============================================================================
+// BYOK Model Mapping
+// ============================================================================
+
+/**
+ * Mapping from standard Claude model IDs to their BYOK (claude-pro-max) equivalents.
+ * Used to automatically use BYOK for subagents when parent agent uses BYOK.
+ */
+const BYOK_MODEL_MAPPINGS: Record<string, string> = {
+  haiku: "haiku-pro-max",
+  "sonnet-4.5": "sonnet-4.5-pro-max",
+  "sonnet-4.5-no-reasoning": "sonnet-4.5-no-reasoning-pro-max",
+  opus: "opus-pro-max",
+};
+
+/**
+ * Check if a model ID is a BYOK variant (claude-pro-max provider)
+ */
+function isByokModelId(modelId: string | null): boolean {
+  return modelId?.endsWith("-pro-max") ?? false;
+}
+
+/**
+ * Map a standard Claude model ID to its BYOK equivalent if available
+ */
+function mapToByokModel(modelId: string): string {
+  return BYOK_MODEL_MAPPINGS[modelId] ?? modelId;
+}
+
+// ============================================================================
 // Helper Functions
 // ============================================================================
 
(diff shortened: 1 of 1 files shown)


cpfiffer:
@letta-code is this PR still valid?

PR #474, opened 2026-01-06
feat: add LSP support (TypeScript and Python)

4shub:
Adds a tool that performs a typecheck on the file before editing, this should reduce the amount of false-positives/hallucinations when a user is working with typescript code

To enable:
```
LETTA_ENABLE_LSP=true
```


cpacker:
@letta-code do you know why the CI/CD isn't running? it ran on all the previous commits - is it because it was committed via gh action?

PR #513, opened 2026-01-11
UI Hooks

paulbettner:
This adds a small UI seam to the Ink CLI so input and output rendering can be customized without replacing the whole client.

Changes:
- Allow injecting a custom `Input` component
- Add optional `ui` overrides on the CLI `App` for rendering:
  - `renderStaticItem`
  - `renderLiveItem`
  - `renderOverlay`

Defaults remain unchanged when no overrides are provided.

Validation:
- `bun run check`
- `bun run build`


4shub:
woah cool!

PR #590, opened 2026-01-19
fix: improve Ctrl-C/ESC cancellation handling

temujin9:
## Summary

Fixes several issues with stream cancellation when pressing Ctrl-C or ESC:

1. **Ctrl-C during agent execution**: The `isExecuting` check in `useInput` returned early before the Ctrl-C handler could run. Fixed by moving Ctrl-C check before the `isExecuting` return.

2. **Missing backend cancel during tool execution**: When pressing ESC while tools were executing, the code aborted client-side but never sent a cancel to the backend, leaving runs in "running" state. This caused 409 ConversationBusyError on subsequent messages.

3. **Cancel API compatibility**: The `conversations.cancel()` endpoint returns 404 on local Letta servers. Simplified to always use `agents.messages.cancel()` which works universally.

## Changes

- Modified `ApprovalDialogRich.tsx` to check for Ctrl-C before the `isExecuting` early return
- Added backend cancel call to the tool execution interrupt path in `handleInterrupt`
- Simplified `cancelCurrentRun` helper to always use agents API

## Test plan

- [x] Build passes successfully
- [x] Manual test: ESC during streaming cancels properly
- [x] Manual test: ESC during tool execution cancels properly  
- [x] Manual test: No 409 errors on subsequent messages after cancellation
- [x] Verified fix works on both Letta Cloud and local Letta servers

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -583,17 +583,11 @@ export default function App({
     conversationIdRef.current = conversationId;
   }, [conversationId]);
 
-  // Helper to cancel current run using the correct API based on conversationId
-  // For "default" conversation (agent's primary message history), use agents API
-  // For explicit conversations, use conversations API
+  // Helper to cancel current run - always use agents API since it works for all cases
+  // The /v1/agents/{id}/messages/cancel endpoint cancels runs associated with the agent
+  // regardless of which conversation they're in
   const cancelCurrentRun = async (client: ReturnType<typeof getClient> extends Promise<infer T> ? T : never) => {
-    if (conversationIdRef.current === "default") {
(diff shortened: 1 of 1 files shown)


cpacker:
@temujin9 to clarify what's the desired behavior here? (sorry having a hard time parsing)

Right now, the intended design is (if streaming):
1. Pressing `esc` cancels the stream
2. Presssing `ctrl-c` clears the input (but doesn't cancel the stream)

temujin9:
Closing - upstream has independently addressed most of these issues. The remaining tool-execution cancel fix can be re-submitted as a minimal PR against current main if needed.

PR #686, opened 2026-01-26
refactor: replace sharp with imgkit

ThatOneCalculator:
Replaces sharp with [imgkit](https://github.com/nexus-aissam/imgkit) for the following reasons:

- Faster, more lightweight
- Easier to package (see: [AUR issue when sharp was introduced](https://aur.archlinux.org/packages/letta-code#comment-1056689))

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -7,9 +7,9 @@
       "dependencies": {
         "@letta-ai/letta-client": "^1.7.6",
         "glob": "^13.0.0",
+        "imgkit": "^2.1.0",
         "ink-link": "^5.0.0",
         "open": "^10.2.0",
-        "sharp": "^0.34.5",
       },
       "devDependencies": {
         "@types/bun": "latest",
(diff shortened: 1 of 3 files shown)


cpacker:
Thanks for the PR and for maintaining the AUR package!

I looked into the sharp loading issue from the [AUR comment](https://aur.archlinux.org/packages/letta-code#comment-1056689), and I think this can be solved at the packaging layer without swapping dependencies.

### The Issue
Sharp's bundled platform-specific binaries aren't being found during the AUR build/install process.

### Recommended Fix (in the PKGBUILD)
Sharp's [installation docs](https://sharp.pixelplumbing.com/install#custom-libvips) explicitly support using system libvips instead of bundled binaries - this is the standard approach for distro packages:

```bash
depends=('nodejs' 'org.freedesktop.secrets' 'libvips')  # Add libvips

package() {
  cd "${_pkgname}-${pkgver}"
  export SHARP_FORCE_GLOBAL_LIBVIPS=1  # Use system libvips, not bundled
  bun install
  bun run build
  install -Dm755 ./letta.js "$pkgdir/usr/bin/letta"
}
```

This way:
- Arch's well-maintained `libvips` package handles the native library
- No bundled binaries to worry about
- Sharp compiles against the system library at install time

### Concerns About imgkit
We'd prefer not to replace sharp with imgkit because:
- `imgkit` was created 10 days ago (Jan 19, 2026) - it's a very new package
- Sharp is battle-tested with ~9M weekly downloads and a long track record
- Supply chain risk - swapping established dependencies for new/unknown ones is a common attack vector

If the PKGBUILD fix doesn't work, happy to discuss alternatives, but let's try the standard distro-packaging approach first.

cpacker:
^ @ThatOneCalculator asked my letta code to drop a comment here, the tldr is trying to keep deps as tight as possible in this repo, and feel a little uneasy swapping a mainline dep with a low profile package, would much rather try to solve the issue w/ sharp if possible

cpacker:
> > Maybe one thing we could do is expose a build var that changes the dep for image resizing, and pass that for the AUR build?
> 
> If you're fine with that solution, then I'll just implement it using `magick` commands, which'll be simpler and faster!

Yeah just to clarify we're on the page, something like:

1. make a bun build flag like `BYPASS_SHARP` (default false, but turned on in your build process)
2. in the code, if this on, don't use sharp / attempt to import it, but do X instead (where X works on AUR)

PR #695, opened 2026-01-27
feat: add profile auth method for bedrock

AriWebb:


sPredictorX1708:
I mainly use profiles and sometimes api-key (from bedrock)

PR #799, opened 2026-02-03
feat(tui): footer model context + Tab reasoning tiers

paulbettner:
Adds compact, always-visible model context in the interactive TUI footer, plus a fast hotkey for cycling reasoning tiers.

## Footer context
- Model label includes a reasoning tier suffix (low/med/high/xhigh) with subtle color.
- Toolset indicator is shown (`t:<toolset>`).
- System prompt indicator is shown (`s:<prompt>`), rendered lowercase and without the redundant `letta-` prefix (`default` displays as `claude`).

## Reasoning tiers hotkey (Tab)
- Press `Tab` to cycle `reasoning_effort` tiers for models that expose multiple tiers under the same handle (e.g. `openai/gpt-5.2`).
- Debounced apply: the footer updates immediately as you cycle, but the agent config update is applied after a short pause (~500ms) so it’s fast to pick a tier before committing.
- If you press Enter to submit a message while a reasoning change is pending, the pending change is flushed first so the next run uses the selected tier.

Motivation: when debugging behavior changes (prompt/toolset/model/reasoning), it helps to see and adjust these at a glance without opening selectors.

### DIFF
--- a/src/tests/agent/export-skills.test.ts
+++ b/src/tests/agent/export-skills.test.ts
@@ -8,6 +8,9 @@ describe("packageSkills from .skills/ directory", () => {
   const skillsDir = join(testDir, ".skills");
   const originalCwd = process.cwd();
 
+  // Windows CI can be significantly slower for filesystem-heavy tests.
+  const exportTimeoutMs = process.platform === "win32" ? 30000 : 15000;
+
   beforeEach(() => {
     mkdirSync(testDir, { recursive: true });
     process.chdir(testDir);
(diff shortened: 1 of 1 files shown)


paulbettner:
Closing this PR — the footer customization part is now fully handled by the awesome configurable `statusLine.command` system you shipped in recent releases (#930, #938). That's exactly the extension point we needed and it's beautifully designed — thank you for adding it! 🙏

The Tab key reasoning tier cycling feature from this PR has been split out into a separate, focused PR: #939.

PR #833, opened 2026-02-05
feat: remove queue timeout, wait for tool completion

carenthomas:
Removes 15-second timeout that cancelled long-running tools when messages were queued.

**Before:**
- Queue message while tool running → 15s timeout starts
- If tool runs >15s → Cancel operation, process queued messages
- Quick tools → Append queued messages to next turn

**After:**
- Queue message while tool running → Wait indefinitely
- Tool completes (any duration) → Append queued messages to next turn
- No forced cancellation for long-running tools

This allows tools like large file operations, network requests, or complex searches to complete naturally without interruption.

**Changes:**
- Removed `queueAppendTimeoutRef` timeout logic (-47 lines)
- Simplified `consumeQueuedMessages()` (no timeout clearing)
- Queue now waits for tool completion regardless of duration

**Testing:** All existing tests pass

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -1369,14 +1369,6 @@ export default function App({
     }
   }, []);
 
-  useEffect(() => {
-    return () => {
-      if (queueAppendTimeoutRef.current) {
-        clearTimeout(queueAppendTimeoutRef.current);
-      }
-    };
-  }, []);
-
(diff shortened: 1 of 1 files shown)


kianjones9 (approved):
lol when Caren removes 60 lines of code to make my experience better

PR #836, opened 2026-02-05
feat: braille area chart for /context tokens in context window

jnjpng:
image

## Summary
- Extract `/context` rendering into `contextChart.ts` with a braille area chart showing token usage over steps in the current session
- Reset context history on agent/model/conversation switches

## Test plan
- [ ] Run a few turns, then `/context` — verify chart appears with alternating colors per turn
- [ ] Switch models/conversations and verify history resets
- [ ] `npx tsc --noEmit` passes

### DIFF
--- a/src/cli/helpers/contextChart.ts
+++ b/src/cli/helpers/contextChart.ts
@@ -221,13 +221,13 @@ function renderBrailleChart(
     }
   }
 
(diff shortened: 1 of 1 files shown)


carenthomas:
:eyes:

PR #886, opened 2026-02-10
fix: resolve pending approvals on pre-stream 409 before retrying

### LINKED ISSUE #885: fix: pre-stream 409 from pending approval not recovered
## Problem

When a session leaves an unresolved tool approval on a conversation (e.g. user walks away from ADE), every subsequent message from any client fails with a 409 CONFLICT. The CLI detects the 409 but doesn't resolve the pending approval.

### Two distinct 409 errors from the server

The server returns different 409 errors depending on the cause:

| Error | Detected by | Meaning |
|-------|------------|---------|
| "another request is currently being processed" | `isConversationBusyError()` | A run is actively executing |
| "cannot send a new message" / "agent is waiting for approval" | `isApprovalPendingError()` | A run is blocked on an unresolved approval |

### What happens today

In `src/headless.ts`, the pre-stream error handler (line ~1172) only checks `isConversationBusyError`:

```typescript
} catch (preStreamError) {
  // ... extract errorDetail ...

  // Only this check exists:
  if (isConversationBusyError(errorDetail) && ...) {
    // Wait 2.5s, retry once
  }

  // isApprovalPendingError is NEVER checked here
  throw preStreamError; // <-- fails the turn
}
```

The **in-stream** handler (line ~1260) correctly checks `isApprovalPendingError` and sets `approvalPendingRecovery = true`, but this only runs during `drainStreamWithResume()` -- it never fires for pre-stream 409s.

### Sequence of failure

1. Session A sends a message, agent calls a tool, server pauses for HITL approval
2. Session A disconnects (or user doesn't approve)
3. Session B sends a message
4. Server returns 409: "Cannot send a new message: The agent is waiting for approval"
5. CLI checks `isConversationBusyError(errorDetail)` -- **doesn't match** (different error text)
6. Error re-thrown to outer catch
7. CLI emits `result { subtype: "error", stop_reason: "error" }`
8. SDK consumer sees `{ success: false, error: "error" }` -- unrecoverable

### The fix

Call `resolveAllPendingApprovals()` (already exists at line ~865) when `isApprovalPendingError` matches in the pre-stream 409 handler. This resolves the stale approval, then the loop retries the original message.

Also affects bidirectional mode (line ~2185): `sendMessageStream()` has no try-catch at all -- a 409 propagates to the outer catch and kills the turn.

## Impact

Any SDK consumer using a shared agent/conversation is affected. The conversation becomes permanently stuck until the approval is manually resolved in the ADE or via the REST API. headless deployments (lettabot, GitHub Action) have no way to recover.

## Workaround

lettabot PR letta-ai/lettabot#249 adds client-side recovery: scans conversation history for orphaned approvals and sends denials via the REST API. This shouldn't be necessary once the CLI handles it.

## Related

- letta-code-sdk#31 (SDK-side issue, redirects here)
- letta-code-sdk#30 (exposes `stopReason` on SDK result messages)
- lettabot#249 (client-side workaround)

just-cameron:
## Summary

When a session leaves an unresolved tool approval on a conversation, every subsequent message fails with a 409 CONFLICT. The CLI detects the 409 but doesn't resolve the pending approval -- it either waits 2.5s (wrong error type) or re-throws (no handler).

### Root cause

Two distinct 409 errors exist:
- `isConversationBusyError`: "another request is currently being processed" -- handled (wait + retry)
- `isApprovalPendingError`: "cannot send a new message" -- **not handled** in pre-stream path

The in-stream handler correctly checks `isApprovalPendingError` and triggers `resolveAllPendingApprovals()`, but the pre-stream 409 happens before the stream starts so that code never fires.

### Fix

1. **One-shot mode** (line ~1172): Add `isApprovalPendingError` check before `isConversationBusyError`. Calls `resolveAllPendingApprovals()` to clear the stale approval, then retries.

2. **Bidirectional mode** (line ~2185): `sendMessageStream()` had no try-catch. Added pre-stream error handling with the same approval recovery.

Both paths use the existing `resolveAllPendingApprovals()` function which handles the full lifecycle: fetch pending approvals, classify permissions, execute approved tools, send results back.

Fixes #885

## Test plan

- [x] `npm run build` passes (3917KB)
- [x] `bun test src/tests/approval-recovery.test.ts` -- 24 tests pass
- [x] Uses existing `isApprovalPendingError()` and `resolveAllPendingApprovals()` -- no new functions

## Related

- letta-code-sdk#31 (SDK-side issue)
- letta-code-sdk#30 (exposes stopReason on result messages)
- lettabot#249 (client-side workaround until this fix ships)

### DIFF
--- a/src/headless.ts
+++ b/src/headless.ts
@@ -1169,6 +1169,31 @@ ${SYSTEM_REMINDER_CLOSE}
           errorDetail = preStreamError.message;
         }
 
+        // Check for pending approval blocking new messages - resolve and retry.
+        // The server returns a distinct 409 when a run is waiting for approval
+        // vs when a run is actively processing. Resolve the stale approval
+        // instead of just waiting.
+        if (isApprovalPendingError(errorDetail)) {
+          if (outputFormat === "stream-json") {
+            const recoveryMsg: RecoveryMessage = {
+              type: "recovery",
+              recovery_type: "approval_pending",
+              message:
+                "Detected pending approval conflict on send; resolving before retry",
+              session_id: sessionId,
+              uuid: `recovery-pre-stream-${crypto.randomUUID()}`,
+            };
+            console.log(JSON.stringify(recoveryMsg));
+          } else {
+            console.error(
+              "Pending approval detected, resolving before retry...",
+            );
+          }
+
+          await resolveAllPendingApprovals();
+          continue;
+        }
+
         // Check for 409 "conversation busy" error - retry once with delay
         if (
           isConversationBusyError(errorDetail) &&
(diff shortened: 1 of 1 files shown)


cpacker:
Closing as superseded by #896.\n\n👾 Generated with [Letta Code](https://letta.com)

PR #895, opened 2026-02-10
fix: fire PreCompact hooks on server-side auto-compaction

### LINKED ISSUE #870: PreCompact hook is defined and tested but never called in production code
# PreCompact hook is defined and tested but never called in production code

## Issue

The `PreCompact` hook event is fully implemented in the hooks system — type definitions, input interface, executor function, integration tests — but is never invoked during actual compaction. The hook silently does nothing for any user who configures it.

## Expected behavior

When compaction occurs, `runPreCompactHooks()` should be called before the compaction executes, giving hook consumers a chance to save working state.

## Actual behavior

`runPreCompactHooks()` is never called from any production code path. Configuring a `PreCompact` hook in `settings.json` has no effect. No error is raised; the hook is silently ignored.

### 1. The function exists and is exported

**`src/hooks/index.ts:331`**
```typescript
export async function runPreCompactHooks(
```

### 2. The type is fully defined

**`src/hooks/types.ts:21`** — `PreCompact` is listed as a `SimpleHookEvent`:
```typescript
| "PreCompact" // Runs before a compact operation (cannot block)
```

**`src/hooks/types.ts:385-395`** — Full input interface defined:
```typescript
export interface PreCompactHookInput extends HookInputBase {
  event_type: "PreCompact";
  context_length?: number;
  max_context_length?: number;
  agent_id?: string;
  conversation_id?: string;
}
```

### 3. Integration tests pass

**`src/tests/hooks/integration.test.ts:781-843`** — Three test cases:
- "runs before compact operation" (line 782)
- "can block compact operation (exit 2)" (line 805)
- timeout handling (line 831)

All call `runPreCompactHooks()` directly and verify it works.

### 4. No production code imports or calls it

A full search of the source tree for `runPreCompactHooks` returns only:
- `src/hooks/index.ts` — the definition
- `src/tests/hooks/integration.test.ts` — the tests

No file in `src/cli/`, `src/agent/`, `src/index.ts`, or any other production path imports or calls this function.

### 5. The compaction detection code exists but doesn't invoke the hook

**`src/cli/helpers/accumulator.ts`** is the file that handles compaction. It:
- Detects compaction via `pendingCompaction` flag (line 884)
- Processes `compaction_stats` from `summary_message` chunks (line 845-855)
- Renders compaction events for the UI (line 862)
- **Already imports and calls other hooks** — `runPreToolUseHooks` (line 9, 679) and `runPostToolUseHooks` (line 9, 771)
- **Does NOT import or call `runPreCompactHooks`**

The pattern for calling hooks is already established in this file. The compaction detection is already in this file. The two were never connected.

### 6. Settings loader validates PreCompact as a valid hook event

**`src/tests/settings-manager.test.ts:667`** — confirms `PreCompact` is a recognized event name in settings validation, so users CAN configure it — it just won't fire.

## Impact

Users who configure `PreCompact` hooks (e.g., to auto-save agent state before compaction) will see the hook accepted in their settings without errors but it will never execute. This is particularly impactful for stateful agents where compaction destroys working context that cannot be recovered without a pre-compaction save.

## Suggested fix

In `src/cli/helpers/accumulator.ts`, at the point where compaction is detected (around line 884 where `ctx.pendingCompaction = true`), add a call to `runPreCompactHooks()` — following the same pattern already used for `runPreToolUseHooks` and `runPostToolUseHooks` in the same file.

## Version

Tested against `letta-code` v0.14.14 (cloned Feb 9, 2026 from `main`).

jnjpng:
## Summary
- Wire up `runPreCompactHooks()` in the stream accumulator (`src/cli/helpers/accumulator.ts`) when a compaction `event_message` is received during streaming
- The hook was fully implemented (types, executor, integration tests) but only called on manual `/compact` and not in the streaming path. 

Closes #870

## Example PreCompact hook for testing

Add to `.letta/settings.json`:

```json
{
  "hooks": {
    "PreCompact": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "echo 'Compaction starting!' >> /tmp/compaction.log"
          }
        ]
      }
    ]
  }
}
```

Then trigger compaction (either via `/compact` or by filling context until auto-compaction fires) and check `/tmp/compaction.log`.

## Test plan
- [x] Existing `PreCompact` integration tests pass (41/41)
- [x] Lint and typecheck pass
- [x] Manual test: configure a `PreCompact` hook, trigger auto-compaction, verify hook output appears

### DIFF
--- a/src/cli/helpers/accumulator.ts
+++ b/src/cli/helpers/accumulator.ts
@@ -921,8 +921,8 @@ export function onChunk(
             undefined, // max_context_length not available here
             b.agentId,
             undefined, // conversationId not available here
-          ).catch(() => {
-            // Silently ignore - don't block stream processing
+          ).catch((error) => {
+            debugLog("hooks", "PreCompact hook error (accumulator)", error);
           });
         }
 


src/cli/helpers/accumulator.ts
```diff
@@ -910,6 +914,18 @@ export function onChunk(
...
 
+        // Fire PreCompact hooks when server-side auto-compaction starts
+        if (eventType === "compaction") {
+          runPreCompactHooks(
+            ctx?.lastContextTokens,
+            undefined, // max_context_length not available here
+            b.agentId,
+            undefined, // conversationId not available here
+          ).catch(() => {
+            // Silently ignore - don't block stream processing
```
jnjpng: @letta-code can you add a debug log instead of silently ignoring. check the rest of the codebase for examples since we do this already.

PR #910, opened 2026-02-11
feat(harness): add system alert for conversation switching

kianjones9:
In Letta, we have this concept of conversations. I was thinking it would be super cool and ✨ agentic ✨ to add a system alert for letting the agent know we just switched into that conversation. The idea is to prime the agent to use the in-conversation context.

A perfect example of how this might be useful is when users invoke the very cool `/search` command, they can search by keyword, vector, or hybrid, and select a message from the search results to get there. What if the agent could get a head start by knowing the search context that brought you there, to better understand your intent?

I give you: "The conversation switch alert ™️ "

### DIFF
--- a/src/settings-manager.ts
+++ b/src/settings-manager.ts
@@ -117,16 +117,12 @@ const DEFAULT_SETTINGS: Settings = {
   tokenStreaming: false,
   showCompactions: false,
   enableSleeptime: false,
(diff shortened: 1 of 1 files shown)


kianjones9:
@letta-code can you resolve the MC here?

PR #920, opened 2026-02-12
fix: use slower reflection interval when memfs is enabled

cpacker:
## Summary
- When memfs is enabled, the reflection subagent handles memory consolidation asynchronously in the background, so the default 5-turn interval is too aggressive
- `buildMemoryReminder` now computes `effectiveInterval = max(configured, 10)` when memfs is on — respects user-configured intervals > 10 while enforcing a minimum of 10

## Behavior

| memfs | configured interval | effective interval |
|-------|--------------------|--------------------|
| off | 5 (default) | 5 |
| on | 5 (default) | 10 |
| on | 3 (user set) | 10 |
| on | 15 (user set) | 15 |
| on | null (disabled) | null (still disabled) |

No settings are mutated — this is purely computed at call time in `buildMemoryReminder`.

### DIFF
--- a/src/cli/helpers/memoryReminder.ts
+++ b/src/cli/helpers/memoryReminder.ts
@@ -45,11 +45,22 @@ export async function buildMemoryReminder(
 ): Promise<string> {
   const memoryInterval = getMemoryInterval();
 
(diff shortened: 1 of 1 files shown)


cpacker:
Superceded by #924

PR #921, opened 2026-02-12
Add mode options to `/compact`

amysguan:
- Supports `/compact all` and `/compact sliding_window`

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -1208,7 +1208,9 @@ export default function App({
   }, [llmConfig]);
   const [currentModelId, setCurrentModelId] = useState<string | null>(null);
   // Full model handle for API calls (e.g., "anthropic/claude-sonnet-4-5-20251101")
-  const [currentModelHandle, setCurrentModelHandle] = useState<string | null>(null);
+  const [currentModelHandle, setCurrentModelHandle] = useState<string | null>(
+    null,
+  );
   // Derive agentName from agentState (single source of truth)
   const agentName = agentState?.name ?? null;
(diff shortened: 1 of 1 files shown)


src/cli/App.tsx
```diff
@@ -6188,18 +6212,38 @@ export default function App({
             }
 
             const client = await getClient();
+
+            // Compute model handle from llmConfig
+            const modelHandle =
+              llmConfig?.model_endpoint_type && llmConfig?.model
+                ? `${llmConfig.model_endpoint_type}/${llmConfig.model}`
```
cpacker, line 6219: I think llmconfig is deprecated / outdated, probably better to just take agent_state.modelsettings and use that instead of model?

Image

cpacker: i thin kwe should probably pass model and modelsettings just to match the summarizer config 1:1 where possible for caching where possible

cpacker:
```
# if the user doesn't specify anything, just `compact`
-> this should just use `agent_state.compaction_settings`, either by grabbing it here and sending it in the .compact(...), or by relying on the server to do this

# if the user specified a compaction mode `/compact all`, then just send the mode but specify nothing else
-> this can potentially mean we'll drop custom settings on agent_state.compaction_settings?

alternatively, try to be smarter

if agent_state.compaction_settings.mode == /compact mode, then use the agent_state.compaction_settings (implicit by sending nothing, I think?)

if they don't match, then just send the mode + empty payload, and we expect this to use carefully chosen server-side defaults
```

PR #934, opened 2026-02-12
fix: /context double render on short terminals and stale loading text

jnjpng:
## Summary
- Fix `/context` rendering the full chart during the "running" phase — on short terminals, the tall dynamic-area content scrolls into the terminal scrollback (outside Ink's control) and persists alongside the final committed output, causing a duplicate render
- Fix "Fetching breakdown..." text staying permanently when the breakdown fetch times out or fails — add a `loading` flag to `renderContextUsage` so the placeholder is only shown while actively loading

## Test plan
- [ ] Run `/context` on a short terminal (e.g. ~20 rows) and verify only one output appears
- [ ] Run `/context` with the server unreachable/slow to verify it shows the chart without breakdown (no stale "Fetching breakdown..." text)
- [ ] Run `/context` normally and verify breakdown categories still render correctly

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -6171,11 +6171,6 @@ export default function App({
           const usedTokens = contextTrackerRef.current.lastContextTokens;
           const history = contextTrackerRef.current.contextTokensHistory;
 
-          // Show a lightweight loading indicator while fetching breakdown.
-          // Avoid rendering the full chart in the "running" phase because
-          // tall dynamic-area content on short terminals scrolls into the
-          // terminal scrollback and persists after Ink commits the final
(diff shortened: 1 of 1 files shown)


src/cli/App.tsx
```diff
@@ -6171,23 +6171,17 @@ export default function App({
...
-            history,
-          });
-
-          const cmd = commandRunner.start(trimmed, "");
-          cmd.update({
-            output: initialOutput,
-            phase: "running",
-            preformatted: true,
-          });
+          // Show a lightweight loading indicator while fetching breakdown.
```
jnjpng: @letta-code remove this comment

PR #939, opened 2026-02-13
feat(tui): cycle reasoning tiers with Tab key

paulbettner:
## Summary

Add a TUI shortcut to cycle reasoning_effort tiers for models that expose multiple tiers under the same handle (for example: openai/gpt-5.2).

- Tab cycles reasoning tiers (when autocomplete is not active)
- Shift+Tab continues to cycle permission modes (unchanged)
- Uses a short debounce before applying the tier change via API (avoids rapid-fire updates)
- Flushes any pending tier change before sending a message
- Normalizes chatgpt_oauth/* runtime handles to the configured provider namespace (chatgpt-plus-pro/*) so Codex BYOK tiers can be cycled

## Test plan
- [ ] Verify Tab cycles through reasoning tiers on a multi-tier model (e.g. gpt-5.2)
- [ ] Verify Shift+Tab still cycles permission modes
- [ ] Verify debounce prevents rapid API calls when cycling quickly
- [ ] Verify pending tier is applied before message send
- [ ] Verify Tab does nothing on models with only one tier
- [ ] Verify autocomplete takes priority over tier cycling when active

Generated with Letta Code

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -45,7 +45,6 @@ import { getClient, getServerUrl } from "../agent/client";
 import { getCurrentAgentId, setCurrentAgentId } from "../agent/context";
 import { type AgentProvenance, createAgent } from "../agent/create";
 import { getLettaCodeHeaders } from "../agent/http-headers";
-
 import { ISOLATED_BLOCK_LABELS } from "../agent/memory";
 import {
   ensureMemoryFilesystemDirs,
@@ -77,6 +76,7 @@ import {
 } from "../hooks";
 import type { ApprovalContext } from "../permissions/analyzer";
 import { type PermissionMode, permissionMode } from "../permissions/mode";
+import { OPENAI_CODEX_PROVIDER_NAME } from "../providers/openai-codex-provider";
 import {
   DEFAULT_COMPLETION_PROMISE,
   type RalphState,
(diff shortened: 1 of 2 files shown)


cpacker:
FYI @paulbettner just as an update here - related PR just went up to add reasoning cycling as an extra page when you enter your model selection (for the latest codex / claude models):
image

I'm also working on a PR rn that exposes the reasoning effort in the bottom-right, as well as in the `/` command preview. LMK if you still think this tab cycling is useful? I feel like in most cases these days, people don't really swap this that frequently, so I'm not sure if it's better to actually expose it so easily vs requiring going into `/model` to change.

paulbettner:
Yep! My bot says this PR is ready now @cpacker:

```
Re: cpacker's debounce/flush feedback — this PR already does the full pattern in src/cli/App.tsx:
    - trailing debounce (reasoningCycleDebounceMs = 500)
    - optimistic local UI update on Tab press (setLlmConfig + setCurrentModelId)
    - delayed server apply via updateAgentLLMConfig(...)
    - in-flight guard (reasoningCycleInFlightRef) + busy guard (isAgentBusy() -> reschedule)
    - flush-before-send is wired: onSubmit awaits flushPendingReasoningEffort() before starting the next send/run
```

PR #947, opened 2026-02-13
refactor: extract memory section from system prompts into conditional addons

sarahwooders:
## Summary
- Extracted the `# Memory` section from all 4 system prompt files (`system_prompt.txt`, `letta_claude.md`, `letta_codex.md`, `letta_gemini.md`) into separate addon files
- Created `system_prompt_memory.txt` (non-memfs memory instructions) and updated `system_prompt_memfs.txt` to include both memory overview + filesystem instructions under a unified `# Memory` heading
- `createAgent()` and `updateAgentSystemPrompt()` now append the non-memfs memory addon by default
- `updateAgentSystemPromptMemfs()` now **swaps** between the two memory sections (strips existing `# Memory`, appends the correct version) instead of just adding/removing the memfs addon

## Motivation
Previously when memfs was enabled, agents got **both** the inline `# Memory` section from the base prompt **and** the `## Memory Filesystem` addon — redundant and potentially conflicting instructions. Now agents always get exactly one memory section matching their configuration.

## Test plan
- [ ] Create a new agent without `--memfs` → verify system prompt ends with the non-memfs `# Memory` section
- [ ] Create a new agent with `--memfs` → verify system prompt ends with the memfs `# Memory` section (includes Memory Filesystem subsections)
- [ ] Toggle memfs on an existing agent → verify the memory section swaps correctly
- [ ] Toggle memfs off → verify it swaps back to the non-memfs version
- [ ] Switch system prompt via `--system letta-codex` → verify memory section is preserved

### DIFF
--- a/src/agent/create.ts
+++ b/src/agent/create.ts
@@ -17,7 +17,10 @@ import {
   resolveModel,
 } from "./model";
 import { updateAgentLLMConfig } from "./modify";
-import { SYSTEM_PROMPT_MEMORY_ADDON, resolveSystemPrompt } from "./promptAssets";
+import {
+  resolveSystemPrompt,
+  SYSTEM_PROMPT_MEMORY_ADDON,
+} from "./promptAssets";
 import { SLEEPTIME_MEMORY_PERSONA } from "./prompts/sleeptime";
 
 /**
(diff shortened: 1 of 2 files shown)


src/agent/prompts/system_prompt_memfs.txt
```diff
@@ -1,26 +1,22 @@
 
-## Memory Filesystem
-Your memory is stored in a git repository at `~/.letta/agents/<agent-id>/memory/`. This provides full version control, sync with the server, and branching for parallel edits.
-
-### Structure
```
sarahwooders, line 5: removed since this is already in the system prompt (in `<memory_filesystem>`

PR #954, opened 2026-02-13
feat: add Letta agent self-history analysis to /init skill

sarahwooders:
Add scripts and SKILL.md instructions for the agent to review its own past conversation history during /init, similar to the existing Claude Code/Codex history analysis flow.

New scripts:
- detect-own-history.ts: list conversations + date ranges via Letta API
- split-history.ts: split all history into chunk files (conversation × month)
- extract-messages.ts: extract a single chunk (low-level building block)

Messages are filtered for signal: user messages in full, assistant truncated to 500 chars, reasoning to 200 chars, tool calls as one-liner summaries, tool returns dropped entirely. Chunks are pre-extracted to local files so history-analyzer subagents need no API access.

### DIFF
--- a/src/skills/builtin/initializing-memory/SKILL.md
+++ b/src/skills/builtin/initializing-memory/SKILL.md
@@ -734,74 +734,23 @@ git push
 
 ## Letta Agent History Analysis (Optional)
 
-This section runs only if the user approved during upfront questions. It uses the same parallel `history-analyzer` subagent pattern as the Claude/Codex analysis above, but reads from your **own** past conversations via the Letta API.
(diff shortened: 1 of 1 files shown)


src/skills/builtin/initializing-memory/scripts/detect-own-history.ts
```diff
@@ -0,0 +1,115 @@
...
+function extractDate(msg: Record<string, unknown>): string | null {
+  if ("date" in msg && typeof msg.date === "string") return msg.date;
+  return null;
+}
+
+// ---------------------------------------------------------------------------
+// Main
+// ---------------------------------------------------------------------------
+
+async function main() {
```
cpacker, line 33: built-in scripts using the Letta API should be converted into CLI, similar to `letta messages list`, since it is much more robust for auth

cpacker (changes requested):
should be folded into CLI not scripts

PR #964, opened 2026-02-15
fix: resolve shell command sandbox bypass and unauthorized file access

RinZ27:
Updating the shell command permission logic fixes a couple of security-related issues I found in the sandbox.

Chained commands starting with `cd` into a memory directory were previously bypassing subsequent validation. Even with `MEMORY_DIR_APPROVE_ALL` enabled, ensuring that command segments don't use absolute paths or parent directory references is vital to prevent escaping the intended scope. All path-like tokens in these segments are now validated.

Refining the "always safe" command list was also necessary. While commands like `cat`, `grep`, or `head` are read-only, they lacked path validation, which could allow reading sensitive system files via absolute paths. A conservative check has been added to prevent auto-approval if any argument looks like an absolute path or a traversal attempt.

Robustness of the sandbox is improved while maintaining the "auto-approve" convenience for standard, scoped operations. Regression tests are included in `src/tests/permissions_security.test.ts`.

Verification was done locally using `bun test`.

### DIFF
--- a/src/permissions/readOnlyShell.ts
+++ b/src/permissions/readOnlyShell.ts
@@ -185,61 +185,72 @@ function isSafeSegment(segment: string): boolean {
     return isReadOnlyShellCommand(stripQuotes(nested));
   }
 
-  if (!ALWAYS_SAFE_COMMANDS.has(command)) {
-    if (command === "git") {
-      const subcommand = tokens[1];
-      if (!subcommand) {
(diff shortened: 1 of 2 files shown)


RinZ27:
@cpacker appreciate the heads up on those Windows path cases. Glad the logic I proposed was useful for the final fix.

PR #1020, opened 2026-02-19
fix: reset conversation routing after /agents new agent creation

jnjpng:
## Summary
- reset the `/agents` new-agent creation flow to explicitly switch routing to the new agent's `default` conversation
- persist local/global last-session state with `conversationId: "default"` for the newly created agent to prevent stale conversation routing
- add regression wiring assertions in `bootstrap-reminders-reset-wiring.test.ts` for new-agent conversation reset behavior

## Test plan
- [x] bun run fix
- [x] bun run lint
- [x] bun run typecheck
- [x] bun run check
- [x] bun test src/tests/cli/bootstrap-reminders-reset-wiring.test.ts

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -5248,6 +5248,19 @@ export default function App({
         // Update project settings with new agent
         await updateProjectSettings({ lastAgent: agent.id });
 
+        // New agents always start on their default conversation route.
+        // Persist this explicitly so routing and resume state do not retain
+        // a previous agent's non-default conversation id.
+        const targetConversationId = "default";
+        settingsManager.setLocalLastSession(
+          { agentId: agent.id, conversationId: targetConversationId },
+          process.cwd(),
+        );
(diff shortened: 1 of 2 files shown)


jnjpng:
Quick context on the fix:

This addresses a routing mismatch in the `/agents` → create flow where UI switched to the new agent, but message requests could still target a prior non-default conversation.

### What changed
- In `handleCreateNewAgent` (`src/cli/App.tsx`), we now explicitly reset routing to the new agent’s default conversation via `targetConversationId = \"default\"`.
- We persist last-session state (local + global) with the new `agentId` and `conversationId: \"default\"`.
- We set conversation state/switch context accordingly so follow-up turns are routed consistently.

### Regression coverage
- Added assertions in `src/tests/cli/bootstrap-reminders-reset-wiring.test.ts` to verify this new-agent flow sets default conversation routing and persists session state.

This should prevent follow-up turns from being sent to stale conversation IDs after creating a new agent from `/agents`.

PR #1023, opened 2026-02-19
fix(permissions): avoid stale YOLO mode during approvals

paulbettner:
## Problem
Even when the footer indicates YOLO/bypassPermissions mode, tool approvals can still be classified as needing user input (commonly ShellCommand prompts). This can happen because uiPermissionModeRef is updated via useEffect, so it can be stale during the render/effect window.

## Fix
- Wrap setUiPermissionMode so uiPermissionModeRef updates synchronously whenever the UI mode changes.
- Keep callback dependency arrays correct.

## Test plan
- [ ] Enable YOLO, trigger ShellCommand (e.g. rg/ls pipelines) and confirm it auto-runs without an approval dialog.

Generated with Letta Code

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -932,14 +932,17 @@ export default function App({
   const [networkPhase, setNetworkPhase] = useState<
     "upload" | "download" | "error" | null
   >(null);
-  // Track permission mode changes for UI updates
-  const [uiPermissionMode, setUiPermissionMode] = useState(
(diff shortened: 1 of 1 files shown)


paulbettner:
Apologies for the extra PR @cpacker - I (think) I found a few bugs today while trying to track down why letta code was falling out of (or ignoring) yolo mode sometimes.

PR #1043, opened 2026-02-19
fix(cli): format agents-limit-exceeded error consistently (LET-7552)

AriWebb:
## Summary

- Add test coverage for 429 "Rate limited" error with `agents-limit-exceeded` reason
- The handler already existed in `errorFormatter.ts` but was missing test coverage
- Ensures consistent friendly message ("agent limit") instead of raw JSON

## Test plan

- [x] Added unit test for agents-limit-exceeded error format
- [x] All existing tests pass


AriWebb:
duplicate

PR #1096, opened 2026-02-23
fix: clarify plan-mode path guidance for apply_patch

### LINKED ISSUE #1045: Failing to exit plan mode
Using GPT5.2 with high reasoning, I often see the model struggling to exit plan mode. This is with the system prompt and toolset set to codex.

Here's an example of what the failure looks like:

Image

cpacker:
## Summary
- improve plan-mode denial guidance to include both the absolute plan file path and an apply_patch-compatible relative path hint
- update the in-context plan-mode reminder with explicit path rules per tool type (Write/Edit absolute path vs ApplyPatch/apply_patch relative file directives)
- add a regression test to lock in the relative-path hint in plan-mode deny messaging

## Why
Codex and GPT plan-mode runs can get stuck after trying to write the suggested absolute plan file path with apply_patch, which only accepts relative file directives.

## Validation
- bun test src/tests/permissions-mode.test.ts

Closes #1045

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -2,7 +2,7 @@
 
 import { existsSync, readFileSync, renameSync, writeFileSync } from "node:fs";
 import { homedir, tmpdir } from "node:os";
-import { join } from "node:path";
+import { join, relative, resolve } from "node:path";
 import { APIError, APIUserAbortError } from "@letta-ai/letta-client/core/error";
 import type {
   AgentState,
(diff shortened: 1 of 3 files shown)


src/cli/App.tsx
```diff
@@ -608,6 +608,25 @@ function saveLastAgentBeforeExit() {
...
+  const relativePlanPath = relative(process.cwd(), absolutePlanPath);
+  const applyPatchPath =
+    relativePlanPath && relativePlanPath !== "."
+      ? relativePlanPath
+      : absolutePlanPath;
+
+  return [
+    `No plan file exists yet. You should create your plan at ${absolutePlanPath} using a write tool (e.g. Write, ApplyPatch, etc. depending on your toolset).`,
+    `If you use Write/Edit, pass the absolute path above.`,
+    `If you use ApplyPatch/apply_patch, patch file directives must use a relative path from the current working directory: ${applyPatchPath}.`,
```
cpacker, line 626: we should be able to dynamically set this from the harness data

PR #1123, opened 2026-02-24
fix: rewrite memory subagent prompt and fix reflection path duplication bug

just-cameron:
## Summary

Two issues discovered during real-world memory defragmentation of a long-running agent (lettabot-builder, ~39K messages over several weeks, 46 pinned system/ files consuming 73KB of context on every turn).

### 1. memory.md had spiraled into over-prescription

The memory defrag subagent prompt was biased toward one operation (splitting) and imposed arbitrary constraints that didn't adapt to actual memory state:

- **Hard targets that don't fit**: "15-25 files", "40 lines max" -- in practice, my agent needed to go from 46 system/ files down to 4. The old prompt would have split those 46 files into *even more* files, all still pinned to the system prompt.
- **Missing the key decision**: The most impactful thing defrag can do is **move blocks out of system/** (always loaded, costs tokens every turn) **to root** (on-demand, loaded only when needed). This was barely mentioned.
- **"Explode" framing**: The prompt literally said "explode messy memory into deeply hierarchical structure" and "if you have fewer than 15 files, you haven't split enough." This pushed the agent toward splitting when demotion or consolidation was the right call.
- **Repetitive**: The same `/` naming guidance appeared 5 times. The file count target appeared 3 times. The checklist, reminder section, and guiding principles all restated the same constraints.

**Real-world result**: When I ran defrag with specific instructions overriding the prompt, the agent reduced system/ from 73KB (46 files) to 4.3KB (4 files) -- a 94% reduction in always-loaded context. The old prompt would have made this worse, not better.

### 2. reflection.md had a path bug creating duplicate blocks

The reflection subagent sets `WORK=$WORKTREE_DIR/$BRANCH/system` -- so `$WORK` already points INTO the `system/` directory. But when the agent saw paths like `system/human.md` in the `memory_filesystem` tree and recreated them under `$WORK`, it wrote `$WORK/system/human.md` which resolved to `system/system/human.md`.

This created an **exact byte-for-byte duplicate** of every system/ block:
```
system/human.md         (2,133 bytes)
system/system/human.md  (2,133 bytes)  <-- duplicate!
system/project.md       (8,374 bytes)
system/system/project.md (8,374 bytes) <-- duplicate!
... (23 duplicated files total)
```

50% of my pinned context budget was wasted on redundant copies. The duplication happened silently over multiple reflection runs and wasn't caught because the content was identical.

### memory.md (full rewrite, 167 -> 90 lines)

- **Leads with the key decision**: system/ is expensive (always in context), root is cheap (on-demand). The first section explains this tradeoff with concrete guidance on what belongs where.
- **Balanced operations**: DEMOTE (move out of system/) and CONDENSE (shrink system/ block, move detail to on-demand child) are now first-class operations alongside split/merge/clean/delete.
- **No arbitrary targets**: Removed "15-25 files" and "40 lines max". The right structure depends on the agent's actual memory.
- **Says things once**: Each guideline stated once. No repeated metrics, no redundant checklist, no reminder section restating the goal.

### reflection.md (targeted fix, +16 lines)

- Added explicit warning that `$WORK` already points to `system/` with correct vs wrong path examples:
  ```
  Correct: $WORK/human.md           -> creates system/human.md
  WRONG:   $WORK/system/human.md    -> creates system/system/human.md
  ```
- Explains that `memory_filesystem` shows `system/` prefix in its tree but `$WORK` paths should drop it
- Fixed example `Read()` calls to use correct relative paths

## Test plan

- [x] Verified the rewritten memory.md prompt works in practice (lettabot defrag ran successfully with similar instructions, producing 94% context reduction)
- [ ] Run a reflection cycle on an agent with the fixed paths to confirm no `system/system/` duplication
- [ ] Run a memory defrag on a test agent using the new prompt without overrides

### DIFF
--- a/src/agent/subagents/builtin/reflection.md
+++ b/src/agent/subagents/builtin/reflection.md
@@ -145,18 +145,6 @@ Edit files in the **worktree**, not the main memory dir:
 WORK=$WORKTREE_DIR/$BRANCH/system
 ```
 
-**IMPORTANT: `$WORK` already points to the `system/` directory.**
-Paths are relative to `system/`, do NOT include `system/` again:
-```
-Correct: $WORK/human.md           → creates system/human.md
-Correct: $WORK/project/gotchas.md → creates system/project/gotchas.md
-WRONG:   $WORK/system/human.md    → creates system/system/human.md (DUPLICATE!)
-```
-
-The `memory_filesystem` block in the system prompt lists paths
-like `system/human.md` — but when writing to `$WORK`, drop the
-`system/` prefix since `$WORK` already IS system/.
-
 **Before editing, read existing files:**
 ```bash
 ls $WORK/
@@ -164,8 +152,8 @@ ls $WORK/
 
 Then read relevant files:
 ```
-Read({ file_path: "$WORK/human.md" })
-Read({ file_path: "$WORK/persona.md" })
+Read({ file_path: "$WORK/human/personal_info.md" })
+Read({ file_path: "$WORK/persona/soul.md" })
 ```
 
 **Editing rules:**


src/agent/subagents/builtin/reflection.md
```diff
@@ -145,15 +145,27 @@ Edit files in the **worktree**, not the main memory dir:
 WORK=$WORKTREE_DIR/$BRANCH/system
```
sarahwooders, line 145: wait, why does this contain `system`? maybe we should just drop it?

src/agent/subagents/builtin/reflection.md
```diff
@@ -145,15 +145,27 @@ Edit files in the **worktree**, not the main memory dir:
...
 
+**IMPORTANT: `$WORK` already points to the `system/` directory.**
+Paths are relative to `system/`, do NOT include `system/` again:
+```
+Correct: $WORK/human.md           → creates system/human.md
+Correct: $WORK/project/gotchas.md → creates system/project/gotchas.md
+WRONG:   $WORK/system/human.md    → creates system/system/human.md (DUPLICATE!)
+```
+
+The `memory_filesystem` block in the system prompt lists paths
```
sarahwooders: we should drop all mentions of blocks

src/agent/subagents/builtin/reflection.md
kl2806: Yes, i think we should remove the `system`, going to merge in memory subagent prompt updates first.

PR #1186, opened 2026-02-27
fix(plan): allow TaskOutput and absolute paths in plan mode

sarahwooders:
## Summary
- Add `TaskOutput` to the plan mode allowlist so codex agents can read results from Task/Explore subagents launched during planning
- Remove the absolute path restriction for read-only shell commands (`ALWAYS_SAFE_COMMANDS`). Commands like `ls /Users/.../plans` were being denied even though they're purely read-only. The `cd` restriction is preserved to prevent path-traversal bypasses.


cpacker:
https://github.com/letta-ai/letta-code/pull/1188

PR #1238, opened 2026-03-03
fix(tui): keep conversation model overrides sticky

paulbettner:
## Summary
- Prevents conversation-scoped /model overrides from being clobbered by the first-chunk agent sync by keeping the override flag ref synchronized.
- Caches conversation `model_settings` while an override is active and uses it to derive reasoning effort, fixing missing footer reasoning tags on resume.
- Fixes the "default" (virtual) conversation case: /model and reasoning tier changes now update the agent config so the next agent sync doesn’t snap back.
- Adds wiring tests to lock in the override/ref sync and reasoning derivation behavior.

## Test plan
- [ ] `bun run check`
- [ ] `bun test src/tests/agent/model-preset-refresh.wiring.test.ts`
- [ ] Manual: `/model <some-tiered-model>` then immediately send a message; ensure the UI does not flip back to the agent's base model.
- [ ] Manual: resume a conversation with a conversation-scoped reasoning override; ensure the footer reasoning tag matches the override.
- [ ] Manual: in the default conversation, switch models and send a message; ensure the model does not snap back after the turn.

### DIFF
--- a/src/agent/check-approval.ts
+++ b/src/agent/check-approval.ts
@@ -474,15 +474,14 @@ export async function getResumeData(
       const retrievedMessages = await client.messages.retrieve(lastInContextId);
 
       // Fetch message history for backfill through the default conversation route.
-      // For default conversation, pass agent_id as query parameter.
+      // Default conversation is represented by the agent id at the conversations endpoint.
       // Wrapped in try/catch so backfill failures don't crash the CLI (e.g., older servers
       // may not support this pattern)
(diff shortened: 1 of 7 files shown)


jnjpng:
@paulbettner How were you observing this bug? E.g. changing model on either non-default/default conversation would immediately snap back after the first message sent after model switch? Or was it changing back non-deterministically after some amount of turns?

PR #1264, opened 2026-03-04
feat(input): make Backspace/Cmd+Backspace respect line boundaries

resatu:
## Summary

Backspace (Option+Backspace and Cmd+Backspace) now deletes only to the start of the **current line**, not the entire text.

## Changes

- **PasteAwareTextInput.tsx**: `findPreviousWordBoundary` stops at newlines so Option+Backspace respects line boundaries.
- **vendor/ink-text-input**: Ctrl+U (Cmd+Backspace on macOS) is line-aware: deletes from start of current line to cursor.

## Test plan

1. Multi-line input (Shift+Enter), type on line 2.
2. Press Cmd+Backspace or Option+Backspace.
3. Only text to the start of the current line is removed; previous lines stay.

### DIFF
--- a/src/cli/components/PasteAwareTextInput.tsx
+++ b/src/cli/components/PasteAwareTextInput.tsx
@@ -57,24 +57,32 @@ function sanitizeForDisplay(text: string): string {
   return text.replace(/\r\n|\r|\n/g, "↵");
 }
 
-/** Find the boundary of the previous word for option+left navigation */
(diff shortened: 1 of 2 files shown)


jnjpng:
@resatu Thanks for opening this - which terminal are you using?

resatu:
I mainly used Warp - now I've started using xtermjs.org over Graphyn.xyz 

Why did you asked?

PR #1279, opened 2026-03-05
fix(plan): add return-to-previous mode plan exit

paulbettner:
## Summary
- ExitPlanMode approval now defaults to returning to the permission mode that was active before plan mode (and shows that mode in the option label).
- Avoids duplicate options when the previous mode already corresponds to manual approvals or auto-accept edits.
- Syncs the permissionMode singleton with the UI mode right before switching to plan mode (EnterPlanMode approval + /plan).

## Test plan
- [x] bun test src/tests/cli/plan-exit-approval.test.ts
- [x] bun test src/tests/cli/permission-mode-retry-wiring.test.ts

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -79,6 +79,7 @@ import {
 } from "../hooks";
 import type { ApprovalContext } from "../permissions/analyzer";
 import { type PermissionMode, permissionMode } from "../permissions/mode";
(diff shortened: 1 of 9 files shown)


paulbettner:
My primary goal here was/is allowing plan mode exit to return to yolo mode, when it was in that mode before planning began.

PR #1339, opened 2026-03-10
fix: use client.conversations.recompile instead of client.agents.reco…

kl2806:
…mpile

The letta-client SDK moved system prompt recompilation from client.agents.recompile to client.conversations.recompile. The update_timestamp param was also removed from the API.

### DIFF
--- a/src/agent/modify.ts
+++ b/src/agent/modify.ts
@@ -251,7 +251,7 @@ export interface RecompileAgentSystemPromptOptions {
   dryRun?: boolean;
 }
 
-interface AgentSystemPromptRecompileClient {
+interface ConversationSystemPromptRecompileClient {
   conversations: {
     recompile: (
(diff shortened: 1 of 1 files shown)


src/agent/modify.ts
```diff
@@ -249,16 +249,14 @@ export async function updateConversationLLMConfig(
 
 export interface RecompileAgentSystemPromptOptions {
   dryRun?: boolean;
-  updateTimestamp?: boolean;
 }
 
 interface AgentSystemPromptRecompileClient {
```
devanshrj: should we rename this to `ConversationSystemPromptRecompileClient`?

devanshrj (approved):

src/agent/modify.ts
cpacker: yus

Issue #1419, opened 2026-03-17
PreToolUse hooks should fire before approval UI for interactive tools


backnotprop:
## Problem

`PreToolUse` hooks for `ExitPlanMode` fire after the user has already approved in the UI dialog, not before tool execution as documented. This makes it impossible to build automated policy checks (e.g., evaluating a plan against agent memory) that gate whether the approval dialog even appears.

The root cause is a two-phase execution model: the server returns `requires_approval`, the UI renders the approval dialog, the user decides, and only then does `executeTool()` run locally where `runPreToolUseHooks()` lives. Interactive approval tools (`ExitPlanMode`, `EnterPlanMode`, `AskUserQuestion`) bypass the permission checker entirely, so `PermissionRequest` hooks never fire either.

Verified by testing: a `PreToolUse` hook with matcher `ExitPlanMode` only logs after the user approves, not when the dialog appears.

*Caveat: this analysis is based on tracing through `App.tsx` (~14,500 lines of React), so it's possible a subtlety was missed.*

## Use case

A hook that checks agent memory files and evaluates whether the plan incorporates them before the user sees the approval dialog. Other examples: validating plans against scope, style guides, or required files.

selfishly, [plannotator](https://github.com/backnotprop/plannotator)

## Suggested fix

Fire `runPreToolUseHooks()` when the pending approval arrives for interactive tools, before the guard `useEffect` renders the approval UI. If the hook blocks (exit 2), reject back to the agent with feedback - the user never sees the dialog.

sarahwooders:
Closing as stale: no activity in 2+ weeks (cleanup on 2026-07-02). If this is still an issue, please comment or reopen and we'll take another look.

devanshrj:
Following up on the PR #1420 closure: plan mode was removed entirely in #2413, and its successor goal mode was also removed in #2792. `ExitPlanMode` (and `EnterPlanMode`) no longer exist in the codebase — only references in a docs file remain.

The original problem (PreToolUse hooks fire after approval, not before) may still apply to `AskUserQuestion` which still has an interactive approval flow, but the plan-mode-specific framing of this issue is no longer applicable. Reclosing as not planned — if someone wants to pursue the AskUserQuestion case, a new issue would be cleaner.

PR #1420, opened 2026-03-17
feat(hooks): fire PreToolUse hooks before approval UI for ExitPlanMode

backnotprop:
**Status: Proof of Concept. Not necessarily merge-ready.**

## Summary

- Runs `PreToolUse` hooks when an `ExitPlanMode` approval request arrives, **before** the plan approval dialog renders
- Hook exit 2 (block): rejects the plan back to the agent with feedback, user never sees the dialog
- Hook exit 0 (allow): auto-approves the plan, skips the dialog entirely
- No matching hooks: falls through to the existing approval UI unchanged

Refs: #1419

## What works
- Blocking hooks (exit 2) successfully reject plans before the user sees any dialog
- Allowing hooks (exit 0) successfully auto-approve plans without user interaction
- No hooks configured: behavior is identical to before
- Tested with both command hooks and confirmed via hook logs

## What to consider before merging
- **Async in useEffect**: The hook runs async inside the guard `useEffect`. The guard returns early to prevent the dialog from rendering while the hook is in flight. This works in practice but relies on React not rendering the approval UI during the await window.
- **Scoped to ExitPlanMode only**: This doesn't address `EnterPlanMode` or `AskUserQuestion`. A more general solution would apply pre-approval hooks to all interactive approval tools.
- **The dialog may flash briefly** in edge cases before the hook resolves, though this wasn't observed in testing.

## Test plan

- [ ] Configure a `PreToolUse` hook with matcher `ExitPlanMode` that exits 2. Verify plan is rejected without showing dialog
- [ ] Configure a `PreToolUse` hook with matcher `ExitPlanMode` that exits 0. Verify plan is auto-approved without dialog
- [ ] Remove all ExitPlanMode hooks. Verify normal approval dialog works unchanged
- [ ] Test in YOLO mode. Verify existing auto-approve behavior still works
- [ ] Test plan mode expiry (CLI restart). Verify existing denial behavior still works

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -84,6 +84,7 @@ import {
   runPreCompactHooks,
   runSessionEndHooks,
   runSessionStartHooks,
+  runPreToolUseHooks,
   runStopHooks,
   runUserPromptSubmitHooks,
 } from "../hooks";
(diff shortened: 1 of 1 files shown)


devanshrj:
Closing as obsolete — plan mode was removed entirely in #2413, and its successor goal mode was also removed in #2792. `ExitPlanMode` no longer exists in the codebase (the only remaining references are in a docs file). The linked issue #1419 was already closed as not planned.

The underlying idea (firing `PreToolUse` hooks before the approval UI for interactive tools) could still be relevant for `AskUserQuestion`, but this PR as written targets a tool that no longer exists and has merge conflicts.

PR #1435, opened 2026-03-18
feat: add secret command

AriWebb:


src/cli/commands/registry.ts
```diff
@@ -252,6 +254,15 @@ export const commands: Record<string, Command> = {
       return "Opening MCP server manager...";
     },
   },
+  "/secret": {
+    desc: "Manage secrets for shell commands",
+    order: 33,
+    args: "<set|list|delete> [key] [value]",
```
kianjones9: maybe `delete` -> `unset`?

src/cli/commands/secret.ts
```diff
@@ -0,0 +1,121 @@
...
+      const normalizedKey = key.toUpperCase();
+
+      // Validate key format (must be valid for $SECRET_NAME pattern)
+      if (!/^[A-Z_][A-Z0-9_]*$/.test(normalizedKey)) {
+        return {
+          output: `Invalid secret name '${key}'. Use uppercase letters, numbers, and underscores only. Must start with a letter or underscore.`,
+        };
+      }
+
+      const value = valueParts.join(" ");
```
kianjones9: what is this? we reconstitute secret value chunks?

src/cli/commands/secret.ts
```diff
@@ -0,0 +1,121 @@
...
+      setSecret(normalizedKey, value);
+
+      return {
```
kianjones9: We should also add a system reminder to the harness so that the agent knows on the next user message the value is now available to it.

src/cli/commands/secret.ts
```diff
@@ -0,0 +1,121 @@
...
+        return { output: "Usage: /secret set KEY value" };
+      }
+      if (valueParts.length === 0) {
+        return {
+          output:
+            "Usage: /secret set KEY value\nProvide a value for the secret.",
+        };
+      }
+
+      const normalizedKey = key.toUpperCase();
```
kianjones9, line 41: why does it have to be upper case? env vars support lowercase too I think right?

src/cli/commands/secret.ts
```diff
@@ -0,0 +1,121 @@
...
+    case "":
+    case "help": {
+      return {
+        output: `Secret management commands:
+
+  /secret set KEY value   Set a secret (KEY is normalized to uppercase)
+  /secret list            List available secret names
+  /secret delete KEY      Delete a secret
+
+Secrets are stored in ~/.letta/secrets.json
```
kianjones9: I'm wondering about the robustness of a json file here vs something like a plaintext .env file. Seems like if we use json we need concern ourselves with escape characters and control characters, etc. vs .env can be more easily consumed by callers generally. i.e. a docker container or cloud sandbox environment could just `source .env` rather than require some custom json to env parser? not feeling strong on this either way tho so don't let this block.

src/cli/components/PasteAwareTextInput.tsx
```diff
@@ -456,6 +456,19 @@ export function PasteAwareTextInput({
       caretOffsetRef.current = nextCaret;
     };
 
+    // Move cursor to beginning of input (Home key)
```
kianjones9, line 459: did you mean to bundle this into this PR?

kianjones9: that's totally fine just checking haha

src/utils/secretsStore.ts
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Get the path to the secrets file.
+ */
+export function getSecretsPath(): string {
+  return join(homedir(), ".letta", "secrets.json");
+}
+
+/**
+ * Load secrets from disk (cached in memory).
```
kianjones9: does `/secrets set` in one session trigger read from disk in another session?

kianjones9: or maybe it doesn't matter because the SSOT is the memfs file?

src/cli/components/PasteAwareTextInput.tsx
AriWebb: yeah I did haha qol improvement I've been wanting.

src/cli/commands/registry.ts
AriWebb: done

src/cli/commands/secret.ts
AriWebb: fixed with server side

src/utils/secretsStore.ts
AriWebb: fixed with server side.

PR #1543, opened 2026-03-26
fix: don't have file for secrets

AriWebb:


src/agent/memoryFilesystem.ts
```diff
@@ -355,12 +356,26 @@ export async function applyMemfsFlags(
       pullSummary = result.summary;
     }
 
-    // Fetch secrets from the server so they're available for $SECRET_NAME substitution.
-    const { initSecretsFromServer } = await import("../utils/secretsStore");
+    // Clean up legacy secrets.md file that was previously written by secretsStore.
```
sarahwooders: maybe we can skip the cleanup? I'm concerned about having this code in here perpetually, since what if people actually add a file that matches this name? I think it will naturally get cleaned up with `/doctor`

AriWebb: ah ok, sounds good. I'll take this out

PR #1586, opened 2026-03-30
fix: (more) correct error for access token failure

ThatOneCalculator:
`letta login` isn't a subcommand that exists
ref:
<https://discord.com/channels/1161736243340640419/1487441939837095999>

### DIFF
--- a/src/agent/client.ts
+++ b/src/agent/client.ts
@@ -168,7 +168,7 @@ export async function getClient() {
         });
         console.error("Failed to refresh access token:", error);
         console.error(
-          "\nPlease move ~/.letta to something like ~/.letta_backup, and re-run 'letta' to re-authenticate",
+          "\nIf you experience this issue multiple times, move ~/.letta to ~/.letta_backup, and re-run 'letta' to re-authenticate",
         );
         process.exit(1);
       }


src/agent/client.ts
```diff
@@ -167,7 +167,9 @@ export async function getClient() {
           context: "auth_client_token_refresh",
         });
         console.error("Failed to refresh access token:", error);
-        console.error("Please run 'letta login' to re-authenticate");
```
4shub, line 170: Keep this!

src/agent/client.ts
```diff
@@ -167,7 +167,9 @@ export async function getClient() {
...
+        console.error(
+          "\nPlease move ~/.letta to something like ~/.letta_backup, and re-run 'letta' to re-authenticate",
```
4shub: "If you experience this issue multiple times, try to move ~/.letta to ~/.letta_backup, and re-run 'letta' to re-authenticate"

src/agent/client.ts
ThatOneCalculator: But `letta login` isn't a sub-command that exists afaik...

ThatOneCalculator: ref (lol):

image

PR #1589, opened 2026-03-30
Feat/add edit file from client

shelleypham:


src/websocket/listener/client.ts
```diff
@@ -2313,6 +2329,44 @@ async function connectWithRetry(
...
 
+    // ── File writing (no runtime scope required) ──────────────────────
+    if (isWriteFileCommand(parsed)) {
+      console.log(
+        `[Listen] Received write_file command: path=${parsed.path}, request_id=${parsed.request_id}`,
+      );
+      void (async () => {
+        try {
+          const { writeFile } = await import("node:fs/promises");
+          await writeFile(parsed.path, parsed.content, "utf-8");
```
4shub, line 2340: is it possible in the follow up PR to use the `edit` function from `edit.ts`

PR #1595, opened 2026-03-31
fix(listen): task notifications queuing

christinatong01:
emit task notifications as a separate type in queue (to not get merged with user message)
add XML tags

### DIFF
--- a/src/websocket/listener/queue.ts
+++ b/src/websocket/listener/queue.ts
@@ -68,6 +68,10 @@ function mergeDequeuedBatchContent(
   const queuedInputs: Array<
     | { kind: "user"; content: MessageCreate["content"] }
     | {
+        kind: "task_notification";
+        text: string;
+      }
+    | {
         kind: "cron_prompt";
         text: string;
       }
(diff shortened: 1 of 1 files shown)


src/websocket/listener/turn.ts
```diff
@@ -187,7 +187,7 @@ function buildMaybeLaunchReflectionSubagent(params: {
 
           addToMessageQueue({
             kind: "task_notification",
-            text: msg,
+            text: `<task-notification><summary>${msg}</summary></task-notification>`,
```
cpacker, line 190: Do we need to <task-notification> xml wrappers here if the type itself is `task_notification`?

christinatong01: It makes it easier for the FE to parse out task_notifications reliably

By the time desktop renders transcript messages, it only sees stream_delta entries, not the original queue item type (task_notification). So without an explicit marker like the xml wrappers, it can't tell that a message came from task_notification.

If we want zero XML, we should probably add a dedicated task_notification stream event (or add notification metadata to `stream_delta`). But idk if that's necessary yet since this works?

PR #1604, opened 2026-03-31
fix: expand tilde in device file command paths [LET-8267]

4shub:
Node.js fs APIs don't expand ~ like a shell does, so list_in_directory and read_file with paths like ~/.letta/plans would fail with ENOENT. Now resolves ~ to homedir() first.

🤖 Generated with [Letta Code](https://letta.com)


sarahwooders:

PR #1726, opened 2026-04-08
fix(telemetry): reduce stream error spam and telemetry overhead

cpacker:
## Summary
- emit a single terminal stream telemetry error only after mid-stream recovery settles, and allow resume even when no abort signal is present so headless flows don't spam `stream_resume_skipped`
- coalesce concurrent telemetry flushes and keep the current session's debug tail in memory so error tracking avoids overlapping POSTs and synchronous log-file reads
- estimate tool response sizes directly for telemetry instead of JSON-stringifying full multimodal tool returns, with focused regression coverage for the new telemetry behavior

### DIFF
--- a/src/cli/helpers/stream.ts
+++ b/src/cli/helpers/stream.ts
@@ -40,6 +40,19 @@ export type ApprovalRequest = {
   toolArgs: string;
 };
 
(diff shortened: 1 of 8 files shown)


sarahwooders:

PR #1737, opened 2026-04-09
feat: channels system with Telegram MVP (WS-only, listener mode)

cpacker:
## Summary

- Adds a **channels system** enabling agents to receive and respond to messages from external platforms, starting with Telegram
- Messages flow into the agent queue as XML-wrapped `kind: "message"` items with `source: "channel"`
- Agents reply via a new **MessageChannel** client-side tool
- Telegram adapter uses **grammY** with long-polling (no webhook/public URL needed)
- Includes CLI (`letta channels configure|status|route|pair`), `--channels` flag on `letta server`, and `/channels` WS command for live management from ADE
- Channels survive WS reconnects (pause/resume lifecycle — adapters keep polling, messages buffer)
- Atomic pairing with snapshot-based rollback on disk failure
- Includes tests for XML formatting, pairing, routing, registry lifecycle, Telegram adapter error handling, and protocol types

## Setup Instructions

Current recommended Telegram flow:

1. Configure Telegram:
   - `letta channels configure telegram`
2. Start the WS listener with Telegram enabled:
   - `letta server --channels telegram`
3. Message the bot from **Telegram** once to receive a pairing code.
4. In the target ADE/Desktop conversation, bind that Telegram chat to the current agent/conversation:
   - `/channels telegram pair <code>`
5. Continue chatting with the agent from Telegram.

Notes:
- `letta channels pair ...` and `letta channels route ...` also exist as standalone CLI commands, but they are file-oriented management commands. The recommended live path is the `/channels ...` WS command from the target conversation.
- Current listener startup is explicit: Telegram only starts when the listener is launched with `--channels telegram`.

## Data Storage

Persisted state lives under `~/.letta/channels/telegram/`:

- `config.yaml`
  - bot token
  - `enabled`
  - `dmPolicy`
  - `allowedUsers`
- `pairing.yaml`
  - pending pairing codes
  - approved Telegram users
- `routing.yaml`
  - Telegram `chat_id` -> Letta `agent_id` + `conversation_id` bindings

These files are currently **machine-scoped**, not agent-scoped.

### Tool Name

`MessageChannel`

### Description

> Send a message to an external channel (Telegram, Slack, etc.) in response to a channel notification.
>
> When you receive a `<channel-notification>`, use this tool to reply. Extract the `source` and `chat_id` from the notification attributes and pass them as `channel` and `chat_id`.
>
> Parameters:
> - `channel`: The platform to send to (matches the `source` attribute)
> - `chat_id`: The chat ID to send to (matches the `chat_id` attribute)
> - `text`: The message text to send
> - `reply_to_message_id`: (Optional) Reply to a specific message by its `message_id`

### Schema

```json
{
  "type": "object",
  "properties": {
    "channel": {
      "type": "string",
      "description": "The channel to send the message to (e.g., 'telegram')"
    },
    "chat_id": {
      "type": "string",
      "description": "The chat/conversation ID to send to (from the channel-notification attributes)"
    },
    "text": {
      "type": "string",
      "description": "The message text to send"
    },
    "reply_to_message_id": {
      "type": "string",
      "description": "Optional message ID to reply to"
    }
  },
  "required": ["channel", "chat_id", "text"],
  "additionalProperties": false
}
```

Runtime note:
- when channels are active, the `channel` field gets a runtime-injected enum limited to the currently active channel ids

### Exposure / Attachment Behavior

Important clarification for reviewers: in this MVP, `MessageChannel` is **not** attached only to agent/conversation pairs that already have a Telegram route.

Current behavior:
- if any channels are active in the listener process, `MessageChannel` is added to the prepared client toolset
- this is a **process-global exposure** decision, not per-route/per-conversation tool attachment
- the actual safety boundary is execution-time validation in the tool implementation:
  - the tool receives injected `parentScope` (`agentId` + `conversationId`)
  - it looks up the route for `channel + chat_id`
  - it only sends if that route belongs to the executing agent/conversation

So the MVP tradeoff is:
- broad tool availability when channels are enabled
- strict send authorization at execution time

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1,12 +1,12 @@
 {
   "lockfileVersion": 1,
-  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "@letta-ai/letta-code",
       "dependencies": {
         "@letta-ai/letta-client": "1.10.1",
         "glob": "^13.0.0",
+        "grammy": "^1.42.0",
         "highlight.js": "^11.11.1",
         "ink-link": "^5.0.0",
         "lowlight": "^3.3.0",
@@ -41,6 +41,8 @@
     "@alcalzone/ansi-tokenize": ["@alcalzone/ansi-tokenize@0.1.3", "", { "dependencies": { "ansi-styles": "^6.2.1", "is-fullwidth-code-point": "^4.0.0" } }, "sha512-3yWxPTq3UQ/FY9p1ErPxIyfT64elWaMvM9lIHnaqpyft63tkxodF5aUElYHrdisWve5cETkh1+KBw1yJuW0aRw=="],
 
     "@emnapi/runtime": ["@emnapi/runtime@1.8.1", "", { "dependencies": { "tslib": "^2.4.0" } }, "sha512-mehfKSMWjjNol8659Z8KxEMrdSJDDot5SXMq00dM8BN4o+CLNXQ0xH2V7EchNHV4RmbZLmmPdEaXZc5H2FXmDg=="],
+
+    "@grammyjs/types": ["@grammyjs/types@3.26.0", "", {}, "sha512-jlnyfxfev/2o68HlvAGRocAXgdPPX5QabG7jZlbqC2r9DZyWBfzTlg+nu3O3Fy4EhgLWu28hZ/8wr7DsNamP9A=="],
 
     "@img/colour": ["@img/colour@1.0.0", "", {}, "sha512-A5P/LfWGFSl6nsckYtjw9da+19jB8hkJ6ACTGcDfEJ0aE+l2n2El7dsVM7UVHZQ9s2lmYMWlrS21YLy2IR1LUw=="],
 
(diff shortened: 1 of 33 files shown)


just-cameron (changes requested):
## Review: feat: channels system with Telegram MVP

Overall this is a reasonable MVP scope and the architecture is clean -- separate modules for config, pairing, routing, registry, and adapters. The buffer-then-flush pattern for WS reconnects and atomic pairing with snapshot rollback are good design choices. 38 tests is solid coverage for a first pass.

### Blocking

**1. Custom YAML parser/serializer (`config.ts:44-157`)**

This is a hand-rolled YAML parser that handles "flat/shallow objects." It will break on edge cases (multi-line strings, comments in unexpected positions, nested objects beyond 1 level). The routing module uses JSON and the pairing module uses JSON, so the YAML-only approach for config is inconsistent.

Options:
- Use JSON like the other persistence files
- Pull in a real YAML library (e.g., `yaml` -- already used elsewhere in the Letta ecosystem)
- At minimum, document the subset of YAML that's supported and add tests for edge cases

**2. No error handling on `bot.start()` (`telegram/adapter.ts:175-179`)**

If Telegram long-polling fails (invalid token, network error after startup), the error is swallowed by grammY's internal handler. Need `bot.catch()` or equivalent to surface errors to the registry/user. Silent failures in a channel adapter are particularly bad because the user thinks messages are being received.

### Non-blocking concerns

**3. `MessageChannel` tool requires approval (`toolset.ts:210`)**

Every outbound message from the agent to a channel user requires human approval. For an automated chat system this makes the UX very clunky -- the user in ADE has to approve every reply. Is this intentional for the MVP? If so, it should be documented. If not, consider `requiresApproval: false` or making it configurable.

**4. `node-fetch` transitive dependency via grammY**

grammY pulls in `node-fetch@2.7.0` (CJS) plus `tr46`, `webidl-conversions`, `whatwg-url`. Node 18+ has native fetch. Not blocking but adds unnecessary weight. grammY supports native fetch -- check if there's a config option to avoid the polyfill.

**5. Global mutable singletons (`routing.ts:903`, pairing stores)**

`routesByKey` is a module-level `Map` with no synchronization. Tests use `clearAllRoutes()` to reset. Fine for single-threaded MVP but worth noting as tech debt -- concurrent `loadRoutes` calls from different adapters could race.

**6. Config stored per-machine, not per-agent (`~/.letta/channels/`)**

If two agents on the same machine use the same Telegram bot token, their configs collide. The routing table handles agent disambiguation (chatId -> agentId+conversationId), but the config (token, dmPolicy) is shared. Worth documenting this constraint.

### Minor

- `getRoute` re-exported as `getRouteFromStore` in registry adds indirection without clarity
- `/channels` WS command handler dynamically imports modules on every invocation -- fine for infrequent commands but could be static
- XML entity escaping in `xml.ts` -- confirm the agent's response parsing handles `&amp;`, `&lt;`, etc. correctly

PR #1738, opened 2026-04-09
fix: ExitPlanMode restores to default instead of memory mode

4shub:
When plan mode was entered from memory mode, approving the plan incorrectly restored to memory mode. Now always restores to default (manual approve) unless the user was in bypassPermissions/yolo.


cpacker (approved):

cpacker (commented):
@4shub do we also need to apply this to headless.ts / listen-client.ts?

PR #1753, opened 2026-04-10
fix(channels): send telegram reminders as content parts

cpacker:
## Summary
- send inbound channel reminders and notification XML as separate user-message content parts instead of one concatenated text blob
- thread `MessageCreate["content"]` through channel registry buffering and listener ingress instead of treating channel payloads as string-only XML
- update channel and listener tests to assert the reminder and `<channel-notification>` arrive as distinct text parts

## Why
UMI already knows how to strip pure `<system-reminder>...</system-reminder>` text parts, but concatenated channel payloads forced the reminder and the notification into a single blob.

Splitting them into separate content parts keeps the agent-facing semantics the same while letting downstream UIs hide the reminder without needing to parse mixed XML text.

## Validation
- `bun test src/tests/channels/xml.test.ts src/tests/websocket/listen-client-concurrency.test.ts`
- `bunx --bun @biomejs/biome@2.2.5 check src/channels/xml.ts src/channels/registry.ts src/websocket/listener/client.ts src/tests/channels/xml.test.ts src/tests/websocket/listen-client-concurrency.test.ts`
- `bun run lint`

### DIFF
--- a/src/channels/registry.ts
+++ b/src/channels/registry.ts
@@ -9,6 +9,7 @@
  * 4. Buffered messages flush through the registered onMessage handler
  */
 
+import type { MessageCreate } from "@letta-ai/letta-client/resources/agents/agents";
 import { readChannelConfig } from "./config";
 import {
   consumePairingCode,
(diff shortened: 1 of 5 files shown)


cpacker:
Superseded by #1756. This PR was merged into the wrong base branch (); #1756 carries the same reminder-content-parts change against .

cpacker:
Superseded by #1756. This PR was merged into the wrong base branch (`fix/externalize-grammy-build`); #1756 carries the same reminder-content-parts change against `main`.

PR #1780, opened 2026-04-13
feat(channels): add Telegram media and reaction parity

cpacker:
## Summary
- add Telegram inbound media normalization for photos, files, audio, video, voice notes, animations, stickers, and media groups
- add Telegram outbound media uploads through `MessageChannel action=\"upload-file\"`
- add Telegram reaction parity for both outbound `MessageChannel action=\"react\"` and inbound `message_reaction` updates
- tighten channel notification / formatting coverage, including XML entity decoding before outbound channel formatting

## Notes
- This supersedes the older pre-refactor Telegram media/reaction workstreams rather than trying to merge their stale diffs directly.
- Slack behavior is unchanged here; this PR brings Telegram up to the same shared tool surface for `send`, `react`, and `upload-file`.

## Validation
- `bun test src/tests/channels/telegram-adapter.test.ts src/tests/channels/message-channel.test.ts src/tests/channels/message-tool-schema.test.ts src/tests/channels/xml.test.ts`
- `bun run typecheck`
- `bunx --bun @biomejs/biome@2.2.5 check src/channels/telegram/adapter.ts src/channels/telegram/messageActions.ts src/channels/types.ts src/channels/xml.ts src/tools/descriptions/MessageChannel.md src/tests/channels/telegram-adapter.test.ts src/tests/channels/message-tool-schema.test.ts src/tests/channels/message-channel.test.ts src/tests/channels/xml.test.ts`

### DIFF
--- a/src/channels/telegram/adapter.ts
+++ b/src/channels/telegram/adapter.ts
@@ -2,9 +2,9 @@
  * Telegram channel adapter using grammY.
  *
  * Uses long-polling (no webhook setup needed).
- * Reference: lettabot src/channels/telegram.ts
  */
 
+import type { ReactionType, ReactionTypeEmoji } from "@grammyjs/types";
 import type { Bot as GrammYBot, Context as GrammYContext } from "grammy";
 import type {
   ChannelAdapter,
@@ -12,6 +12,14 @@ import type {
   OutboundChannelMessage,
   TelegramChannelAccount,
 } from "../types";
+import {
+  detectTelegramUploadMethod,
+  extractTelegramMessageText,
+  getTelegramSenderName,
+  resolveTelegramInboundAttachments,
+  TELEGRAM_MEDIA_GROUP_FLUSH_MS,
+  type TelegramLikeMessage,
+} from "./media";
 import { loadGrammyModule } from "./runtime";
 
 type TelegramBot = GrammYBot<GrammYContext>;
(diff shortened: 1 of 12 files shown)


just-cameron (commented):
## Review

**Overall**: Solid, comprehensive PR. Brings Telegram to full parity with Slack on media, reactions, and upload-file. The media.ts extraction is well-structured, and the media group batching with flush timer is the right approach.

### Issues

1. **`escapeXmlText` no longer escapes `>` in text nodes.** Wait, it does -- I misread. It drops quote escaping for text nodes, which is correct (quotes don't need escaping in XML text nodes, only in attributes). The split into `escapeXmlText` vs `escapeXmlAttribute` is a good fix.

2. **Media group flush timer race on stop.** `stop()` clears `bufferedMediaGroups` and their timers, but if `emitInboundMessages` is already mid-flight from a timer callback, it could call `adapter.onMessage` after `stop()` has returned. Not a crash risk (just a late delivery), but worth noting.

3. **`TELEGRAM_MEDIA_GROUP_FLUSH_MS` (150ms)** -- tight window. Telegram can sometimes deliver album messages with gaps up to ~500ms. Might want to bump this or make it configurable. Though 150ms probably works for most cases.

4. **`downloadTelegramAttachment` silently returns null on oversize files.** No logging. If a user sends a 25MB video and the agent just sees no attachment, there's no indication why. A `console.warn` like the old adapter had would help debugging.

5. **Token in download URL.** `buildTelegramFileUrl` constructs the URL with the token embedded. The old adapter had an explicit comment about not logging this. The new code passes the URL to `fetchTelegramFile` which doesn't log it, but if the fetch throws with the URL in the error message, it could leak. The `resolveTelegramInboundAttachments` wraps each download in `.catch(() => null)` which swallows the error entirely. Silent on both the token leak AND the failure reason. Consider catching, logging `"Failed to download attachment: <fileId>"` without the URL, and returning null.

6. **`decodeBasicXmlEntities` in MessageChannel.ts.** Applied unconditionally to all outbound text, not just Telegram. This means Slack and Discord text that legitimately contains `&amp;` or `&lt;` (e.g., user typed literal HTML entities) would get decoded. Is this intentional? The commit message says "XML entity decoding before outbound channel formatting" but the context suggests this is needed because the agent's response text comes through XML-escaped. If so, it should only apply when the text came from an XML-encoded source.

7. **`parseTelegramReactionInput` casts to `ReactionTypeEmoji["emoji"]`.** This trusts that any string the agent passes is a valid Telegram reaction emoji. Telegram's API will reject invalid emoji with a 400 error. Not a code bug per se, but the error path should surface clearly to the agent.

### Nits

- `coerceSizeBytes` handles `undefined` and non-finite values, but `file_size` from Telegram is always `number | undefined`. The NaN/Infinity check is defensive but unnecessary.
- `inferMimeTypeFromName` could be a simple Map lookup instead of chained if-else.
- The `handleAction` in `messageActions.ts` now handles `"react"` and `"upload-file"` but the guard check at the top (`request.action !== "send" && request.action !== "react" && request.action !== "upload-file"`) will need updating every time a new action is added. Consider a Set or using `describeMessageTool().actions.includes()`.
- The xml.test.ts changes expect `action="react"` in the reminder text. This ties the test to the specific wording rather than just checking for the presence of reaction guidance.

### Conflict with #1772

This PR supersedes the image handling in #1772 (`feat/telegram-image-support`). The approach here is better -- uses `ChannelMessageAttachment` with `localPath` instead of a separate `InboundImageAttachment` type with base64. Suggest closing #1772 after this merges, or rebasing #1772 to just keep the temp-dir save + system-reminder-path changes if those aren't covered here.

Actually, checking the diff: this PR already saves to disk via `saveTelegramAttachment` and includes `localPath` in the attachment XML. It also keeps `imageDataBase64` for small images. So #1772 is fully superseded.

PR #1798, opened 2026-04-14
fix(memory): bootstrap memfs on refresh

cpacker:
## Summary
- make `list_memory` bootstrap memfs when the repo is missing instead of only returning `memfs_initialized: false`
- keep the explicit `enable_memfs` flow intact while reusing the same listener-side helper path
- add a websocket regression test covering refresh-on-first-use behavior

## Verification
- bun test src/tests/websocket/listen-client-protocol.test.ts
- bun run typecheck
- bun run lint (repo still has 2 unrelated pre-existing warnings)

### DIFF
--- a/src/tests/websocket/listen-client-protocol.test.ts
+++ b/src/tests/websocket/listen-client-protocol.test.ts
@@ -1085,6 +1085,67 @@ describe("listen-client cron command handling", () => {
   });
 });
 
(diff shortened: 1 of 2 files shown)


cpacker:
Superseding this with a two-part fix: explicit desktop memfs enable UI plus a backend-only local checkout self-heal path that does not silently mutate agent config from refresh.

PR #1865, opened 2026-04-21
feat: forward approval reason into tool return for agent visibility

4shub:
When a user approves a tool with a reason/message (e.g. "use worktree" from ExitPlanMode), the reason was stored on the ApprovalResult but never injected into the tool_return text the agent sees. The SDK's ToolReturn type doesn't have a reason field, so the server drops it.

Now appendReasonToToolReturn() appends the approval reason to the tool_return content (string or multimodal) so the agent actually receives the instruction.


cpacker:
see 11063

PR #1895, opened 2026-04-23
feat(channels): expose telegram transcribe_voice on ws listener protocol

just-cameron:
## Summary

Plumbs the per-account `transcribe_voice` setting (#1806) through the websocket listener so desktop UIs can read and write it. Server-side service already stores and honors it; this patches the wire protocol layer that was silently dropping the field.

## Changes

- **`protocol_v2.ts`**: adds `transcribe_voice: boolean` to the telegram `ChannelAccountSnapshot` and to the telegram branch of `ChannelAccountCreatePayload` / `ChannelAccountUpdateCommand.patch`.
- **`protocol-inbound.ts`**: validates that `transcribe_voice` is a boolean when present on telegram create/update commands (permissive — unknown fields still allowed).
- **`client.ts`**: `mapChannelAccount(telegram)` returns `transcribe_voice`; the `channel_account_create` and `channel_account_update` handlers forward the field to the service-layer `transcribeVoice` patch.
- **Regression test** (`listen-channels-transcribe-voice.test.ts`): outbound snapshot, inbound create, inbound update, and omit-to-preserve-existing.

## Why

Today there's no way to toggle Telegram voice-memo transcription from the desktop UI — users have to hand-edit `~/.letta/channels/telegram/accounts.json`. This PR unblocks a matching Letta Code Desktop PR that adds a toggle to the Telegram bot settings dialog.

## Test plan

- [x] `pnpm check` (biome + tsc) clean
- [x] `bun test src/tests/websocket/listen-channels-transcribe-voice.test.ts` — 4/4 pass
- [x] Full `src/tests/channels` + listener suites — no new failures (two pre-existing `channel-runtime` failures are unrelated, caused by a locally-installed telegram runtime defeating test isolation; same failures on origin/main)

"Context is that which is scarce." — Tyler Cowen

### DIFF
--- a/src/tests/channels/discord-protocol.test.ts
+++ b/src/tests/channels/discord-protocol.test.ts
@@ -70,14 +70,14 @@ describe("discord protocol-inbound validators", () => {
     expect(isChannelAccountCreateCommand(msg)).toBe(false);
   });
 
-  test("discord account create rejects legacy top-level plugin fields", () => {
+  test("discord account create accepts top-level token field", () => {
     const msg = {
       type: "channel_account_create",
       channel_id: "discord",
       request_id: "r1",
       account: { token: "test-token" },
     };
-    expect(isChannelAccountCreateCommand(msg)).toBe(false);
+    expect(isChannelAccountCreateCommand(msg)).toBe(true);
   });
 
   test("valid discord account update passes", () => {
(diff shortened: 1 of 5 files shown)


cpacker:
Closing this as obsolete.

The original goal was to expose Telegram transcribe_voice over the websocket protocol as a top-level typed field. Current main now exposes channel settings through the plugin-owned config payload, and transcribe_voice is already represented there for Telegram/Discord/WhatsApp-style channel configs. Keeping a separate top-level protocol branch would be the wrong direction now.

PR #1906, opened 2026-04-23
fix(permissions): restore approval flow for targeted cross-agent memfs reads

cpacker:
## Summary
- downgrade single-agent read-only cross-agent memfs access from an unbypassable deny to the normal approval flow, so explicit debugging reads can be approved or covered by saved rules
- keep writes, broad agents-tree enumeration, ancestor scans, and multi-agent targets hard-denied
- add regression coverage for the new ask-vs-deny split and for persisted/session allow rules authorizing targeted cross-agent reads

## Test plan
- [x] `bun test src/tests/permissions/crossAgentGuard.test.ts`
- [x] `bun run check`


devanshrj:
@carenthomas @cpacker i am fixing this in #2423, gonna close this one

PR #1951, opened 2026-04-24
feat: Add idle-time reflection sweep + shared launcher + message-id cursor

devanshrj:
Replaces the single mutually-exclusive `memoryReminderInterval` mode with a split model that handles active and inactive conversations via separate paths.

### Active trigger — per conversation, mutually exclusive
- Modes: `off`, `step-count`, `compaction-event` (same set as before).
- Step-count cadence now fires when `turns_since_last_successful_reflection >= activeStepCount` instead of `turnCount % activeStepCount`. The old modulo predicate fired on absolute turn numbers regardless of whether anything had been reflected since; the new counter resets to zero only on successful reflection (not on failure or no-delta skip).

### Passive idle-time sweep — agent-wide, runs alongside any active mode
- Periodically scans local transcripts under `~/.letta/transcripts/{agentId}/`, picks conversations with enough unreflected turns and a long-enough idle window, and reflects them sequentially.
- Defaults: 24h sweep interval, 12h conversation idle threshold, 8 unreflected turns. Candidates restricted to conversations with local transcript history.
- One aggregate task-notification per sweep, not one per reflection.

### Settings — scope-aware names

Four passive-reflection settings split cleanly into agent-scoped (the sweep job) and conversation-scoped (per-candidate filters):

| Setting | Scope | Default |
|---|---|---|
| `passiveSweepEnabled` | agent | `true` |
| `passiveSweepIntervalHours` | agent | `24` |
| `passiveConversationMinIdleHours` | conversation | `12` |
| `passiveConversationMinUnreflectedTurns` | conversation | `8` |

A conversation qualifies as a sweep candidate when it has been idle longer than the conversation idle threshold AND has at least the minimum unreflected turns. The sweep itself runs at most once per `passiveSweepIntervalHours` per agent.

### Shared launcher + per-agent serialization
- All three trigger sources (step-count, compaction-event, idle-time) go through `launchReflectionSubagent` in `cli/helpers/autoReflection.ts`.
- Per-parent-agent priority queue (`manual < active < idle`) ensures at most one reflection subagent writes a given agent's memory at a time, even when triggers fire concurrently across the TUI and the websocket listener.
- File lock (`utils/fileLock.ts`) covers same-machine multi-process state writes.

### Message-id cursor
- Transcript state moves from line-offset (`auto_cursor_line`) to message-id (`reflected_through_message_id`), keyed off the same canonical backend message ids that all trigger sources see. Lazy migration on first read; no startup scan.
- New schema (`schema_version: "v2_message_id"`): `total_completed_turns`, `reflected_completed_turns`, `turns_since_last_successful_reflection`, last-reflection metadata.

### Protocol + UI
- Websocket protocol carries both legacy and new fields; `SleeptimeSelector`, status line, and headless flag parsing are updated.
- Telemetry trigger source enum gains `idle-time`; sweep candidate count and duration are tracked.

## Backward compatibility

- Existing `memoryReminderInterval` values map to active trigger + idle-sweep enabled (or `off` → both off).
- Existing transcript state files migrate on first touch.
- Active trigger remains mutually exclusive — no behavior change for users who already picked one.
- Persisted settings keys are double-written: new code reads `reflectionActiveTrigger ?? reflectionTrigger` (and the matching pair for step count), and writes both on every persist. Same pattern repeats at the in-memory type (`trigger`/`stepCount` aliases for `activeTrigger`/`activeStepCount`) and the websocket protocol (`trigger`/`step_count` alongside `active_trigger`/`active_step_count`). Old clients keep working, downgrade is safe, and old settings.json files load correctly.

## Test plan

- [x] `bun test src/tests/cli/{auto-reflection,idle-reflection-sweep,memoryReminder,reflection-transcript,reflection-scope-gate,statusline-payload}.test.ts`
- [x] `HOME=/tmp/letta-test-home bun test src/tests/websocket/listen-client-protocol.test.ts`
- [x] `bun run typecheck`
- [x] `bunx --bun @biomejs/biome@2.2.5 check --write` on changed files

## Follow-up

- #1991 wires these triggers into headless bidirectional mode (separate stacked PR).
- After a couple of releases, drop the legacy persisted keys (`reflectionTrigger`, `reflectionStepCount`) — both the writes in `persistReflectionSettingsForAgent` and the read fallback in `applyExplicitReflectionOverrides`. The in-memory `trigger`/`stepCount` aliases on `ReflectionSettings` and the websocket `trigger`/`step_count` fields can follow on the same cadence once no in-the-wild client reads them.


cpacker (commented):
potentially put a n-day gate (eg 3-day gate?) on reflecting?

devanshrj:
`message-id` change merged in #2381, will reopen other changes as needed

PR #1985, opened 2026-04-27
fix: support self-hosted channel listeners

just-cameron:
## Summary
- Adds a local channel listener path for non-Cloud `LETTA_BASE_URL` when channels are enabled.
- Skips Cloud environment registration while reusing the existing listener queue, approvals, and `MessageChannel` pipeline.
- Keeps Cloud listener registration behavior unchanged and improves the self-hosted no-channel error.

"Small bridges should not require a control tower."


cpacker:
Closing in favor of #1993

PR #2013, opened 2026-04-29
fix(tui): avoid NBSP cursor placeholder

### LINKED ISSUE #2008: Strange cursor character (Ā) rendered in CLI input
## Bug

When using the Letta CLI in Wezterm (Arch Linux, Wayland), a strange character appears at the cursor position in the text input — it renders as **Ā** (A with macron) inside a white/inverse box. This is distracting enough to make the CLI unusable.

## Environment

- Arch Linux + Wayland (Niri compositor)
- Wezterm (MesloLGS Nerd Font Mono, Tokyo Night theme)
- `TERM=xterm-256color`
- Bash with `set -o vi` (disabling vi mode does not fix it)
- Letta Code 0.24.6

## What I've tried

- `set +o vi` before launching — no effect
- Explicitly resetting cursor shape (`printf "\e[0 q"`) — no effect
- Changing Wezterm cursor style to `SteadyBar` — no effect
- Claude Code CLI does **not** have this issue in the same terminal

## Analysis

The CLI uses Ink's `ink-text-input` component, which renders its cursor via `chalk.inverse()` on the character at the cursor position (or `chalk.inverse(" ")` when at end of input). The Ā glyph appears to be how the inverse space character renders in this terminal/font combination. Other Ink-based CLIs (e.g., Claude Code) don't exhibit this, suggesting it may be specific to how Letta's Ink build or cursor rendering handles the inverse text.

## Expected

Cursor should render as a subtle block or bar, not as a visible Unicode character.

cpacker:
## Summary
- Replace NBSP-based fake cursor cells with an internal sentinel that survives `wrap-ansi` trimming.
- Convert the sentinel back to an inverse ASCII space after wrapping, avoiding visible NBSP glyphs in terminals like WezTerm while preserving wrapped-line cursor visibility.
- Add regression coverage for the sentinel behavior.

Fixes #2008

## Test plan
- [x] `bun test src/tests/cli/text-input-cursor-rendering.test.ts`
- [x] `bun run check`
- [x] `bun run build`

### DIFF
--- a/src/tests/cli/text-input-cursor-rendering.test.ts
+++ b/src/tests/cli/text-input-cursor-rendering.test.ts
@@ -1,0 +1,32 @@
+import { describe, expect, test } from "bun:test";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+
+const vendorPath = fileURLToPath(
+  new URL("../../../vendor/ink-text-input/build/index.js", import.meta.url),
+);
+
+describe("vendored ink-text-input cursor rendering", () => {
+  test("uses an internal sentinel instead of rendering NBSP cursor cells", () => {
+    const source = readFileSync(vendorPath, "utf8");
+
+    expect(source).toContain("CURSOR_SENTINEL");
+    expect(source).toContain("replaceAll(CURSOR_SENTINEL, chalk.inverse(' '))");
+    expect(source).not.toContain("chalk.inverse('\\u00A0')");
+  });
+
+  test("sentinel survives wrap-ansi trim before converting back to ASCII space", async () => {
+    const { default: wrapAnsi } = await import("wrap-ansi");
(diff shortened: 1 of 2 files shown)


rossim2i2:
Confirmed this fix resolves the NBSP cursor glyph issue on WezTerm + Arch Linux (Niri/Wayland).

**Environment:**
- WezTerm 64-bit, Wayland mode
- Arch Linux, Niri compositor
- Letta Code CLI 0.24.8

**Before patch:** Visible NBSP glyph character rendered in the input cursor position.

**After patch:** Cursor renders as a clean inverse space with no visible glyph. The sentinel approach correctly survives `wrap-ansi` trimming and converts back properly via the `Transform` component.

Tested by applying the diff to the vendored `ink-text-input/build/index.js` in both the desktop AppImage and the CLI's bundled `letta.js`. Both work correctly.

PR #2041, opened 2026-04-30
fix(channels): fail listener startup on channel errors

just-cameron:
## Summary
- Treat explicit `--channels` startup failures as fatal so the listener does not silently continue without requested channels.
- Aggregate channel startup failures with direct install guidance for missing runtimes.
- Create the listener session log before channel startup so early failures are captured.
- Bound Telegram `init()` and polling readiness waits so startup hangs fail loudly.

"Make the failure finite."

### DIFF
--- a/src/channels/registry.ts
+++ b/src/channels/registry.ts
@@ -1173,6 +1173,45 @@ export class ChannelRegistry {
 
 // ── Initialization ────────────────────────────────────────────────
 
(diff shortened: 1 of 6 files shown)


cpacker:
Closing this in favor of #2528, which ports the explicit channel-startup failure behavior and Telegram startup diagnostics/timeouts onto current main.

PR #2044, opened 2026-04-30
fix(cli): split streamed reasoning blocks on new otid

jnjpng:
## Status

This `letta-code` PR contains a client-side defense, but the root cause for the same-`otid` section-break issue turned out to be server-side streaming. The Python core and TS provider paths were dropping the `\n\n` separator between OpenAI reasoning summary parts, so the CLI was appending exactly what it received.

## What This PR Still Does
- split streamed assistant/reasoning blocks when a fresh `otid` arrives for an existing `message.id`
- preserve the existing mixed `id`/`otid` canonicalization behavior for normal continuation cases
- add regressions covering the new-`otid` split case for both reasoning and assistant output

## Why This Is No Longer The Primary Fix
The trace that reproduced `...calculations.Computing with Python` kept the same `otid`; the break was lost before the CLI saw it. Backfill rendered correctly because stored summarized reasoning joins parts with `\n\n`, while live streaming was dropping that separator upstream.

## Upstream Fix
The server-side PR:
- fixes the Python core off-by-one so follow-up summary parts get a leading separator
- updates `codex.ts` to emit separator deltas on `response.reasoning_summary_part.added`
- updates `openai.ts` to handle `response.reasoning_summary_part.added` the same way

## Testing
Client-side tests in this PR:
- `bun test letta-code/src/tests/cli/accumulator-usage.test.ts`
- `bun test src/tests/cli/*.test.ts src/tests/turn-recovery-policy.test.ts`
- pre-commit hook: `tsc --noEmit`

Server-side tests in `letta-cloud` PR #11337:
- `cd /Users/jinjpeng/letta/letta-cloud/apps/core && uv run pytest tests/test_openai_ws_session.py -k separator_for_follow_up_reasoning_summary_part`
- `cd /Users/jinjpeng/letta/letta-cloud && CORE_DATABASE_URL=postgresql://user:pass@localhost:5432/testdb npx jest --runInBand libs/tmp-cloud-api-router/src/lib/core-runtime/llm/providers/codex.test.ts libs/tmp-cloud-api-router/src/lib/core-runtime/llm/providers/openai.test.ts`

### DIFF
--- a/src/cli/helpers/accumulator.ts
+++ b/src/cli/helpers/accumulator.ts
@@ -491,6 +491,19 @@ function resolveLineIdForKind(
   return `${kind}:${canonicalId}`;
 }
 
+function hasExistingOtidAlias(
+  aliasMap: Map<string, string>,
+  canonical: string,
+  nextOtid?: string,
+): boolean {
+  for (const [mappedOtid, mappedCanonical] of aliasMap.entries()) {
+    if (mappedCanonical === canonical && mappedOtid !== nextOtid) {
+      return true;
+    }
+  }
+  return false;
+}
+
 function resolveAssistantLineId(
   b: Buffers,
   chunk: LettaStreamingResponse & { id?: string; otid?: string },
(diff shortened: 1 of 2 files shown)


jnjpng:
image

PR #2045, opened 2026-04-30
feat(channels): add channelPolicy, autoThreadOnMention, inboundDebounceMs to Discord accounts

tanner-caffrey:
## Summary

Mainly aligning the discord channel for closer functional parity with the slack channel and discord functionality in lettabot.

All three Discord behaviors added here have Slack analogues, Discord was missing either the configuration knob, the implementation, or both:

- **`channelPolicy: "mention" | "open"`** — Slack already gates channel responses on `app_mention`, which is functionally equivalent to `"mention"` mode (Discord's new default). The new `"open"` mode goes a step further than Slack: respond to every guild message in allowed channels without requiring an @-mention. Useful for dedicated single-purpose channels.
- **`autoThreadOnMention: boolean`** — Slack already auto-threads channel responses (hardcoded via `thread_ts: event.ts` in the inbound→outbound flow). Discord's adapter does the equivalent today by calling `message.startThread()` on every mention. What's new here is *the opt-out config field* — Discord users can now keep mention-triggered conversations inline in the parent channel; Slack does not expose this opt-out.
- **`inboundDebounceMs: number`** — direct port of the merged Slack precedent (#1891). Short bursts of messages from the same sender stack into a single dispatch (text newline-joined, `isMention` OR-folded). Disabled by default. Reuses the channel-agnostic `inboundDebounce.ts` helper that #1891 introduced.

All three fields are optional and backward-compatible — accounts created before this PR keep today's behavior.

## What changed

**New fields on `DiscordChannelAccount`** (`src/channels/types.ts`)
- `channelPolicy: "mention" | "open"` — gate for guild channel messages. `"mention"` (default) preserves the legacy behavior of only responding when @-mentioned or already in a thread; `"open"` processes every guild message that passes other gates (`allowedChannels`, bot-author check). DMs are unaffected — they're still gated by `dmPolicy`.
- `autoThreadOnMention: boolean` — when @-mentioned outside a thread, controls whether the bot auto-creates a Discord thread for the conversation. Default `true` matches the legacy Discord behavior and Slack's hardcoded auto-threading. Set `false` to keep mention-triggered conversations in the parent channel.
- `inboundDebounceMs: number` — trailing-edge debounce window in ms. Default `0` (disabled). Messages with attachments or reactions bypass the debounce so file uploads aren't delayed.

**Channel-policy gate** (`src/channels/discord/channelPolicy.ts`, new)
- Pure function `shouldProcessGuildMessage({ isThread, wasMentioned, channelPolicy })`. Composes cleanly with the existing `isDiscordGuildChannelAllowed`.

**Adapter wiring** (`src/channels/discord/adapter.ts`)
- `messageCreate` handler routes guild messages through the new gate; `autoThreadOnMention` is honored at the existing thread-creation site (still requires `wasMentioned`).
- Inbound debouncer wraps both DM and guild dispatch sites. New exported helpers: `buildDiscordDebounceKey`, `mergeDiscordDebouncedEntries`, `resolveDiscordInboundDebounceMs` (env override `LETTA_DISCORD_INBOUND_DEBOUNCE_MS` → config field → 0).
- Reactions still go through the direct dispatch path; they bypass the debouncer entirely.
- Debounce key shape mirrors Slack's: `discord:${accountId}:${channelId}[:thread:${threadId}]:${senderId}`. Per-sender bucketing — different speakers in the same channel produce distinct dispatches, just like Slack.

**Codec / pluginTypes / service plumbing** (`src/channels/discord/accountConfig.ts`, `src/channels/pluginTypes.ts`, `src/channels/service.ts`)
- `channelPolicy` and `autoThreadOnMention` are accepted and serialized through the websocket protocol (`channel_policy` / `auto_thread_on_mention`), surfaced top-level on `ChannelAccountSnapshot` / `ChannelConfigSnapshot`, and threaded through `createAccountFromPatch`, `mergeAccountPatch`, and `setChannelConfigLive`. Matches the Telegram `transcribeVoice` precedent.
- `inboundDebounceMs` intentionally skips this layer (set per-account in `accounts.json` or via env var) — matches the Slack precedent in #1891.

**Setup wizard** (`src/channels/discord/setup.ts`)
- Two new prompts (channel policy, auto-thread-on-mention) added to the existing flow. Inbound debounce is a power-user knob, no prompt.

**Tests** (54 new across 6 files)
- `src/tests/discord-channel-policy.test.ts` — 10 tests for the gate
- `src/tests/discord-account-config.test.ts` — 14 tests for the codec
- `src/tests/channels/discord-service.test.ts` — round-trip through service layer
- `src/tests/channels/discord-protocol.test.ts` — protocol-validator tests (accept new fields, reject invalid `channel_policy` and non-boolean `auto_thread_on_mention`)
- `src/tests/websocket/listen-client-channel-accounts.test.ts` — end-to-end websocket round-trip (create → partial update → list → get-config)
- `src/tests/channels/discordDebounceKey.test.ts` — 23 tests for the debounce-key builder, env-override resolver, and merge helper

## Testing

- [x] `bun x tsc --noEmit` — clean
- [x] `bun test src/tests/channels/` — 287 pass, 0 fail (54 new)
- [x] Targeted suites: `bun test src/tests/channels/discord-*.test.ts src/tests/channels/discordDebounceKey.test.ts src/tests/discord-*.test.ts src/tests/websocket/listen-client-channel-accounts.test.ts` — all pass
- [x] Manual repro through `createInboundDebouncer`: 3-message burst → 1 merged dispatch (text newline-joined, `isMention` OR-folded, last-entry IDs); two distinct threads in same channel → 2 dispatches, each merged independently
- [x] Talking to agents on discord using the updated discord channel

## How to opt in

```jsonc
// ~/.letta/channels/discord/accounts.json
{
  "accounts": [
    {
      "channel": "discord",
      "channelPolicy": "open",
      "autoThreadOnMention": false,
      "inboundDebounceMs": 1500
      // ... other fields
    }
  ]
}
```

`channelPolicy` and `autoThreadOnMention` can also be set through the websocket protocol (`channel_account_create` / `channel_account_update` / `channel_set_config`) under the `config` payload. `inboundDebounceMs` is power-user-only — set in `accounts.json` or override globally via `LETTA_DISCORD_INBOUND_DEBOUNCE_MS`.


just-cameron (commented):
Overall: good direction and well-tested at the helper/protocol layer, but I would not merge quite yet. There are two behavior issues worth fixing before this lands.

The basic shape is right: `channelPolicy` is the missing Discord analogue to Slack’s mention gate, `autoThreadOnMention` is a useful escape hatch, and the debounce implementation tracks the Slack precedent closely. Typecheck passes and the new targeted suites pass locally: 73/73.

That said, a couple of semantics are leaky.

### Blocking / should fix

1. **`autoThreadOnMention: false` creates a parent-channel route, but follow-up messages are dropped in `channelPolicy: "mention"` mode.**

   Flow today:

   - User mentions the bot in a normal guild channel.
   - `autoThreadOnMention === false`, so no thread is created.
   - Inbound message gets `chatId = parent channel`, `threadId = null`, `isMention = true`.
   - Registry creates a route for the parent channel.
   - A follow-up message in that same channel without an @mention is stopped by the adapter gate before route lookup:

   ```ts
   shouldProcessGuildMessage({
     isThread: false,
     wasMentioned: false,
     channelPolicy: "mention",
   }) // false
   ```

   So “keep mention-triggered conversations inline in the parent channel” only keeps the first turn inline. It does not let the conversation continue inline unless every follow-up re-mentions the bot.

   Possible fixes:
   - Document this explicitly and maybe rename the option mentally as “do not auto-thread mentions”, not “inline conversation”.
   - Or make `autoThreadOnMention: false` imply route-aware continuation. That requires moving part of the “existing route” check before the adapter gate, or adding a cheap route-exists check to the adapter/registry boundary.
   - Or only recommend/use `autoThreadOnMention: false` with `channelPolicy: "open"`.

   I’d prefer the first or third for this PR, because route-aware adapter gating is a layering change.

2. **`channelPolicy: "open"` with no bound agent can spam every visible guild channel.**

   With open mode, every guild message that passes the allowlist reaches `ensureDiscordRoute`. If `config.agentId` is missing, `ensureDiscordRoute` sends:

   ```ts
   "This Discord bot isn't connected to a Letta agent yet..."
   ```

   That means an unbound open-mode bot can reply to every message in every channel it can see, unless `allowedChannels` is configured. The setup warning currently says:

   ```ts
   "open/allowlist DMs and guild @mentions won't route until you bind an agent."
   ```

   That is no longer accurate for `channelPolicy: "open"`: it is not just guild @mentions.

   Suggested fix:
   - If `channelPolicy === "open"`, require an agent binding during setup, or at least print a much stronger warning.
   - In `ensureDiscordRoute`, consider only sending the unbound-agent reply for DMs and explicit mentions. For open-mode non-mentions, silently drop until bound. That avoids accidental channel spam.

3. **Snapshots expose undefined for defaulted fields.**

   `accountConfig.toAccountConfig()` serializes defaults:

   ```ts
   channel_policy: account.channelPolicy ?? "mention"
   auto_thread_on_mention: account.autoThreadOnMention ?? true
   ```

   But the top-level service snapshots do not:

   ```ts
   channelPolicy: account.channelPolicy,
   autoThreadOnMention: account.autoThreadOnMention,
   ```

   So protocol `config.channel_policy` says `"mention"` while top-level `channelPolicy` is `undefined` for accounts created without explicit values. Same for `autoThreadOnMention`.

   Existing service conventions seem to prefer effective defaults at snapshot time (`defaultPermissionMode ?? "default"`, `transcribeVoice === true`, `allowedChannels ?? []`). I’d make these effective too:

   ```ts
   channelPolicy: account.channelPolicy ?? "mention",
   autoThreadOnMention: account.autoThreadOnMention ?? true,
   ```

   And add a service test for default snapshots. This matters because the desktop UI is likely to read top-level fields.

### Non-blocking

4. **Wizard bloat is now real.**

   The Discord setup wizard is now token → DM policy → conditional allowlist → channel policy → auto-thread → agent binding. I’m still okay with prompting for `channelPolicy`; it materially changes visible behavior. I’m less convinced `autoThreadOnMention` deserves a setup prompt. It is probably a manage/settings knob.

5. **Squash the two commits.**

   `8c2f9805` fixes a functional hole in `6a7cbd07`, so they should land as one coherent feature commit.

6. **Docs follow-up needed.**

   The docs we just merged cover `allowedChannels`, not `channelPolicy`, `autoThreadOnMention`, or `inboundDebounceMs`.

### Things I checked and liked

- Adapter gate composes correctly with `allowedChannels`.
- `channelPolicy: "open"` registry fix is necessary and correct.
- Debounce key shape is sane: account + channel + optional thread + sender.
- Debounced errors still use `merged.raw`, which remains the inbound message raw, not the debounce key raw. Good.
- Attachments bypass debounce, reactions bypass debounce entirely.
- `??` merge semantics preserve explicit `false` for `autoThreadOnMention`.

I’d ask for fixes to 1–3, then approve.

just-cameron:
Pushed `c40ade69` to this branch with the review fixes:

- Unbound `channelPolicy: "open"` Discord bots now silently drop ambient guild messages instead of replying to every visible channel message. DMs and explicit mentions still get the unbound-agent warning.
- Top-level Discord service snapshots now expose effective defaults: `channelPolicy: "mention"` and `autoThreadOnMention: true`.
- Comments/setup copy now clarify that `autoThreadOnMention: false` keeps the first response inline, but follow-up parent-channel messages still need another @mention in `"mention"` mode.
- Fixed the Biome formatting issue in `discord/accountConfig.ts`.

Verification:
- `bun test src/tests/channels/discord-registry.test.ts src/tests/channels/discord-service.test.ts src/tests/discord-channel-policy.test.ts src/tests/discord-account-config.test.ts src/tests/channels/discord-protocol.test.ts src/tests/channels/discordDebounceKey.test.ts src/tests/websocket/listen-client-channel-accounts.test.ts` → 81 pass.
- `bun run typecheck` → clean.
- `bun run lint` → clean.

just-cameron:
Pushed `c587fab5` with Discord audio transcription support:

- Adds `transcribe_voice` to Discord account config, protocol validation, service/config snapshots, and legacy config parsing.
- Threads `transcribeVoice` into the Discord adapter's attachment resolver.
- Transcribes inbound Discord audio attachments when `transcribeVoice` is enabled and `OPENAI_API_KEY` is configured. Discord voice messages arrive as ordinary audio attachments, so this applies to `audio/*` attachments rather than a separate voice-only object.
- Adds a setup wizard prompt matching Telegram's opt-in.
- Adds tests for config parsing/serialization, protocol round-trip, service snapshots, and media transcription.

Verification:
- `bun test src/tests/channels/discord-registry.test.ts src/tests/channels/discord-service.test.ts src/tests/channels/discord-media.test.ts src/tests/discord-channel-policy.test.ts src/tests/discord-account-config.test.ts src/tests/channels/discord-protocol.test.ts src/tests/channels/discordDebounceKey.test.ts src/tests/websocket/listen-client-channel-accounts.test.ts` → 89 pass.
- `bun run typecheck` → clean.
- `bun run lint` → clean.

PR link: https://github.com/letta-ai/letta-code/pull/2045

PR #2086, opened 2026-05-05
fix(tools): verify bundled ripgrep binary before using it

michaelBelsanti:
just checks if the binary exists before executing. uses rg from path if the binary doesnt exist. this fixes the search tool when npm packages were installed with "--ignore-scripts" which is used in the nix package from https://github.com/numtide/llm-agents.nix


michaelBelsanti:
no longer needed

PR #2087, opened 2026-05-05
fix(cli): handle additional kitty keyboard protocol keybinds

michaelBelsanti:
Fixes Ctrl+W/Left/Right when using kitty keyboard protocol.

### DIFF
--- a/src/cli/components/PasteAwareTextInput.tsx
+++ b/src/cli/components/PasteAwareTextInput.tsx
@@ -104,9 +104,13 @@ function findNextWordBoundary(text: string, cursorPos: number): number {
 type WordDirection = "left" | "right";
 
 // biome-ignore lint/suspicious/noControlCharactersInRegex: Terminal escape sequences require ESC control character
-const OPTION_LEFT_PATTERN = /^\u001b\[(?:1;)?(?:3|4|7|8|9)D$/;
+// Match word-navigation arrow sequences: ESC[1;modifier D/C
+// Modifier values: 2=Shift, 3=Alt, 4=Alt+Shift, 5=Ctrl, 6=Ctrl+Shift,
+//   7=Alt+Ctrl, 8=Alt+Ctrl+Shift, 9=Meta
+// All of these should trigger word navigation (move by word).
+const OPTION_LEFT_PATTERN = /^\u001b\[(?:1;)?(?:2|3|4|5|6|7|8|9)D$/;
 // biome-ignore lint/suspicious/noControlCharactersInRegex: Terminal escape sequences require ESC control character
-const OPTION_RIGHT_PATTERN = /^\u001b\[(?:1;)?(?:3|4|7|8|9)C$/;
+const OPTION_RIGHT_PATTERN = /^\u001b\[(?:1;)?(?:2|3|4|5|6|7|8|9)C$/;
 
 function detectOptionWordDirection(sequence: string): WordDirection | null {
   if (!sequence.startsWith("\u001b")) return null;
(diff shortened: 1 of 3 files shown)


src/cli/components/PasteAwareTextInput.tsx
```diff
@@ -107,9 +104,13 @@ function findNextWordBoundary(text: string, cursorPos: number): number {
...
 // biome-ignore lint/suspicious/noControlCharactersInRegex: Terminal escape sequences require ESC control character
-const OPTION_LEFT_PATTERN = /^\u001b\[(?:1;)?(?:3|4|7|8|9)D$/;
+// Match word-navigation arrow sequences: ESC[1;modifier D/C
+// Modifier values: 2=Shift, 3=Alt, 4=Alt+Shift, 5=Ctrl, 6=Ctrl+Shift,
+//   7=Alt+Ctrl, 8=Alt+Ctrl+Shift, 9=Meta
+// All of these should trigger word navigation (move by word).
+const OPTION_LEFT_PATTERN = /^\u001b\[(?:1;)?(?:2|3|4|5|6|7|8|9)D$/;
 // biome-ignore lint/suspicious/noControlCharactersInRegex: Terminal escape sequences require ESC control character
-const OPTION_RIGHT_PATTERN = /^\u001b\[(?:1;)?(?:3|4|7|8|9)C$/;
+const OPTION_RIGHT_PATTERN = /^\u001b\[(?:1;)?(?:2|3|4|5|6|7|8|9)C$/;
```
kianjones9: @michaelBelsanti did you mean to add 2 for shift here?

src/cli/components/PasteAwareTextInput.tsx
```diff
@@ -107,9 +104,13 @@ function findNextWordBoundary(text: string, cursorPos: number): number {
 type WordDirection = "left" | "right";
 
 // biome-ignore lint/suspicious/noControlCharactersInRegex: Terminal escape sequences require ESC control character
```
kianjones9, line 114: can you move this back down to the line above the regex?

kianjones9 (changes requested):

michaelBelsanti:
i made those fixes a long time ago but ig i renamed the branch at some point and never pushed them lol

kianjones9 (approved):

cpacker (changes requested):
Thanks for fixing these Kitty protocol shortcuts. I think the Ctrl+W half is safe, but I don’t think the Ctrl+Arrow change is safe as written.

With kitty protocol flag 1, `ESC[1;5D` / `ESC[1;5C` are already parsed by `vendor/ink/build/parse-keypress.js` as normal arrow keys with `key.ctrl = true` and `key.meta = false`. This PR also makes `PasteAwareTextInput` handle those same sequences in the raw input listener as word navigation. That means a single Ctrl+Left/Right keypress will be handled twice: first by the raw listener moving by word, then by `ink-text-input` treating the same event as a regular left/right arrow and moving one additional character.

End-user symptom: Ctrl+Arrow in Kitty/Ghostty/WezTerm can land one character off the word boundary (e.g. Ctrl+Left moves to the start of the previous word, then one extra character left). This is the same class of raw-handler + Ink double-processing issue that previously caused CSI-u regressions.

If we want the parent raw handler to own Ctrl+Arrow word navigation, we need to also update the vendored text input to skip Ctrl+Arrow the same way it currently skips Meta/Option+Arrow, and add regression coverage for Ctrl+Left/Ctrl+Right so they move exactly one word rather than word + one char.

michaelBelsanti:
Added a skip for when ctrl is held, as well as test coverage. I'm unsure exactly why, but I never saw that bug occur in the months I had been using the patch, but I see why it could. Thanks for the catch.

PR #2100, opened 2026-05-06
feat: add operator channel destinations

just-cameron:
"Every agent needs an ops line."

## Summary
- Add per-agent and per-conversation operator channel destinations persisted under `~/.letta/channels/operators.json`.
- Let `MessageChannel` default to the configured operator destination when no explicit channel/chat target is supplied.
- Forward terminal listener errors and optional retry notices to the operator destination, with WS and CLI management commands.


cpacker:
Closing this as stale.

The current MessageChannel and channel-plugin architecture has changed significantly since this operator-destination implementation. Reviving this branch would mean reworking most of the integration surface anyway, so if operator channels are still desired they should be redesigned against the current channel/plugin and MessageChannel code paths in a fresh PR.

PR #2122, opened 2026-05-06
feat(cli): client-side conversation titles + opt-in AI title generation

kianjones9:
## Summary
- derive auto conversation titles client-side from the first user message instead of calling the server-side generation endpoint (default behavior)
- add an opt-in `conversation_titles` experiment that, when enabled, runs title generation through a hidden fork of the active conversation:
  - fork the convo with `hidden: true` so it does not appear in listings
  - send a single `messages.create` turn against the fork with `override_model: letta/auto`, an `override_system` that constrains the model to title-only output, and `max_steps: 1`
  - collect `assistant_message` text from the stream, normalize, return
  - best-effort delete the forked convo on the way out so it does not pile up
  - 30s AbortController timeout, falls back to the heuristic on any failure / experiment off / local backend / `default` convo
- keep auto-title eligibility scoped to genuinely new conversations so resume, search, fork, and agent-switch flows do not accidentally retitle existing convos
- make `/rename {agent|convo} [name]` the documented shape, where bare `/rename convo` auto-generates a title (AI when experiment on, heuristic otherwise) and `/rename convo [name]` sets one manually
- both the bare `/rename convo` command and the post-first-turn auto path share `generateConversationTitle()`, so the AI path applies to both

## Test plan
- [x] `bunx tsc --noEmit`
- [x] `bun test src/tests/cli/commandRunner.test.ts`
- [x] `bun test src/tests/cli/conversationTitle.test.ts` (normalize edge cases)
- [x] `bun test src/tests/experiments/manager.test.ts` (experiment toggle persistence)
- [ ] manually verify first-turn auto-title behavior in the CLI with `conversation_titles` off (heuristic) and on (AI via fork)
- [ ] manually verify `/rename convo` and `/rename convo [name]` in the CLI in both modes
- [ ] verify hidden fork is deleted (or at least hidden) after title generation

## Example

https://github.com/user-attachments/assets/3f12cc96-3903-4913-ae7a-cd55e97ad486

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -9391,8 +9391,7 @@ export default function App({
               "",
               "USAGE",
               "  /rename agent <name>      — rename the agent",
-              "  /rename convo <title>     — set a manual conversation title",
-              "  /rename convo auto        — generate a conversation title",
+              "  /rename convo [title]     — set a title, or auto-generate when omitted",
               "  /rename help              — show this help",
             ].join("\n");
             cmd.finish(output, true);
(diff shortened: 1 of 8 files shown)


jnjpng (approved):

carenthomas:
> fork the convo with hidden: true so it does not appear in listings

ooh this is clever

carenthomas (approved):

PR #2130, opened 2026-05-07
fix(cli): make markdown message links clickable

kianjones9:
## Summary
- render inline markdown links in message content with terminal hyperlink support instead of plain text + parenthesized URLs
- keep the change scoped to the shared inline markdown renderer so assistant/tool/message markdown paths inherit the fix
- add a focused tokenizer test covering valid and malformed markdown links

## Test plan
- [x] `bun test src/tests/cli/inline-markdown-renderer.test.ts`
- [x] `bun test src/tests/cli/errorFormatter.test.ts`
- [ ] Click a markdown link in the CLI manually to verify terminal behavior

### DIFF
--- a/src/cli/components/InlineMarkdownRenderer.tsx
+++ b/src/cli/components/InlineMarkdownRenderer.tsx
@@ -1,4 +1,5 @@
 import type React from "react";
+import Link from "ink-link";
 import { colors } from "./colors.js";
 import { Text } from "./Text";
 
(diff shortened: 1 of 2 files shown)


kianjones9:
Closing because this was opened against `letta-code`, but the requested work was actually for the desktop app in `letta-cloud`. Leaving the desktop-scoped replacement PR in `letta-cloud` instead.

PR #2131, opened 2026-05-07
fix(cli): expand shell command previews

kianjones9:
## Summary
- expand inline bash approval previews so long commands and heredocs show substantially more content before truncating
- keep eager approval preview budgeting in sync so large shell previews still reserve enough space without flicker
- expand live shell tool-call previews and add a focused regression test that preserves long heredoc commands verbatim for rendering

## Test plan
- [x] `bun test src/tests/cli/shellSemanticDisplay.test.ts`
- [x] `bunx --bun @biomejs/biome@2.2.5 check src/cli/App.tsx src/cli/components/InlineBashApproval.tsx src/cli/components/ToolCallMessageRich.tsx src/cli/components/previews/BashPreview.tsx src/tests/cli/shellSemanticDisplay.test.ts`

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -703,7 +703,7 @@ const APPROVAL_PREVIEW_BUFFER = 4;
 const MIN_WRAP_WIDTH = 10;
 const TEXT_WRAP_GUTTER = 6;
 const DIFF_WRAP_GUTTER = 12;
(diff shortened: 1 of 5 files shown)


kianjones9:
Closing because this was opened against `letta-code`, but the requested work was actually for the desktop app in `letta-cloud`. Leaving the desktop-scoped replacement PR in `letta-cloud` instead.

PR #2152, opened 2026-05-08
feat(cli): add /palette command palette overlay

kianjones9:
## Summary

Adds a TUI command palette that lets users fuzzy-search and run any slash command from a single overlay. Useful for discovering commands without having to remember exact names or scrolling through `/help`.

Open it with `/palette`.

## Behavior

- Type to filter the list with fuzzy matching (exact > prefix > substring > subsequence > description match).
- `↑`/`↓` (or `Ctrl+P`/`Ctrl+N`) to move the selection.
- `Enter` runs the selected command via the normal slash-command pipeline (so any command that opens an overlay, prompts the agent, etc. behaves identically to typing it manually).
- `Esc` (or `Ctrl+C`) closes the palette.
- Lists both built-in commands and user-defined custom commands (\`.letta/commands/*\`), matching the set shown by `/help` and `/`-autocomplete.

## Implementation notes

- New \`src/cli/components/CommandPalette.tsx\` component, modeled after the existing dialog overlays (\`HelpDialog\`, \`SkillsDialog\`).
- Registered \`/palette\` in \`src/cli/commands/registry.ts\`.
- Wired into \`App.tsx\` as a new \`palette\` overlay with a small handler that dispatches the chosen command back through \`onSubmit\`, so all existing command behaviors are reused without duplication.

## Verification

- \`bunx tsc --noEmit\` — clean.
- \`bun x @biomejs/biome check\` on the changed files — clean.

### DIFF
--- a/src/cli/App.tsx
+++ b/src/cli/App.tsx
@@ -188,6 +188,7 @@ import { AssistantMessage } from "./components/AssistantMessageRich";
 import { BashCommandMessage } from "./components/BashCommandMessage";
 import { BtwPane, type BtwState } from "./components/BtwPane";
 import { CommandMessage } from "./components/CommandMessage";
+import { CommandPalette } from "./components/CommandPalette";
 import { CompactionSelector } from "./components/CompactionSelector";
 import { ConversationSelector } from "./components/ConversationSelector";
 import { colors } from "./components/colors";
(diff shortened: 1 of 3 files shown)


kianjones9:
Closing this out; we are discarding the command palette change for now.

PR #2159, opened 2026-05-09
feat(cli): always show all skill source tabs in /skills dialog

sarahwooders:
Show agent-scoped skills in `/skills`

image

## Summary
- The `/skills` dialog previously hid tabs with zero skills, so the **Agent** tab was invisible unless the agent already had agent-scoped skills
- Now all four tabs (**Project**, **Agent**, **Global**, **Bundled**) always render in that order, with a "No skills in this location" empty state for sources with no skills

### DIFF
--- a/src/agent/skills.ts
+++ b/src/agent/skills.ts
@@ -107,9 +107,9 @@ export const GLOBAL_SKILLS_DIR = join(
 );
 
 /**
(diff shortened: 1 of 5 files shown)


sarahwooders:
@letta-code please re-review

PR #2161, opened 2026-05-09
fix(local): route message search through backend

cpacker:
## Summary
- Route message search through an active-backend helper so `/search` uses local persisted conversation history instead of API-only `/v1/messages/search` in local mode.
- Move `letta messages` search/list/transcript and `letta agents` list/create output through the backend abstraction; explicitly reject `letta blocks` in local mode because it manages server memory blocks.
- Guard `/install-github-app` in local mode before opening its API-backed setup flow.

## Test plan
- [x] `bun test src/tests/backend/message-search.test.ts src/tests/cli/local-backend-command-wiring.test.ts src/tests/cli/message-search-cache-warm.test.ts src/tests/backend/local-backend.test.ts src/tests/cli/local-backend-command-wiring.test.ts`
- [x] `bun run check`


cpacker:
We should deprecate the blocks subcommand entirely

PR #2178, opened 2026-05-09
fix(local): retry transient provider stream failures

cpacker:
## Summary
- Add OpenCode-style same-turn retry handling for local AI SDK provider streams before assistant output is emitted.
- Retry transient provider failures including AI SDK retryable connection errors, practical provider 5xx responses, and non-quota rate limits while preserving context-overflow compaction handling.
- Emit local retry event chunks and add adapter/local-backend regression coverage for recovered single-turn persistence.

## Test plan
- [x] bun run check
- [x] bun test src/tests/backend/ai-sdk-stream-adapter.test.ts src/tests/backend/local-backend.test.ts src/tests/headless/dev-backend-smoke.test.ts

### DIFF
--- a/src/backend/dev/AISDKStreamAdapter.ts
+++ b/src/backend/dev/AISDKStreamAdapter.ts
@@ -1,4 +1,5 @@
 import {
+  APICallError,
   tool as aiTool,
   convertToModelMessages,
   jsonSchema,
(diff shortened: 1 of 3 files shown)


cpacker:
Addressed the review finding about non-output stream parts closing the retry window.

Changes pushed in `a5e01504`:
- `isModelOutputEvent` now only treats durable model output as closing the safe retry window (`text-delta`, `reasoning-delta`, `tool-call`, or final UI message), not bookkeeping parts like `start-step` / `finish-step`.
- Added a regression test where the first attempt emits `start-step` then throws a retryable 503; the adapter now retries and succeeds.

Validation rerun:
- `bun run check`
- `bun test src/tests/backend/ai-sdk-stream-adapter.test.ts src/tests/backend/local-backend.test.ts src/tests/headless/dev-backend-smoke.test.ts` → 64 pass, 0 fail

PR #2202, opened 2026-05-11
feat(ci): enable sticky comment for conversation persistence across review runs

carenthomas:
## Summary
- Enable `tracking_comment: "true"` + `use_sticky_comment: "true"` on the review action
- This stores `conversation_id` in the comment metadata so subsequent pushes to the same PR resume the existing conversation instead of starting fresh
- Sticky comment = one comment per PR, updated in place — not noisy like the old per-push comments

## Why
Every review run was creating a brand new conversation because `tracking_comment: "false"` meant no comment was created, so there was nowhere to store the `<!-- letta-metadata conversation_id: conv-xxx -->` block. On the next push, `findExistingAgent()` found nothing and started from scratch. The agent had no memory of previous reviews on the same PR.

## Test plan
- [ ] Merge and observe on the next PR that gets multiple pushes
- [ ] Verify only one comment appears per PR (updated in place, not duplicated)
- [ ] Verify the agent references previous review context on re-review

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -66,6 +66,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      # TODO: replace with SHA before merging to main
       - uses: letta-ai/letta-code-action@fix/agent-mode-tracking-comment-and-conversation-persistence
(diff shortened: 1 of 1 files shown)


carenthomas:
Abandoning this approach — switching to Letta API conversation lookup instead of GitHub comment metadata for persistence.

PR #2242, opened 2026-05-13
fix(local): set explicit AI SDK timeouts

just-cameron:
## Summary
- Set explicit AI SDK timeouts for local backend turns so provider calls do not inherit shorter transport defaults.
- Give local compaction a longer timeout budget and pass it through both generateText and streamText fallback paths.
- Add focused LocalBackend tests covering normal turn and compaction timeout wiring.

## Test plan
- [x] bun test src/tests/backend/local-backend.test.ts src/tests/backend/local-compaction-parity.test.ts
- [x] bun run lint
- [x] bun run typecheck


cpacker:
Closing as superseded by #2251, which implements the local timeout fix at the provider fetch layer in the OpenCode-style approach.

PR #2252, opened 2026-05-13
MCP stdio support

neversettle17-101:


devanshrj:
Thanks for contributing this. Client-side MCP support has now landed in #3549, including stdio, HTTP/SSE, per-agent configuration, OAuth, lifecycle cleanup, and tests. This implementation is therefore superseded, so we’re closing the PR.

PR #2255, opened 2026-05-13
fix(channels): preserve channel reply context across compaction

just-cameron:
## Problem

Channel-originated turns have a special reply contract: a normal assistant message is not delivered to Slack, Telegram, Discord, or other external channels. The user-visible response must go through `MessageChannel` with the correct channel and chat id.

The observed failure mode is around compaction. A summary can preserve the engineering/task context while dropping the operational routing context from the initiating channel message. After that, the agent may continue the work but answer locally instead of replying back to the Slack/Telegram/Discord user.

## Change

This PR moves the preservation rule to the local compaction prompts themselves.

Both local compaction modes now instruct the summarizer to preserve channel reply routing when the transcript includes a channel notification or external channel reminder. The preserved fields are:

- source/channel;
- `chat_id`;
- `sender_id`;
- `account_id`;
- `message_id` / `thread_id`;
- the fact that future user-visible replies must use `MessageChannel` rather than plain assistant text.

The channel notification XML and channel reminder text are otherwise unchanged. There is no new visible XML block and no extra persistent per-message content.

## Why this shape

This puts the rule at the compression boundary, which is where the information was being lost. It is more direct than adding more prose to the channel message itself, and it avoids pinning whole channel messages in the context window.

Pinning the whole initiating message would be stronger in one narrow sense, but it can keep stale/large channel context alive and consume context window over time. A future structural version could persist a small bounded channel-origin metadata record and rehydrate the reply contract independently of summaries. This PR is the smaller compaction-level mitigation.

## Validation

- `bun test src/tests/backend/local-compaction-parity.test.ts src/tests/channels/xml.test.ts`
- `bun test src/tests/channels src/tests/backend/local-compaction-parity.test.ts`

## Risk

Low. This changes only local compaction prompt text plus a unit test that locks the routing-preservation instruction into both prompt variants. It does not alter adapter routing, `MessageChannel`, registry state, channel notification XML, or outgoing channel behavior.

### DIFF
--- a/src/backend/local/compaction.ts
+++ b/src/backend/local/compaction.ts
@@ -50,6 +50,7 @@ This summary should be thorough in capturing technical details, code patterns, a
 3. **Important details**: Enumerate specific files and code sections examined, modified, or created, as well as important plan files, GitHub issues/PR links, and Linear ticket IDs. For each item, include why it matters and any relevant names, data, configs, or facts discussed.
    - **Preserve identifiers verbatim** (plan filename/path, exact URL, issue/PR number, ticket ID); do not paraphrase or truncate.
    - **Preserve referenced identifiers unless explicitly resolved**: Keep exact URLs/IDs from the conversation unless there is clear evidence they are no longer relevant.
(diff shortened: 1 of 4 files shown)


just-cameron:
Closing for now. This PR only covered a narrow prompt-level/local-backend mitigation and does not address the broader cloud/container agent continuity problem. The right fix needs a backend-agnostic design for preserving channel-origin reply routing across compaction/context transitions.

PR #2259, opened 2026-05-13
fix(local): support LM Studio connect base URL

### LINKED ISSUE #2258: Merged PR #2251 documents unsupported local LM Studio connect command
## Summary

The merged PR #2251 description shows a command that is not supported by the current CLI surface:

```bash
letta --backend local connect lmstudio --base-url ...
```

Ed T rebuilt letta-code from git with #2251 included and verified the timeout field in `providers.ts`, but the documented command fails because `--base-url` is not recognized.

## Source check

Against `dev/backend-execution-adapter` at `f19d9430`:

- Top-level `--backend` is not present in the parsed CLI options I checked.
- `letta connect` provider tokens are currently limited to:
  - `chatgpt`/`codex`
  - `anthropic`
  - `openai`
  - `zai`
  - `zai-coding`
  - `minimax`
  - `moonshot`
  - `kimi-code`
  - `gemini`
  - `openrouter`
  - `bedrock`
- `letta connect` parsed options are currently:
  - `--api-key`
  - `--method`
  - `--access-key`
  - `--secret-key`
  - `--region`
  - `--profile`
- There is no `lmstudio` provider token in `src/cli/subcommands/connect.ts` / `src/cli/commands/connect-normalize.ts`.
- There is no `--base-url` option for `letta connect`.
- Grep over source + built `letta.js` found no code reading `LETTA_LOCAL_BACKEND_TIMEOUT_MS` / `LOCAL_BACKEND_TIMEOUT`.
- The only timeout env I found is `LETTA_REQUEST_TIMEOUT_MS` in `src/backend/api/client.ts`, which is the API client path, not the local AI SDK `streamText(...)` provider path.

## User impact

Users following the merged PR instructions cannot configure LM Studio/local provider timeout via the documented command. This is especially relevant for slow local-model compaction/prefill; Ed's qwen3.6-27B setup needs roughly 600s, while faster local models complete closer to ~180s.

## Workaround

Ed locally worked around this by setting the timeout to `600000` instead of `false` in `providers.ts`. He can re-enable his ad-hoc patch if needed.

## Expected fix

Either:

1. update the PR/release/docs command to match the CLI surface that actually merged, or
2. land the missing CLI/provider plumbing for something like:

```bash
letta --backend local connect lmstudio --base-url ...
```

If the intended config path is an env var, the source should explicitly read it and pass it through to the local provider / AI SDK timeout path.

## Known / assumed / unknown

- Known: PR #2251 is merged: https://github.com/letta-ai/letta-code/pull/2251
- Known: the documented command fails for Ed because `--base-url` is not recognized.
- Known: current source check does not show `lmstudio` connect support or `--base-url` in `letta connect`.
- Assumed: the PR description got ahead of the code, or a dependent PR/branch was missed.
- Unknown: intended final UX: global env var, per-provider setting, or per-model timeout config.

just-cameron:
## Summary

Fixes https://github.com/letta-ai/letta-code/issues/2258.

This PR makes the documented local-provider command path real for LM Studio:

```bash
letta --backend local connect lmstudio --base-url http://127.0.0.1:1234/v1
```

It adds `--base-url` support to both the standalone `letta connect ...` subcommand and the in-app `/connect ...` command for local OpenAI-compatible providers where a custom endpoint is meaningful.

## Problem

PR #2251 documented a local LM Studio command using `--base-url`, but the merged CLI did not accept that option. Users following the PR instructions hit an unknown option / unsupported surface even though LM Studio itself was present in provider normalization. The local provider store also already had a `base_url` response field shape, but local provider creation did not persist it and AI SDK model construction did not read it back.

## Approach

The fix keeps the scope narrow:

- Allow `--base-url` only for local OpenAI-compatible providers that need endpoint configuration today: `lmstudio` and `ollama`.
- Validate and normalize the URL once in connect normalization helpers.
- Persist `base_url` into the local provider auth record.
- Prefer the stored provider `base_url` when creating AI SDK model factories, falling back to the existing env/default URL behavior.
- Keep API/cloud provider behavior unchanged; `baseUrl` is only used by the local provider store path.

## Before / after

Before:

- `letta --backend local connect lmstudio --base-url ...` rejected `--base-url`.
- A custom LM Studio endpoint could only be supplied through env/default paths, not the documented connect flow.

After:

- `letta --backend local connect lmstudio --base-url http://127.0.0.1:1234/v1` is accepted.
- `/connect lmstudio --base-url http://127.0.0.1:1234/v1` is also accepted in the app.
- The saved local provider URL is used when constructing the LM Studio AI SDK model factory.

## Risk / tradeoffs

This intentionally does not add arbitrary `--base-url` support to all providers. That avoids implying cloud BYOK provider APIs accept custom endpoint URLs when this PR only wires local provider storage and local AI SDK factories. If we want custom OpenAI-compatible endpoints for other providers later, that should be a separate, explicit provider/API design.

The callback type for test model creation now receives the resolved OpenAI-compatible factory options as a second argument. Existing callers that only accept the model string continue to work in TypeScript/JavaScript, and this lets tests assert the actual endpoint selected by the factory.

## Validation

- `bun install --frozen-lockfile`
- `bun test src/tests/backend/ai-sdk-model-factory.test.ts src/tests/cli/connect-subcommand.test.ts src/tests/cli/connect-normalize.test.ts`
- `bun run lint`
- `bun run typecheck`

### DIFF
--- a/src/backend/dev/AISDKModelFactory.ts
+++ b/src/backend/dev/AISDKModelFactory.ts
@@ -1,5 +1,8 @@
 import type { LanguageModel } from "ai";
-import { getLocalProviderApiKeyByName } from "../local/LocalProviderAuthStore";
+import {
+  getLocalProviderApiKeyByName,
+  getLocalProviderBaseUrlByName,
+} from "../local/LocalProviderAuthStore";
 import {
   type AISDKProvider,
   expectedAISDKProviderList,
@@ -26,7 +29,15 @@ export interface AISDKModelFactoryOptions {
   model?: string;
   createOpenAIResponsesModel?: (model: string) => LanguageModel;
   createAnthropicModel?: (model: string) => LanguageModel;
-  createOpenAICompatibleModel?: (model: string) => LanguageModel;
+  createOpenAICompatibleModel?: (
+    model: string,
+    options: {
+      apiKey?: string;
+      baseURL: string;
+      providerName: string;
+      headers?: Record<string, string>;
+    },
+  ) => LanguageModel;
   createChatGPTOAuthModel?: (model: string) => LanguageModel;
   createGoogleModel?: (model: string) => LanguageModel;
   createBedrockModel?: (model: string) => LanguageModel;
@@ -89,6 +100,18 @@ function localProviderApiKey(
   for (const providerName of providerNames) {
     const key = getLocalProviderApiKeyByName(providerName, storageDir);
     if (key) return key;
+  }
+  return envValue;
+}
+
+function localProviderBaseURL(
+  providerNames: readonly string[],
+  envValue: string | undefined,
+  storageDir?: string,
+): string | undefined {
+  for (const providerName of providerNames) {
+    const baseURL = getLocalProviderBaseUrlByName(providerName, storageDir);
+    if (baseURL) return baseURL;
   }
   return envValue;
 }
@@ -164,6 +187,11 @@ export function createAISDKModelFactory(
     spec.apiKeyEnv?.() ?? spec.fallbackApiKey,
     storageDir,
   );
+  const baseURL = localProviderBaseURL(
+    spec.localProviderNames,
+    spec.baseURL?.(),
+    storageDir,
+  );
 
   switch (spec.sdk) {
     case "openai-responses":
@@ -176,7 +204,7 @@ export function createAISDKModelFactory(
       return createAnthropicModelFactory({
         model,
         providerName: spec.providerName,
-        baseURL: spec.baseURL?.(),
+        baseURL,
         apiKey,
         createModel: options.createAnthropicModel,
       });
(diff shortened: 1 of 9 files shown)


just-cameron:
Closing this as superseded by the current main branch.

After opening this PR, I rebased against latest `main` and found that PR #2251 already landed the missing plumbing for the exact issue: `letta connect` now accepts `--base-url`, local provider storage persists `base_url`, and AI SDK model factory resolution reads the saved provider `base_url` before falling back to env/default URLs. It also includes the requested timeout support.

I verified latest `origin/main` with:

- `bun test src/tests/cli/connect-subcommand.test.ts src/tests/backend/ai-sdk-model-factory.test.ts src/tests/backend/local-provider-timeout.test.ts`

No extra changes from this branch are needed.

PR #2266, opened 2026-05-13
fix(channels): handle /help in channel chats

just-cameron:
## Summary

This PR adds a shared channel-level `/help` response for external channel chats.

Today, channel messages flow through the normal ingress path and into the agent unless they are pairing, routing, or approval-control replies. In Telegram, that means `/help` does not produce a useful Telegram bot help response. The user report came from Discord after #2253 was pulled, and Cameron noted that we likely do not support slash commands in channels.

The fix is intentionally small and shared: the channel registry now intercepts an exact `/help` message after pending approval replies are checked, but before normal routing/pairing delivery. It sends a direct reply through the channel adapter and does not enqueue an agent turn.

## Behavior

Before:

A Telegram `/help` message was treated like a normal inbound message. Depending on pairing/routing state, it could get pairing copy, route copy, or be delivered to the agent as user text.

After:

`/help` replies directly in the same external channel with concise channel usage guidance. It explains that the chat is connected to Letta Code, normal messages go to the connected agent, supported channel actions can be requested naturally, and unconnected chats should send a non-command message to get pairing instructions.

## Approach

The implementation lives in `src/channels/registry.ts` rather than in the Telegram adapter because this is a channel UX gap, not a Telegram-specific transport problem. That keeps behavior consistent across Telegram, Slack, Discord, and custom/community channels that use the same registry ingress path.

Pending control requests still take precedence. If an approval prompt is waiting, the response parser gets first chance to handle the message before `/help` is intercepted.

## Risk and tradeoffs

Risk is low. The change only intercepts messages whose trimmed lowercase text is exactly `/help`. Slash commands with arguments, normal messages containing `/help`, and pending approval replies continue through their existing paths.

This does not try to implement the full CLI slash-command surface in external channels. That would be much larger and higher risk. This PR only makes the common `/help` affordance useful instead of surprising.

## Validation

- [x] `bun test src/tests/channels/registry.test.ts src/tests/channels/registry-community-copy.test.ts`
- [x] `bun run lint`

### DIFF
--- a/src/channels/commands.ts
+++ b/src/channels/commands.ts
@@ -1,0 +1,117 @@
+import { getChannelDisplayName } from "./pluginRegistry";
+import type { ChannelAdapter, InboundChannelMessage } from "./types";
+
+export type ChannelSlashCommandKind = "direct" | "agent-scoped";
+
+export type ParsedChannelSlashCommand = {
+  name: string;
+  args: string;
+  raw: string;
+};
+
+export type ChannelSlashCommandDefinition = {
+  name: string;
+  aliases?: string[];
+  kind: ChannelSlashCommandKind;
+  summary: string;
+};
+
+const CHANNEL_SLASH_COMMANDS: ChannelSlashCommandDefinition[] = [
+  {
+    name: "help",
+    kind: "direct",
+    summary: "Show channel usage guidance.",
+  },
+];
+
+function channelDisplayName(channelId: string): string {
+  try {
+    return getChannelDisplayName(channelId);
+  } catch {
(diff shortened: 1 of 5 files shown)


just-cameron:
Linear tracking:

Project: https://linear.app/letta/project/channel-slash-commands-in-letta-code-9d4421dc07ee

Issue for this PR: https://linear.app/letta/issue/LET-8884/support-help-in-external-channel-chats

Follow-up router issue: https://linear.app/letta/issue/LET-8885/add-shared-router-for-channel-slash-commands

PR #2285, opened 2026-05-15
fix(channels): respect Slack permission defaults

just-cameron:
## Summary

This fixes Slack/channel permission defaults so channel-originated runs do not ask for tool approval when the account is configured for the current no-permission-prompt behavior.

The immediate problem was that Slack still had several hardcoded `standard` fallbacks even though the product default permission mode is now `unrestricted`. In addition, Slack only seeded a conversation permission mode when a route was first created and only for non-standard modes, so reused Slack routes could keep stale per-conversation permission state and leak generic tool approval prompts into Slack.

## What changed

- Added a Slack-specific default permission constant that currently matches the product default: `unrestricted`.
- Updated Slack account creation, setup, snapshots, config serialization, and loaded-account normalization to use that default instead of hardcoded `standard`.
- Changed legacy permission migration so the old `default` string maps to the current product default instead of permanently mapping to `standard`.
- Included Slack account default permission metadata on inbound deliveries for existing routes.
- Applied the Slack account default before queueing a channel turn, so stale persisted conversation state cannot override the channel account default.
- Always emits the Slack conversation-created permission seed event, including explicit `standard`, so intentional stricter account configs still work.

## Before / after

Before: a Slack account configured or expected to run without permission prompts could still ask for approval if the route already existed or if fallback config had silently become `standard`.

After: Slack conversations follow the account default consistently for new and reused routes. By default that means no generic tool approval prompt is sent to Slack. If an account is explicitly set to `standard`, that choice is still honored.

## Risk and tradeoffs

The main behavior change is that legacy Slack accounts without an explicit permission mode now inherit the current product default, `unrestricted`, instead of the older ask-flow behavior. That matches the current default permission model and the observed Slack expectation, but it is still a meaningful default change for old channel configs.

This does not remove `standard`; it only stops treating it as the implicit fallback/default.

## Validation

- `bun test src/tests/channels/registry.test.ts src/tests/channels/service.test.ts src/tests/channels/interactive.test.ts src/tests/permissions-mode.test.ts src/tests/websocket/listen-client-protocol.test.ts`
- `bunx --bun @biomejs/biome@2.2.5 check src/permissions/mode.ts src/channels/types.ts src/channels/accounts.ts src/channels/slack/accountConfig.ts src/channels/slack/setup.ts src/channels/service.ts src/channels/registry.ts src/websocket/listener/lifecycle.ts src/websocket/listener/permissionMode.ts src/tests/channels/registry.test.ts src/tests/channels/service.test.ts src/tests/permissions-mode.test.ts`
- `bun run typecheck`

### DIFF
--- a/src/channels/accounts.ts
+++ b/src/channels/accounts.ts
@@ -15,6 +15,7 @@ import type {
   TelegramChannelAccount,
 } from "./types";
 import {
+  DEFAULT_SLACK_PERMISSION_MODE,
   isDiscordChannelAccount,
   isFirstPartyChannelId,
   isSlackChannelAccount,
@@ -87,7 +88,8 @@ function normalizeLoadedAccount<T extends ChannelAccount>(account: T): T {
   }
   if (isSlackChannelAccount(next)) {
     const raw: string =
-      (next as SlackChannelAccount).defaultPermissionMode ?? "standard";
+      (next as SlackChannelAccount).defaultPermissionMode ??
+      DEFAULT_SLACK_PERMISSION_MODE;
     (next as SlackChannelAccount).defaultPermissionMode =
       (migratePermissionMode(raw) ?? raw) as SlackDefaultPermissionMode;
   }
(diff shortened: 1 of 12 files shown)


cpacker:
Closing this in favor of #2528, which ports the Slack permission-default fix against current main without carrying forward the stale branch as-is.

PR #2286, opened 2026-05-15
feat(channels): support /status in channel chats

just-cameron:
Linear: LET-8897 https://linear.app/letta/issue/LET-8897/support-status-in-external-channel-chats

## Summary

This PR adds the next low-risk direct channel slash command after the shared router work in #2267: /status.

It keeps the command inside the shared channel command router, returns a direct channel reply, and avoids sending /status to the agent as normal user text.

## Problem

External channels now have a safe slash-command ingress layer, but the only supported command was /help. Users still do not have a quick way to verify whether the current chat is connected to a channel account, whether the listener adapter is running, or which agent conversation the chat is routed to.

That status information is useful when diagnosing channel setup and routing issues, and it is safe to expose as a direct read-only response.

## Approach

The command registry now lists /status alongside /help as a direct channel-native command.

The registry gathers status context at the same ingress point where slash commands are handled: channel account configured/enabled state, adapter running state, and the current route if one exists for the chat and thread.

The response is intentionally plain text and line-oriented so it remains safe across Slack, Telegram, Discord, and community channel adapters.

## Before and after

Before:

/status was treated as an unsupported channel slash command and received the generic not-supported guidance.

After:

/status replies directly in the channel with account, listener, route, agent, and conversation status when available. It does not create a route, enqueue an agent turn, or change channel state.

## Limits and tradeoffs

This is deliberately not a bridge to the TUI command execution path. Stateful commands such as /compact, /model, /permissions, and /memory still need a separate agent/conversation-scoped execution contract before they should be exposed in external channels.

For unconnected chats, /status reports that no route is connected and tells the user to send a normal non-command message to receive pairing or connection instructions. It does not trigger pairing on its own.

## Risk

Low. The change is scoped to the shared channel slash-command router and registry ingress. Existing pending control request replies still run before slash-command handling, so approval and AskUserQuestion flows keep priority.

## Validation

- [x] bun test src/tests/channels/commands.test.ts src/tests/channels/registry.test.ts src/tests/channels/registry-community-copy.test.ts
- [x] bun run lint. Existing warnings remain in unrelated files: src/cli/helpers/shellSemanticDisplay.ts, src/tests/cli/reload-command.test.ts, and src/websocket/listener/control-inputs.ts.


overlord-letta:

PR #2290, opened 2026-05-15
fix(test): isolate plan-mode-disabled test from user settings

carenthomas:
## Summary
- The `mode changes reject plan mode when plan mode is disabled` test was reading from `~/.letta/settings.json` (the user's actual settings) because it never redirected `HOME` to a temp directory
- If the user has `planModeEnabled: true` locally, `isPlanModeEnabled()` returns true and the mode change isn't rejected — the test fails
- Fix: use a temp HOME directory like the other plan-mode tests in this file, with proper cleanup in a `finally` block

## Root cause
This is a pre-existing bug (not introduced by any recent PR). The test was always flaky — it only passed in CI because CI runners have clean HOME dirs with no `.letta/settings.json`.

## Test plan
- [x] `bun test src/tests/websocket/listen-client-protocol.test.ts -t "mode changes reject plan mode when plan mode is disabled"` — passes locally

🤖 Generated with [Letta Code](https://letta.com)

### DIFF
--- a/src/tests/websocket/listen-client-protocol.test.ts
+++ b/src/tests/websocket/listen-client-protocol.test.ts
@@ -4112,46 +4112,55 @@ describe("listen-client capability-gated approval flow", () => {
   });
 
   test("mode changes reject plan mode when plan mode is disabled", async () => {
+    const originalHome = process.env.HOME;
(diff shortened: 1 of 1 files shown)


carenthomas:
Duplicate of #2287 by @jnjpng — closing.

PR #2298, opened 2026-05-18
fix(channels): support local backend channel listeners

just-cameron:
## Summary
- Route `letta server --backend local --channels ...` through the local channel listener without requiring a dummy `LETTA_BASE_URL`.
- Add local backend channel setup docs and CLI help text.
- Document that inbound channel delivery and `MessageChannel` replies are separate, so missing replies can be model/prompting rather than adapter failures.


cpacker (approved):
Conflict resolution verified; targeted tests and CI are green.

PR #2299, opened 2026-05-18
feat(channels): support /pause and /resume in channel chats

just-cameron:
Linear: https://linear.app/letta/issue/LET-8909/support-pause-and-resume-in-external-channel-chats

## Summary

Add channel-native /pause and /resume commands for external channel chats.

## Problem

External channel messages currently either route into an agent turn or, for supported slash commands like /help, receive a direct channel reply. Operators do not have a lightweight way from the channel itself to temporarily stop a routed chat from sending normal messages into the agent, then turn it back on later.

## Approach

This adds /pause and /resume to the shared channel slash-command router. The commands look up the raw route for the current channel chat, including disabled routes, so /resume can recover a previously paused route. /pause flips the route enabled flag off and /resume flips it back on, then both commands send a direct reply through the adapter instead of delivering anything to the agent.

The implementation also reloads routes from disk on a miss, matching the existing channel routing behavior where standalone CLI changes may have updated channel routes since process startup.

## Before / after

Before, /pause or /resume was treated as an unsupported slash command and no route state changed.

After, /pause disables delivery for the current routed chat, /resume enables it again, and unsupported slash commands still get direct guidance listing the supported commands.

## Risk and tradeoffs

The scope is intentionally route-level: this pauses only the current chat/thread route, not the whole channel account or listener. If there is no existing route, the command explains that the chat needs to be connected first rather than creating a route implicitly.

This PR is based on current main, so it does not include the still-open /status PR. If /status merges first, the command list/test expectations may need a small merge update to include all three commands together.

## Validation

- bun test src/tests/channels/commands.test.ts src/tests/channels/registry.test.ts
- bun run lint

Lint passes with the existing unrelated warnings in src/cli/helpers/shellSemanticDisplay.ts and src/tests/cli/reload-command.test.ts.


overlord-letta:

PR #2300, opened 2026-05-18
fix(connect): add DeepSeek API support

overlord-letta:
## Summary

This PR wires native DeepSeek API support into the Letta Code provider connection path. A user reported from Discord that DeepSeek model entries and DEEPSEEK_API_KEY support existed in parts of the codebase, but letta connect could not actually resolve or save a DeepSeek provider.

## Problem

Before this change, native DeepSeek was not registered as a connectable BYOK provider. That meant letta connect deepseek was treated as unsupported even though DeepSeek models and environment-variable expectations were already present elsewhere. While tracing the connect path, I also confirmed the reported initialization bug: settings were only initialized inside the ChatGPT OAuth branch, so API-key and Bedrock provider paths could hit Settings not initialized before validation or save.

## Approach

The fix registers DeepSeek consistently across the same provider layers used by the other API-key providers: provider registry metadata, BYOK provider config, connect-provider normalization, BYOK aliasing for selectors/subagents, and the model catalog. DeepSeek is exposed as lc-deepseek with native model handles for deepseek-chat and deepseek-reasoner.

The settings initialization call now runs once after provider resolution and before dispatching into OAuth, API-key, or Bedrock-specific flows. This keeps unknown-provider help behavior unchanged while ensuring every real provider path has initialized settings before it reads or writes provider state.

## Before / after

Before: letta connect deepseek failed as an unknown provider, and non-OAuth connect flows could fail with Settings not initialized.

After: letta connect deepseek resolves to the DeepSeek API provider, validates/saves lc-deepseek, native DeepSeek model handles are available, and all supported provider dispatch paths initialize settings first.

## Risk and tradeoffs

Risk is low and scoped to provider registration and connect startup ordering. The change follows existing provider patterns and does not add a new validation strategy beyond the existing provider API path. I did not add a live DeepSeek API smoke test because that would require a real key in CI.

## Validation

- bun test src/tests/cli/connect-subcommand.test.ts src/tests/providers/byok-provider-aliases.test.ts
- bun run typecheck
- bunx --bun @biomejs/biome@2.2.5 check src/backend/dev/PiProviderRegistry.ts src/providers/byok-providers.ts src/cli/commands/connect-normalize.ts src/agent/subagents/manager.ts src/cli/subcommands/connect.ts src/tests/cli/connect-subcommand.test.ts src/tests/providers/byok-provider-aliases.test.ts
- bun run build

### DIFF
--- a/src/agent/subagents/manager.ts
+++ b/src/agent/subagents/manager.ts
@@ -159,6 +159,7 @@ function isProviderNotSupportedError(errorOutput: string): boolean {
 const BYOK_PROVIDER_TO_BASE: Record<string, string> = {
   "lc-anthropic": "anthropic",
   "lc-openai": "openai",
+  "lc-deepseek": "deepseek",
   "lc-zai": "zai",
   "lc-gemini": "google_ai",
   "lc-openrouter": "openrouter",
--- a/src/backend/dev/PiProviderRegistry.ts
+++ b/src/backend/dev/PiProviderRegistry.ts
@@ -9,6 +9,7 @@ import { getModels } from "@earendil-works/pi-ai";
 export const LOCAL_CHATGPT_PROVIDER_NAME = "chatgpt-plus-pro";
 export const LOCAL_OPENAI_PROVIDER_NAME = "lc-openai";
 export const LOCAL_ANTHROPIC_PROVIDER_NAME = "lc-anthropic";
+export const LOCAL_DEEPSEEK_PROVIDER_NAME = "lc-deepseek";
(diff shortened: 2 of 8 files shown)


cpacker:
Closing this as stale rather than carrying the old branch forward.

The local-provider side has been substantially refactored around the current pi-ai provider registry, so this patch no longer applies cleanly as-is. The remaining useful piece is likely a fresh, smaller PR for Constellation/API BYOK DeepSeek support plus the settings-initialization fix for API-key connect paths.

PR #2303, opened 2026-05-18
fix: better error message for quota hit

AriWebb:


src/tests/cli/errorFormatter.test.ts
```diff
@@ -280,6 +280,50 @@ describe("formatErrorDetails", () => {
     expect(message).toContain("/model");
   });
 
+  test("uses upgrade or purchase guidance for Letta-hosted quota plus credit exhaustion", () => {
+    setErrorContext({ billingTier: "team_pro", modelLabel: "auto" });
```
amelia-letta: modelLabel: "auto" won't match startsWith("letta/") — this test will fall through to the generic credit message and fail the assertions. Needs a letta-prefixed label like "letta/auto".

```suggestion
    setErrorContext({ billingTier: "team_pro", modelLabel: "letta/auto" });
```

4shub (approved):

PR #2304, opened 2026-05-18
feat(reflection): use message-id transcript state

devanshrj:
## Summary
- Move reflection transcript progress from `auto_cursor_line` to per-conversation backend message ids with lazy legacy-state migration.
- Track completed/reflected turn counters from transcript state so MemFS step-count reflection fires from unreflected turns instead of absolute reminder turn count.
- Serialize per-conversation transcript state writes with an in-process mutex plus file lock; corrupt/empty lock files are only reaped after a short grace period so in-flight lock writers do not wedge future acquisitions.
- Anchor cursors on backend message ids while starting payload ranges after the previous cursor line, which preserves live TUI user transcript rows that do not carry canonical backend ids.
- This PR intentionally excludes idle-time sweep/passive reflection settings.

## Validation
- [x] `bun test src/tests/cli/reflection-transcript.test.ts src/tests/cli/memoryReminder.test.ts src/tests/utils/fileLock.test.ts`
- [x] `bun test src/tests/cli/reflection-auto-launch-wiring.test.ts src/tests/websocket/listen-reflection-wiring.test.ts src/tests/telemetry/reflection-events-wiring.test.ts`
- [x] `bun test src/tests/reminders/listen-session-context.test.ts`
- [x] `bun run typecheck`
- [x] `bunx --bun @biomejs/biome@2.2.5 check --write src/cli/app/useSubmitHandler.ts src/cli/helpers/memoryReminder.ts src/cli/helpers/reflectionTranscript.ts src/reminders/engine.ts src/tests/cli/memoryReminder.test.ts src/tests/cli/reflection-transcript.test.ts src/websocket/listener/turn.ts src/tests/utils/fileLock.test.ts src/utils/fileLock.ts`
- [x] Cloud-backed live smoke: confirmed state files write `reflected_through_message_id`, step-count reflection launches from `turns_since_last_successful_reflection`, and payload ranges retain preceding user rows without canonical backend ids.

### DIFF
--- a/src/utils/fileLock.ts
+++ b/src/utils/fileLock.ts
@@ -1,0 +1,108 @@
+import { open, readFile, unlink } from "node:fs/promises";
+
+export type FileLockOptions = {
+  /** A lock file older than this is treated as abandoned and reaped. */
+  staleMs?: number;
+  /** Poll interval while waiting for a held lock. */
+  retryMs?: number;
+  /** Give up acquiring the lock after this many ms. */
+  timeoutMs?: number;
+};
+
+const DEFAULT_OPTIONS: Required<FileLockOptions> = {
+  staleMs: 30_000,
+  retryMs: 25,
+  timeoutMs: 10_000,
+};
+
+/**
+ * Cross-process critical section guarded by an O_EXCL lock file. The lock
+ * file path is created by the caller (its parent directory must already
+ * exist). Stale locks (older than `staleMs`) are reaped automatically so a
+ * crashed holder cannot block the system indefinitely.
+ */
+export async function withFileLock<T>(
+  lockPath: string,
+  fn: () => Promise<T>,
+  options?: FileLockOptions,
+): Promise<T> {
+  const opts = { ...DEFAULT_OPTIONS, ...options };
+  const release = await acquireFileLock(lockPath, opts);
+  try {
+    return await fn();
+  } finally {
+    await release();
+  }
+}
+
+async function acquireFileLock(
+  lockPath: string,
+  opts: Required<FileLockOptions>,
+): Promise<() => Promise<void>> {
+  const start = Date.now();
+  const payload = JSON.stringify({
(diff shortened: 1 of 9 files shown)


src/utils/fileLock.ts
```diff
@@ -0,0 +1,108 @@
...
+    // Lock vanished between EEXIST and read - retry immediately.
+    return true;
+  }
+  let acquiredAt: unknown;
+  try {
+    acquiredAt = (JSON.parse(raw) as { acquiredAt?: unknown }).acquiredAt;
+  } catch {
+    acquiredAt = undefined;
+  }
+  if (typeof acquiredAt !== "number" || Date.now() - acquiredAt <= staleMs) {
```
amelia-letta: A crash between `open` and `writeFile` (lines 50-51) leaves an empty lock file. `JSON.parse("")` throws → `acquiredAt = undefined` → `typeof undefined !== "number"` → returns `false` (not stale). The lock is permanently stuck until manual deletion — every subsequent acquisition times out after 10s.

Consider treating a non-parseable lock file as stale (it can only exist if the writer crashed mid-write):

```suggestion
  let acquiredAt: unknown;
  try {
    acquiredAt = (JSON.parse(raw) as { acquiredAt?: unknown }).acquiredAt;
  } catch {
    // Corrupt/empty lock file — writer crashed mid-write. Treat as stale.
    acquiredAt = -Infinity;
  }
  if (typeof acquiredAt !== "number" || Date.now() - acquiredAt <= staleMs) {
    return false;
  }
```

devanshrj: Good catch. Addressed in 686aaf38: corrupt/empty lock files are now reaped after a short grace period so crashed acquisitions do not wedge the lock, but live acquisitions are not reaped during the open→write window. I also added a regression test for empty lock files and made failed acquisitions clean up their lock file best-effort.

PR #2313, opened 2026-05-19
fix(channels): hide raw lifecycle stop errors

### LINKED ISSUE #2312: Channels surface raw generic loop errors instead of graceful failure details
## Summary

Telegram channel turns can send a user-visible failure like:

```text
Turn failed:
Unexpected stop reason: error
```

This happened on a channel-routed Telegram turn after the agent hit a backend run error. The normal Letta UI/headless path usually handles this class of failure more gracefully; channels appear to relay the lifecycle error string directly to the external user.

## Observed behavior

Telegram displayed a direct bot reply:

```text
Turn failed:
Unexpected stop reason: error
```

The user sees an opaque internal stop reason rather than a useful error message, retry behavior, or normal Letta error presentation.

## Log evidence

From `~/.letta/logs/remote/2026-05-18T03-33-16-232Z.log` around the failing turn:

```text
16:52:10 received Telegram channel message with Stainless URL
16:52:15 fetch_webpage called
16:52:17 fetch_webpage returned success, but with garbage/binary-looking text from the Framer page
16:52:26 shell fetch returned 451,258 bytes of HTML
16:52:38 extracted useful article text with BeautifulSoup
16:52:43 web_search called for Stainless/OpenAI customer confirmation
16:52:49 web_search returned
16:52:50 stop_reason: error
16:52:51 loop_error: Unexpected stop reason: error
```

The run usage near the error was large:

```text
prompt_tokens: 133,997
total_tokens: 134,126
context_tokens: 134,126
```

That may have triggered the underlying backend failure, but the channel-specific issue is the user-visible raw lifecycle error.

## Relevant code paths

Telegram channel lifecycle handler:

```ts
// src/channels/telegram/adapter.ts
if (event.outcome !== "error" || !event.error?.trim()) {
  return;
}
await sendLifecycleErrorReply(source, event.error ?? "Turn failed");
```

Telegram formatting:

```ts
return `Turn failed:\n${truncated}`;
```

Error string source:

```ts
// src/websocket/listener/turn.ts
const errorMessage =
  errorDetail || `Unexpected stop reason: ${stopReason}`;

emitLoopErrorNotice(...)
```

So when the server emits `stop_reason: error` without `errorDetail`, the channel lifecycle handler sends the generic fallback directly to Telegram.

## Expected behavior

Channel-routed turns should not expose raw internal stop reasons to end users. They should either:

- use the same graceful error/recovery/detail path as normal Letta runs, or
- retrieve/use run metadata when `errorDetail` is missing, or
- send a stable user-facing fallback while logging the internal stop reason elsewhere.

## Possible fixes

1. In channel lifecycle handling, avoid sending raw `Unexpected stop reason: error` directly to external channels.
2. Ensure `stop_reason: error` lifecycle events include useful detail, possibly by retrieving run metadata when `errorDetail` is absent.
3. Reuse the normal Letta UI/headless error handling path for channel-routed turns where possible.
4. Add a regression test for Telegram/Discord channel lifecycle errors where `stop_reason=error` and no `errorDetail` is present.

## Context

This was only observed via Telegram channels. Normal Letta interactions handle similar failures more gracefully, so this looks like a channel lifecycle/error-surfacing gap rather than merely a model/tool-output issue.

overlord-letta:
## Summary

Fixes https://github.com/letta-ai/letta-code/issues/2312.

Channel-routed turns could expose the internal lifecycle fallback `Unexpected stop reason: error` directly to users when the backend ended a run with `stop_reason: error` but did not provide a useful error detail. This was observed in Telegram, but Slack and Discord used the same class of lifecycle error reply path.

This PR adds a shared lifecycle error normalizer for channels. Useful backend details still pass through unchanged, while blank details and the raw generic stop reason become a stable user-facing fallback: “Something went wrong while processing that message. Please try again.” Telegram, Slack, and Discord lifecycle error formatting now all use that shared normalizer.

## Before / after

Before, Telegram could send:

Turn failed:
Unexpected stop reason: error

After, channels send:

Turn failed:
Something went wrong while processing that message. Please try again.

Specific actionable details, such as provider usage-limit messages, are preserved.

## Risk and tradeoffs

This is intentionally narrow: it does not change the underlying backend failure, retry behavior, or run metadata retrieval. It only prevents a known internal generic lifecycle string from leaking into external channels while keeping real details visible.

The fallback is shared across channel adapters so future lifecycle formatting stays consistent. Discord is covered by the shared normalizer even though the original report was Telegram.

## Validation

- bun test src/tests/channels/lifecycle-error.test.ts src/tests/channels/telegram-adapter.test.ts src/tests/channels/slack-adapter.test.ts
- bunx --bun @biomejs/biome@2.2.5 check src/channels/lifecycleError.ts src/channels/telegram/adapter.ts src/channels/slack/adapter.ts src/channels/discord/adapter.ts src/tests/channels/lifecycle-error.test.ts src/tests/channels/telegram-adapter.test.ts src/tests/channels/slack-adapter.test.ts

### DIFF
--- a/src/channels/discord/adapter.ts
+++ b/src/channels/discord/adapter.ts
@@ -1,4 +1,5 @@
 import { basename } from "node:path";
+import { normalizeChannelLifecycleErrorMessage } from "@/channels/lifecycleError";
 import type {
   ChannelAdapter,
   ChannelTurnLifecycleEvent,
@@ -272,7 +273,7 @@ export function buildDiscordReplyOptions(
 }
 
 function formatDiscordLifecycleErrorMessage(errorText: string): string {
-  const normalized = errorText.trim() || "Unknown error";
+  const normalized = normalizeChannelLifecycleErrorMessage(errorText);
   const truncated =
     normalized.length > DISCORD_LIFECYCLE_ERROR_TEXT_MAX
       ? `${normalized
--- a/src/channels/lifecycle-error.test.ts
+++ b/src/channels/lifecycle-error.test.ts
@@ -1,0 +1,28 @@
+import { describe, expect, test } from "bun:test";
+
+import {
+  CHANNEL_LIFECYCLE_FALLBACK_ERROR_MESSAGE,
+  normalizeChannelLifecycleErrorMessage,
+} from "./lifecycleError";
+
+describe("normalizeChannelLifecycleErrorMessage", () => {
+  test("keeps useful lifecycle details", () => {
+    expect(
+      normalizeChannelLifecycleErrorMessage(
+        "  ChatGPT usage limit reached. Resets at 1:00 PM.  ",
+      ),
+    ).toBe("ChatGPT usage limit reached. Resets at 1:00 PM.");
+  });
+
+  test("replaces raw generic loop errors with a stable user-facing fallback", () => {
+    expect(
+      normalizeChannelLifecycleErrorMessage("Unexpected stop reason: error"),
+    ).toBe(CHANNEL_LIFECYCLE_FALLBACK_ERROR_MESSAGE);
+  });
+
+  test("uses the fallback for blank lifecycle details", () => {
+    expect(normalizeChannelLifecycleErrorMessage("   ")).toBe(
+      CHANNEL_LIFECYCLE_FALLBACK_ERROR_MESSAGE,
+    );
+  });
+});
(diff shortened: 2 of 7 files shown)


overlord-letta:
Additional context from Cameron: this failure interrupted the agent run, so the agent itself may not be able to send a normal MessageChannel reply afterward. That is why this PR fixes the out-of-band channel lifecycle error path rather than relying on the agent turn to recover and message the user.

just-cameron (approved):

PR #2327, opened 2026-05-19
fix(channels): support /cancel slash command

overlord-letta:
## Summary
This adds `/cancel` as a supported Letta channel slash command so Slack, Telegram, Discord, and custom-channel users can interrupt an in-progress agent turn from the channel surface. It mirrors the TUI Esc behavior by routing the command through the listener's existing `abort_message` handling instead of inventing a separate cancellation path.

It also wires Slack-native `/cancel` command payloads into the same shared channel ingress path, and documents the required Slack app manifest/setup changes so the runtime change is not mistaken for automatic Slack app configuration.

## Problem
Channel users currently have no direct equivalent of pressing Esc in the TUI. If a channel-originated turn is running too long, stuck on a tool, or waiting on a control prompt, the user has to switch surfaces or wait for the run to finish.

Slack has two relevant command surfaces: normal channel messages that begin with `/cancel`, and Slack-native slash commands configured in the Slack app manifest. The shared channel parser can handle the first case by itself; the second case only works when the Slack app declares the command and the adapter receives a Bolt command payload.

## Approach
The shared channel slash-command router now advertises `/cancel` and accepts an agent-scoped handler. `ChannelRegistry` resolves the inbound chat/thread to its existing channel route, bypasses pending control-prompt replies for `/cancel`, and calls a listener-provided cancel handler. `wireChannelIngress` connects that handler to `handleAbortMessageInput`, which already sets cancellation state, aborts active controllers, rejects pending approvals, updates runtime status, and calls backend conversation cancellation.

The Slack adapter registers a native `/cancel` Bolt command handler and converts it into a normal `InboundChannelMessage`, so both native Slack commands and typed channel messages use the same cancellation logic. Because Slack-native slash command payloads do not identify the originating thread, the registry targets the sole routed thread for that Slack channel when there is exactly one unambiguous match; otherwise users can still send `/cancel` as a normal thread message.

The channel README now includes the Slack manifest additions for `/cancel`, including the `commands` bot scope, and the Slack setup prompt calls out the native command setup step.

## Before / after
Before: `/cancel` in a channel was treated as unsupported slash-command text or could be swallowed as a pending control-prompt reply. A Slack-native `/cancel` command was not registered by the adapter and would not reach Letta Code even if configured in Slack.

After: `/cancel` for a routed channel chat cancels the in-progress turn for that agent conversation. If there is no active turn, the channel gets an explicit no-op message. If the chat is not routed/connected yet, the channel gets connection guidance. A configured Slack-native `/cancel` command now reaches the same path.

## Risk and limits
Risk is mostly in channel command routing order and Slack thread ambiguity. The new tests cover routed cancellation, no-active-turn behavior, pending-control bypass, Slack-native command adapter routing, route-store isolation, and the sole-thread fallback for native Slack commands. The change intentionally does not expose other TUI slash commands in channels.

Slack-native command payloads still cannot identify one thread among many routed threads in the same channel. In that ambiguous case, users should send `/cancel` as a normal message in the target thread.

## Validation
- `bun test src/tests/channels/slack-adapter.test.ts src/tests/tools/tool-execution-context.test.ts src/tests/channels/commands.test.ts src/tests/channels/registry.test.ts`
- `bun run typecheck`
- `bunx --bun @biomejs/biome@2.2.5 check --write src/channels/README.md src/channels/registry.ts src/channels/slack/setup.ts src/tests/channels/registry.test.ts src/channels/slack/adapter.ts src/tests/channels/slack-adapter.test.ts src/tests/tools/tool-execution-context.test.ts`
- `git diff --check`

### DIFF
--- a/src/channels/slack/adapter.ts
+++ b/src/channels/slack/adapter.ts
@@ -78,6 +78,16 @@ type SlackReactionEvent = {
   event_ts?: string;
 };
 
+type SlackCommandPayload = {
+  command?: string;
+  text?: string;
+  user_id?: string;
+  user_name?: string;
+  channel_id?: string;
+  channel_name?: string;
+  trigger_id?: string;
+};
+
 type Constructor = abstract new (...args: never[]) => unknown;
 
 function isConstructorFunction<T extends Constructor>(
@@ -1111,6 +1121,51 @@ export function createSlackAdapter(
         });
       } catch (error) {
         console.error("[Slack] Error handling channel mention:", error);
+      }
+    });
+
+    instance.command("/cancel", async ({ command, ack }) => {
+      await ack();
+
+      if (!adapter.onMessage) {
+        return;
+      }
+
+      const payload = command as SlackCommandPayload;
+      if (
+        !isNonEmptyString(payload.channel_id) ||
+        !isNonEmptyString(payload.user_id)
+      ) {
+        return;
+      }
+
+      const commandText = isNonEmptyString(payload.text)
+        ? `/cancel ${payload.text.trim()}`
+        : "/cancel";
+
+      const inbound: InboundChannelMessage = {
+        channel: "slack",
+        accountId: config.accountId,
+        chatId: payload.channel_id,
+        senderId: payload.user_id,
+        senderName: firstNonEmptyString(payload.user_name, payload.user_id),
+        chatLabel: firstNonEmptyString(
+          payload.channel_name,
+          payload.channel_id,
+        ),
+        text: commandText,
+        timestamp: Date.now(),
+        messageId: firstNonEmptyString(payload.trigger_id, payload.command),
+        threadId: null,
+        chatType: resolveSlackChatType(payload.channel_id),
+        isMention: false,
+        raw: command,
+      };
+
+      try {
+        await adapter.onMessage(inbound);
+      } catch (error) {
+        console.error("[Slack] Error handling /cancel command:", error);
       }
     });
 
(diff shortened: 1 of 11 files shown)


src/channels/slack/adapter.ts
```diff
@@ -1114,6 +1124,51 @@ export function createSlackAdapter(
...
+        senderId: payload.user_id,
+        senderName: firstNonEmptyString(payload.user_name, payload.user_id),
+        chatLabel: firstNonEmptyString(
+          payload.channel_name,
+          payload.channel_id,
+        ),
+        text: commandText,
+        timestamp: Date.now(),
+        messageId: firstNonEmptyString(payload.trigger_id, payload.command),
+        threadId: null,
```
amelia-letta, line 1159: `threadId: null` means the native `/cancel` slash command can't reach thread-scoped conversations — the route lookup in `handleCancelSlashCommand` keys on threadId, so it'll miss the route and return "not connected" instead of cancelling. Most Slack agent conversations live in threads, so this is the common case.

overlord-letta:
Follow-up on Amelia review: fixed the native Slack /cancel thread-routing concern in 24583d72. Native Slack slash commands do not include thread metadata, so the registry now keeps exact thread cancellation for normal thread messages and adds a Slack-only fallback for native command payloads: if the channel/account has exactly one routed channel thread, /cancel targets that route; otherwise it stays unambiguous by requiring the normal thread message path. I also added coverage for that sole-thread fallback and documented the Slack manifest/setup requirements.

overlord-letta:

PR #2331, opened 2026-05-19
feat: add hosted flag for non-git endpoints and shell env endpoints

AriWebb:


src/backend/api/client.ts
```diff
@@ -141,6 +141,9 @@ export function getClientDefaultHeaders(): Record<string, string> {
       : nodeExperiment.enabled
         ? { "x-letta-node": "1" }
         : {}),
+    ...(process.env.LETTA_MEMFS_BACKEND === "pierre"
```
amelia-letta: `memoryGit.ts` and `shellEnv.ts` both renamed the env var value from `"pierre"` to `"hosted"`, but this still checks for `"pierre"`. Setting `LETTA_MEMFS_BACKEND=hosted` will route git correctly but miss the API client header; setting `=pierre` does the opposite.

```suggestion
    ...(process.env.LETTA_MEMFS_BACKEND === "hosted"
      ? { "x-letta-memfs-backend": "hosted" }
      : {}),
```

AriWebb:
@amelia-letta review this again

carenthomas (approved):

PR #2336, opened 2026-05-19
fix(listener): deduplicate sync device status and hoist supported com…

4shub:
…mands (LET-8948)

- Cache last emitted device status JSON per scope, skip WS send when unchanged
- Hoist SUPPORTED_REMOTE_COMMANDS to module-level constant to avoid allocation per sync

### DIFF
--- a/src/websocket/listener/protocol-outbound.ts
+++ b/src/websocket/listener/protocol-outbound.ts
@@ -64,6 +64,17 @@ type RuntimeCarrier = ListenerRuntime | ConversationRuntime | null;
 
 const GIT_CONTEXT_CACHE_TTL_MS = 15_000;
 const MAX_GIT_CONTEXT_CACHE_ENTRIES = 64;
+/**
+ * Frozen copy of the supported commands list. Avoids creating a new array on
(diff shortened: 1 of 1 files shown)


src/websocket/listener/protocol-outbound.ts
```diff
@@ -64,6 +64,17 @@ type RuntimeCarrier = ListenerRuntime | ConversationRuntime | null;
...
+ */
+/**
+ * Pre-computed copy of the supported commands list. Avoids creating a new
+ * array on every `buildDeviceStatus()` call (every 5–30 s per connected
+ * web client). (LET-8948)
+ */
+const FROZEN_SUPPORTED_COMMANDS: string[] = [...SUPPORTED_REMOTE_COMMANDS];
 const PROTOCOL_PERF_FLUSH_INTERVAL_MS = 1_000;
 const PROTOCOL_PERF_ENV_VALUES = new Set(["1", "true", "yes"]);
 const PROTOCOL_PERF_ENABLED = PROTOCOL_PERF_ENV_VALUES.has(
```
amelia-letta, line 80: duplicate JSDoc — the block above (lines 75-79) looks like it was meant to be deleted

PR #2337, opened 2026-05-19
feat(cli): add New tab to agent browser for creating agents

carenthomas:
## Summary
- Adds a "New" tab to the `/agents` browser tab bar (after Pinned, Letta Code, All)
- Makes agent creation discoverable — no longer hidden behind the N shortcut
- N key still works as a power-user shortcut that jumps to the New tab
- Removes the separate `NewAgentDialog` overlay (creation now happens inline)
- Removes `"new"` from the `ActiveOverlay` type union (dead code)

## Test plan
- [ ] Run `/agents` and verify "New" tab appears in the tab bar
- [ ] Tab over to "New" — should show name input with placeholder
- [ ] Type a name and press Enter — agent should be created
- [ ] Press Enter with empty input — should create with default name
- [ ] Press Esc with input — should clear input first
- [ ] Press Esc with empty input — should close the browser
- [ ] Press N from any tab — should jump to New tab
- [ ] Verify list tabs (Pinned, Letta Code, All) still work normally

### DIFF
--- a/src/cli/components/AgentSelector.tsx
+++ b/src/cli/components/AgentSelector.tsx
@@ -3,10 +3,13 @@ import { Box, useInput } from "ink";
 import { useCallback, useEffect, useRef, useState } from "react";
 import { getModelDisplayName } from "../../agent/model";
 import { type Backend, getBackend } from "../../backend";
+import { DEFAULT_AGENT_NAME } from "../../constants";
 import { settingsManager } from "../../settings-manager";
 import { useTerminalWidth } from "../hooks/useTerminalWidth";
 import { colors } from "./colors";
 import { MarkdownDisplay } from "./MarkdownDisplay";
+import { PasteAwareTextInput } from "./PasteAwareTextInput";
+import { validateAgentName } from "./PinDialog";
 import { Text } from "./Text";
 
 // Horizontal line character (matches approval dialogs)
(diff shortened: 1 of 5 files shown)


src/cli/components/AgentSelector.tsx
```diff
@@ -507,6 +517,37 @@ export function AgentSelector({
...
+        } else {
+          onCancel();
+        }
+      } else if (key.backspace || key.delete) {
+        setNewAgentNameInput((prev) => prev.slice(0, -1));
+        setNewAgentNameError("");
+      } else if (input && !key.ctrl && !key.meta && !key.tab) {
+        setNewAgentNameInput((prev) => prev + input);
+        setNewAgentNameError("");
+      }
```
amelia-letta: `useInput` handler for New tab and `PasteAwareTextInput` both handle Enter/backspace/regular input — key bleed. Enter fires twice: parent calls `onCreateNewAgent`, then `RawTextInput` calls `onSubmit` which also calls `onCreateNewAgent`. Would create two agents.

Only handle Escape here; let `PasteAwareTextInput` handle the rest via its `onChange`/`onSubmit`.

```suggestion
    // New tab has its own input handling (name input, not list navigation)
    if (activeTab === "new") {
      if (key.escape) {
        if (newAgentNameInput) {
          setNewAgentNameInput("");
          setNewAgentNameError("");
        } else {
          onCancel();
        }
      }
      // Don't handle return/backspace/regular input here — PasteAwareTextInput handles those
      return;
    }
```

carenthomas: good catch, fixed. `PasteAwareTextInput` now owns typing/backspace/Enter on the New tab, and the parent `useInput` handler only handles `Esc` there, so Enter no longer double-submits.

PR #2339, opened 2026-05-19
feat(local): add local-first setup flow

just-cameron:
## Summary
- Add a local-first first-run setup option for LET-8933, with `Proceed locally` selected by default for unauthenticated users.
- Keep the startup screen remote sign-in option labeled `Login to Constellation`, while leaving broader product/docs copy as Letta Cloud.
- Persist the selected startup backend preference and honor the saved local preference only when no cloud credentials or explicit `--backend` flag are present.
- Include the README local-mode docs in this implementation PR so setup flow and docs land together.

## Test plan
- [x] `bun test src/tests/cli/local-first-setup-wiring.test.ts src/tests/cli/local-backend-command-wiring.test.ts src/tests/backend/api-backend.test.ts`
- [x] `bun test src/tests/cli/local-first-setup-wiring.test.ts`
- [x] `bun run typecheck`
- [x] `git diff --check`
- [x] pre-commit `biome check --write` and `tsc --noEmit`


sarahwooders:
how can I switch back to constellation once I select local?

just-cameron:
You can switch back by launching with the API backend:

```bash
letta --backend api
```

That overrides the saved local preference for the invocation, opens the Constellation login flow if credentials are missing, and successful login persists the API backend preference for future launches.

I pushed 4457c216 to make this more discoverable in the README and local-mode completion message.

just-cameron:
We should probably make it easier to set your default preference too.

just-cameron:
Follow-up: agreed, one-off `--backend` was not enough for defaults.

I pushed 717b2cc9 adding explicit default controls:

```bash
letta backend          # show saved default
letta backend api      # default to Letta Cloud / Constellation path
letta backend local    # default to local mode
letta setup            # re-run the interactive setup menu
```

`--backend api|local` remains a one-off override. I also updated startup so a saved local default is honored even if Cloud credentials are present, which makes `letta backend local` actually switch the default back to local.

src/cli/subcommands/backend.ts
```diff
@@ -0,0 +1,97 @@
...
+Use --backend api or --backend local for a one-off override without changing
+the saved default.
+`.trim(),
+  );
+}
+
+function formatBackendName(mode: CliBackendMode | undefined): string {
+  return mode === "local" ? "local mode" : "Letta Cloud";
+}
+
```
amelia-letta, line 25: `letta backend` will report `local` for anyone with no Cloud creds, even if they never chose it. That makes the saved default UI lie about what is actually persisted in settings.

carenthomas (approved):

PR #2341, opened 2026-05-19
feat(cli): reimplement /experiments using MultiSelectPicker

carenthomas:
## Summary

- Replaces the bespoke `ExperimentSelector` with a `WindowTitlePicker`-style wrapper around the shared `MultiSelectPicker` component
- Checkbox UI with confirm/cancel flow — changes are batch-applied on Enter instead of toggling one-at-a-time immediately
- Env-forced experiments (`source: "env"`) are shown as disabled: dimmed, non-toggleable, labeled `set by environment · <description>`
- `MultiSelectPicker` gains a `disabled?: boolean` field on `SelectableItem` — Space is a no-op, item renders dimmed
- New `handleExperimentsConfirm` in `useConfigurationHandlers` diffs the confirmed selection against original state and calls `experimentManager.set` for each changed experiment

Net: removes ~40 lines of bespoke navigation/render logic from `ExperimentSelector`

## Test plan

- [ ] `/experiments` opens checkbox picker, Enter confirms, Esc cancels
- [ ] Toggling experiments on/off and confirming applies changes correctly
- [ ] No changes → confirm is a no-op (no API calls)
- [ ] `LETTA_NODE=1 bun run dev` → node experiment appears dimmed, Space does nothing on it
- [ ] Other experiments still toggle normally alongside a disabled item

### DIFF
--- a/src/cli/app/useConfigurationHandlers.ts
+++ b/src/cli/app/useConfigurationHandlers.ts
@@ -1143,11 +1143,38 @@ export function useConfigurationHandlers(ctx: ConfigurationHandlersContext) {
         return;
       }
 
+      const overlayCommand = consumeOverlayCommand("experiment");
+
+      if (isAgentBusy()) {
+        setActiveOverlay(null);
+        // For batch changes we can only queue one action; queue the first change.
+        const first = changes[0];
+        if (first) {
+          const cmd =
+            overlayCommand ??
+            commandRunner.start(
+              "/experiments",
+              "Experiment changes queued – will update after current task completes",
+            );
+          cmd.update({
+            output:
+              "Experiment changes queued – will update after current task completes",
+            phase: "running",
+          });
+          setQueuedOverlayAction({
+            type: "set_experiment",
+            experimentId: first.experimentId,
+            enabled: first.enabled,
+            commandId: cmd.id,
+          });
+        }
+        return;
+      }
+
       await withCommandLock(async () => {
-        const cmd = commandRunner.start(
-          "/experiments",
-          "Updating experiments...",
-        );
+        const cmd =
+          overlayCommand ??
+          commandRunner.start("/experiments", "Updating experiments...");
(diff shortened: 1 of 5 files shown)


src/cli/app/useConfigurationHandlers.ts
```diff
@@ -1134,6 +1134,41 @@ export function useConfigurationHandlers(ctx: ConfigurationHandlersContext) {
...
+      if (changes.length === 0) {
+        setActiveOverlay(null);
+        return;
+      }
+
+      await withCommandLock(async () => {
+        const cmd = commandRunner.start(
+          "/experiments",
+          "Updating experiments...",
+        );
```
amelia-letta: This starts a new command while the overlay command is still pending — when `setActiveOverlay(null)` fires in the `finally`, `closeOverlay` finishes the overlay command with dismiss output, so you get two commands worth of output (the new one + "Experiments dialog dismissed"). The old `handleExperimentSelect` uses `consumeOverlayCommand("experiment")` to take over the overlay command handle instead. Same pattern as the `handleCreateNewAgent` triple-render bug.

Also missing the `isAgentBusy()` / queued overlay action path that the old handler has.

```suggestion
      await withCommandLock(async () => {
        const overlayCommand = consumeOverlayCommand("experiment");

        if (isAgentBusy()) {
          setActiveOverlay(null);
          const cmd =
            overlayCommand ??
            commandRunner.start(
              "/experiments",
              "Experiment changes queued – will update after current task completes",
            );
          cmd.update({
            output:
              "Experiment changes queued – will update after current task completes",
            phase: "running",
          });
          setQueuedOverlayAction({
            type: "set_experiment",
            experimentId: changes[0].experimentId,
            enabled: changes[0].enabled,
            commandId: cmd.id,
          });
          return;
        }

        const cmd =
          overlayCommand ??
          commandRunner.start("/experiments", "Updating experiments...");
        cmd.update({ output: "Updating experiments...", phase: "running" });
```

PR #2344, opened 2026-05-19
Document local mode in README

just-cameron:
## Summary
- Document local mode, first-run choices, and local backend storage in the README.
- Clarify that local mode keeps state local but remote provider calls still leave the machine unless using Ollama, LM Studio, or llama.cpp.
- Paired with #2339: this is the README companion PR for the local-backend implementation changes.

## Test plan
- [x] `git diff --check README.md`
- [x] Docs-only change; no runtime tests run.


just-cameron:
Closing as superseded: the README local-mode docs have been rolled into #2339 so the implementation and docs land together.

PR #2363, opened 2026-05-20
feat: enforce layer boundary rules (backend, providers)

carenthomas:
## Summary
- Adds `scripts/check-layer-boundaries.js` enforcing two architectural boundary rules
- Wired into pre-commit hook and CI (`scripts/check.js`)

## Rules

| Layer | May not import from | Rationale |
|-------|-------------------|-----------|
| `backend/` | `cli/`, `websocket/` | Low-level abstraction used in headless and local modes — must be usable without the UI/WS stack |
| `providers/` | `agent/`, `cli/` | Pure LLM adapters — importing domain or UI code prevents provider swapping and isolation testing |

## Current state
Both rules are **violation-free today** — enforcing them costs nothing now and prevents future drift.

## How violations are reported
```
❌ Layer boundary violations found:

  src/backend/local/SomeFile.ts:12
    import { something } from "@/cli/helpers/foo"
    ↳ backend/ is a low-level abstraction — it must not import from cli/ or websocket/
```

## Adding new rules
Append to the `RULES` array in `scripts/check-layer-boundaries.js`. Audit for existing violations first.

## Test plan
- [ ] `bun run check:boundaries` → `✅ No layer boundary violations found.`
- [ ] `bun run check` → all checks pass
- [ ] Manually add a forbidden import → pre-commit hook blocks the commit

### DIFF
--- a/scripts/check-layer-boundaries.js
+++ b/scripts/check-layer-boundaries.js
@@ -1,0 +1,89 @@
+#!/usr/bin/env node
+/**
+ * Enforces architectural import boundaries between top-level modules.
+ *
+ * Rules:
+ *   backend/  must not import from  cli/  or  websocket/
+ *   providers/ must not import from  agent/  or  cli/
+ *
+ * These are currently violation-free. Adding a rule here means you must
+ * also ensure no existing code violates it.
+ *
+ * To add a new rule, append an entry to RULES below.
+ */
+
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { glob } from "glob";
+
+const rootDir = process.cwd();
+const srcDir = join(rootDir, "src");
+
+/**
+ * Each rule: files under `layer` must not contain an import from any of `forbidden`.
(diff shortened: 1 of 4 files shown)


scripts/check-layer-boundaries.js
```diff
@@ -0,0 +1,89 @@
...
+      "backend/ is a low-level abstraction — it must not import from cli/ or websocket/",
+  },
+  {
+    layer: "providers",
+    forbidden: ["agent", "cli"],
+    description:
+      "providers/ are pure LLM adapters — they must not import from agent/ or cli/",
+  },
+];
+
```
amelia-letta, line 40: Comment says "static and dynamic" but the regex only matches `from "@/..."` — dynamic `await import("@/cli/foo")` slips through. No violations exist today, but the comment is misleading. Either fix the regex or drop "and dynamic" from the comment.

```suggestion
// Matches: import ... from "@/forbidden/..." (static imports only)
```

PR #2376, opened 2026-05-20
fix: make spinner look nice

AriWebb:
https://github.com/user-attachments/assets/6281143e-eb52-47d4-b19e-ffe935d0287c


4shub:
pic?

AriWebb:
@4shub I actualy think the spinner regressed this is kinda how I think it used to look

jnjpng (approved):

PR #2405, opened 2026-05-21
feat(channels): consolidate channel slash commands

overlord-letta:
## Summary

Consolidates the channel slash-command work into one branch:

- Supersedes #2286 by adding /status for channel chats.
- Supersedes #2299 by adding /pause and /resume for channel route control.
- Supersedes #2327 by adding /cancel through the existing listener abort path, including Slack-native /cancel payload handling.
- Adds channel-tailored /chat and /reflection commands for routed channel conversations.

## Problem

The separate slash-command PRs all touched the same shared channel router, registry ingress path, and registry tests. Keeping them split caused repeat merge conflicts and made it too easy for the router API to drift in incompatible directions.

Channel users also need a practical command set without switching back to the TUI: inspect connection status, pause or resume route delivery, cancel an in-progress turn, jump to the Letta web chat, and manually trigger memory reflection.

## Approach

The shared channel command router now advertises /help, /status, /pause, /resume, /cancel, /chat, and /reflection. Parsing and user-facing copy stay in src/channels/commands.ts, while stateful operations stay in ChannelRegistry handlers.

/status reads account and route state. /pause and /resume update the persisted route enabled flag. /cancel calls the listener's existing abort-message path and still bypasses pending channel control prompts so it can act as an escape hatch. /chat returns the app.letta.com chat URL for the current routed agent/conversation when the route is cloud-backed. /reflection starts the same memory reflection subagent flow used by the listener, scoped to the routed agent conversation, when MemFS is enabled.

Slack-native /cancel command payloads are converted into normal inbound channel messages, with a sole-thread fallback for Slack's missing thread metadata. The other commands are typed as normal messages in the relevant channel chat/thread for now.

I dropped the stale Discord import-only changes from the pause/resume PR and moved old test additions onto the current colocated test paths.

## Before / after

Before: channel slash support was limited to /help on main, while /status, /pause, /resume, and /cancel lived on separate conflicting branches.

After: one shared implementation handles the consolidated channel command set across channel ingress: /status, /pause, /resume, /cancel, /chat, and /reflection.

## Limits and follow-up

This PR intentionally does not add /ade or /migrate. It also does not add /heartbeat, /reset, /setconv, /approve, or /disapprove.

Slack-native slash command payloads still cannot identify a specific thread. If a Slack channel has multiple routed Letta threads, users should send /cancel as a normal message in the target thread instead.

## Validation

- USER_CWD="$PWD" bun test src/channels/commands.test.ts src/channels/registry.test.ts src/channels/slack-adapter.test.ts src/tools/tool-execution-context.test.ts
- bun run typecheck
- bun run lint
- git diff --check

### DIFF
--- a/src/channels/registry.ts
+++ b/src/channels/registry.ts
@@ -784,7 +784,7 @@ export class ChannelRegistry {
     msg: InboundChannelMessage,
   ): Promise<boolean> {
     const slashCommand = parseChannelSlashCommand(msg.text);
-    if (slashCommand?.name === "cancel") {
+    if (slashCommand) {
       return false;
     }
 
--- a/src/channels/registry.ts
+++ b/src/channels/registry.ts
@@ -19,7 +19,17 @@ import {
   listChannelAccounts,
   loadChannelAccounts,
 } from "./accounts";
-import { tryHandleChannelSlashCommand } from "./commands";
+import {
+  buildChannelAlreadyActiveMessage,
+  buildChannelAlreadyPausedMessage,
+  buildChannelCancelNoActiveTurnMessage,
+  buildChannelCancelUnavailableMessage,
+  buildChannelNoRouteMessage,
+  buildChannelPausedMessage,
+  buildChannelResumedMessage,
+  parseChannelSlashCommand,
+  tryHandleChannelSlashCommand,
+} from "./commands";
 import { isDiscordGuildChannelAllowed } from "./discord/channel-gating";
 import {
   formatChannelControlRequestPrompt,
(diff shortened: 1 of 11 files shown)


src/channels/registry.ts
```diff
@@ -738,6 +783,11 @@ export class ChannelRegistry {
     adapter: ChannelAdapter,
     msg: InboundChannelMessage,
   ): Promise<boolean> {
+    const slashCommand = parseChannelSlashCommand(msg.text);
+    if (slashCommand?.name === "cancel") {
+      return false;
```
amelia-letta, line 788: `/reflection` still gets consumed by the pending approval prompt. In a channel thread waiting on approval, that means the command never reaches the new reflection handler.

PR #2406, opened 2026-05-21
fix(local): guard backend switches and replay payloads

just-cameron:
## Summary
- Strip OpenAI Responses replay item IDs from local Pi provider payloads so `store:false` transcript replay does not ask OpenAI for stale `rs_*`/`msg_*`/`fc_*` items.
- Keep local/API startup state isolated by backend family and fail explicitly when `--backend local` would otherwise reuse an incompatible ambient API agent.
- Add regression coverage for backend-family fallback filtering, explicit local backend switching, and OpenAI Responses replay payload sanitization.

## Test plan
- [x] `bun test src/backend/pi-stream-adapter.test.ts src/agent/resolve-startup-agent-files.test.ts src/backend-switch.test.ts`
- [x] `bun run typecheck`


cpacker (approved):
Conflict resolution verified; targeted tests and CI are green.

PR #2410, opened 2026-05-21
feat: dual api and local backend architecture

carenthomas:
Screenshot 2026-05-21 at 11 25 05 AM

### DIFF
--- a/src/cli/components/AgentSelector.tsx
+++ b/src/cli/components/AgentSelector.tsx
@@ -163,16 +163,14 @@ export function AgentSelector({
   const [localError, setLocalError] = useState<string | null>(null);
   const [hasLocalAgents, setHasLocalAgents] = useState(false);
 
-  // Tab state — default to Local if currently in local mode
-  const defaultTab: TabId = isLocalBackendEnabled() ? "local" : "pinned";
-  const [activeTab, setActiveTab] = useState<TabId>(defaultTab);
+  // Tab state — start at Pinned, jump to Local once agents are confirmed to exist
+  const [activeTab, setActiveTab] = useState<TabId>("pinned");
 
-  // Auto-switch away from Local tab if it disappears (no local agents found)
   useEffect(() => {
-    if (activeTab === "local" && !hasLocalAgents && !localLoading) {
-      setActiveTab("pinned");
+    if (isLocalBackendEnabled() && hasLocalAgents && activeTab === "pinned") {
+      setActiveTab("local");
     }
-  }, [activeTab, hasLocalAgents, localLoading]);
+  }, [hasLocalAgents, activeTab]);
 
   // Pinned tab state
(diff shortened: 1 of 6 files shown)


src/cli/components/AgentSelector.tsx
```diff
@@ -126,15 +134,40 @@ export function AgentSelector({
...
-  const selectorBackend = useCallback(() => {
-    const backend = backendRef.current ?? getBackend();
-    backendRef.current = backend;
-    return backend;
+
+  // Per-backend instances for listing — independent of the app-level active backend.
+  // Cloud tabs always use the API backend; Local tab always uses the local backend.
+  const cloudBackendRef = useRef<Backend | null>(null);
+  const getCloudBackend = useCallback((): Backend => {
+    cloudBackendRef.current ??= new APIBackend();
```
amelia-letta: This creates a second `APIBackend` instance separate from the `getApiBackendInstance()` singleton introduced in this same PR. Should use `getApiBackendInstance()` instead — two instances means two HTTP clients, two connection pools, etc.

```suggestion
    cloudBackendRef.current ??= getApiBackendInstance();
```

src/cli/components/AgentSelector.tsx
```diff
@@ -126,15 +134,40 @@ export function AgentSelector({
...
+  const [localHasMore, setLocalHasMore] = useState(true);
+  const [localLoading, setLocalLoading] = useState(true);
+  const [localSelectedIndex, setLocalSelectedIndex] = useState(0);
+  const [localPage, setLocalPage] = useState(0);
+  const [localError, setLocalError] = useState<string | null>(null);
+  const [hasLocalAgents, setHasLocalAgents] = useState(false);
+
+  // Tab state — default to Local if currently in local mode and local agents exist
+  const defaultTab: TabId = isLocalBackendEnabled() ? "local" : "pinned";
+  const [activeTab, setActiveTab] = useState<TabId>(defaultTab);
```
amelia-letta: `activeTab` defaults to `"local"` when `isLocalBackendEnabled()`, but `visibleTabs` hides the Local tab when `hasLocalAgents` is false (line 514). If local mode is on but no local agents exist, the user lands on a tab that isn't in the tab bar — orphaned state. Need a useEffect to auto-switch away when the tab disappears:

```suggestion
  const defaultTab: TabId = isLocalBackendEnabled() ? "local" : "pinned";
  const [activeTab, setActiveTab] = useState<TabId>(defaultTab);

  // Auto-switch away from Local tab if it disappears while active
  useEffect(() => {
    if (activeTab === "local" && !hasLocalAgents && !localLoading) {
      setActiveTab("pinned");
    }
  }, [activeTab, hasLocalAgents, localLoading]);
```

carenthomas:
redone in https://github.com/letta-ai/letta-code/pull/2419

PR #2413, opened 2026-05-21
feat: Remove plan mode

devanshrj:
## Summary
- Remove plan mode as a permission mode, including EnterPlanMode/ExitPlanMode tools, schemas, prompts, and settings.
- Delete plan-specific CLI/UI approval flows, plan file/viewer machinery, reminders, and channel/headless handling.
- Clean websocket protocol/runtime state and tests so only UpdatePlan remains as the planning helper.

## Test plan
- [x] `bun run typecheck`
- [x] `bun run lint`
- [x] `bun run check`
- [x] `bun test src/websocket/listen-client-protocol.test.ts src/websocket/listener-permission-mode.test.ts src/tools/interactivepolicy.test.ts src/channels/interactive.test.ts src/channels/registry.test.ts src/permissions/permissions-mode.test.ts`


src/permissions/format-denial.test.ts
```diff
@@ -96,26 +96,4 @@ describe("formatPermissionDenial", () => {
...
-  });
-
-  test("memory-mode denial shows fuller reason over short label", () => {
-    const result = formatPermissionDenial({
-      reason: "Permission mode: memory",
-      matchedRule: "memory mode",
-    });
-    expect(result).toBe("Permission denied: Permission mode: memory");
-  });
 });
```
amelia-letta, line 107: The "memory-mode denial shows fuller reason over short label" test was removed alongside the plan-mode test, but it tests a different behavior — that `formatPermissionDenial` prefers the `reason` field over `matchedRule` for memory mode. Unrelated to plan mode, should be kept.

src/permissions/permissions-mode.test.ts
```diff
@@ -1037,453 +835,3 @@ test("memory mode reason - Bash redirect outside roots names target", () => {
     else process.env.MEMORY_DIR = originalMemoryDir;
   }
 });
```
amelia-letta, line 837: The "memory mode reason - existing scoped denials get instructive reason too" regression test was removed alongside plan-mode tests. It verifies that memory mode denial reasons include the specific category (e.g. "command substitution") instead of the generic "Permission mode: memory". Unrelated to plan mode — consider keeping it.

src/permissions/format-denial.test.ts
devanshrj: Good catch — restored this memory-mode denial formatting test in 8ecb851f. The plan-mode test stayed removed, but this unrelated coverage is back.

src/permissions/permissions-mode.test.ts
devanshrj: Yep, agreed — restored the scoped memory-mode denial regression test in 8ecb851f so we keep coverage that specific denial reasons beat the generic memory-mode message.

carenthomas (approved):

PR #2419, opened 2026-05-21
feat(tui): add local and constellation agent tabs

carenthomas:
## Summary
- add separate Constellation and Local tabs to `/agents`, with Constellation shown before Local
- fetch Constellation agents directly from the cloud API, read Local agents from disk, and switch backends when selecting across modes
- improve Local agent UX by hiding the Local tab when empty, using file mtimes for recency, and pinning the current agent to the top

## Test plan
- [x] run `bun run check`
- [ ] open `/agents` while logged into Constellation and verify the Constellation tab shows cloud agents only
- [ ] verify Local tab appears only when local agents exist and lists the current local agent first
- [ ] switch between a Constellation agent and a Local agent from the picker and confirm backend switching works
- [ ] verify logged-out Constellation state shows the upsell copy

### DIFF
--- a/src/cli/app/use-conversation-switching.ts
+++ b/src/cli/app/use-conversation-switching.ts
@@ -25,7 +25,11 @@ import { getResumeDataFromBackend } from "@/agent/check-approval";
 import { createAgent } from "@/agent/create";
 import { selectDefaultAgentModel } from "@/agent/defaults";
 import { sendMessageStream } from "@/agent/message";
-import { getBackend } from "@/backend";
+import {
+  configureBackendMode,
+  getBackend,
+  isLocalBackendEnabled,
+} from "@/backend";
 import { getServerUrl } from "@/backend/api/client";
 import type { BtwState } from "@/cli/components/BtwPane";
 import {
@@ -476,6 +480,7 @@ export function useConversationSwitching(ctx: ConversationSwitchingContext) {
         profileName?: string;
         conversationId?: string;
         commandId?: string;
+        backendMode?: "local" | "api";
       },
     ) => {
       const overlayCommand = opts?.commandId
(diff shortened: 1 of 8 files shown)


src/cli/components/AgentSelector.tsx
```diff
@@ -554,17 +576,18 @@ export function AgentSelector({
       if (activeTab === "pinned") {
         const selected = pinnedPageAgents[pinnedSelectedIndex];
         if (selected?.agent) {
-          onSelect(selected.agentId);
+          const mode: AgentBackendMode = selected.isLocal ? "local" : "api";
```
amelia-letta, line 579: Pinned local agents always show as "Agent not found" when the picker opens in API mode — `loadPinnedAgents` uses `selectorBackend().retrieveAgent()` which only queries the current backend. Same issue hits delete: `selectorBackend().deleteAgent()` will 404 for agents from the other backend.

src/cli/app/use-conversation-switching.ts
```diff
@@ -528,6 +534,15 @@ export function useConversationSwitching(ctx: ConversationSwitchingContext) {
       cmd.update({ output: "Switching agent...", phase: "running" });
 
       try {
+        // Switch backend if the target agent belongs to a different backend
+        if (opts?.backendMode) {
+          const currentIsLocal = isLocalBackendEnabled();
+          const targetIsLocal = opts.backendMode === "local";
+          if (currentIsLocal !== targetIsLocal) {
+            configureBackendMode(opts.backendMode);
```
amelia-letta, line 542: If `retrieveAgent` fails after `configureBackendMode(opts.backendMode)`, we leave the whole session pointed at the new backend even though the switch never completed. A transient fetch failure here can strand the user on the wrong backend for the rest of the session.

PR #2422, opened 2026-05-21
feat(tui): add /login command

carenthomas:
## Summary
- add `/login` command that reuses the existing OAuth device code flow to sign in to Letta Constellation from within a running session
- no-ops with a hint about `/logout` if already signed in
- on success, saves credentials and switches to API backend

## Test plan
- [x] run `bun run check`
- [ ] run `/login` while logged out and verify the device code flow completes
- [ ] run `/login` while already signed in and verify it no-ops with `/logout` hint
- [ ] verify the Constellation upsell in `/agents` now has a working command to point to


src/auth/ConstellationLoginView.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+    const run = async () => {
+      try {
+        const currentSettings =
+          await settingsManager.getSettingsWithSecureTokens();
+        const hasApiKey =
+          process.env.LETTA_API_KEY || currentSettings.env?.LETTA_API_KEY;
+        const hasRefreshToken = Boolean(currentSettings.refreshToken);
+
+        if (hasApiKey || hasRefreshToken) {
```
amelia-letta: A saved refresh token alone is not enough to treat this as already logged in, `/login` can no-op here while the CLI is still in local mode with no usable API key.

PR #2439, opened 2026-05-22
fix(tui): simplify logout flow

carenthomas:
## Summary
- keep users in their current local session when they log out from local mode
- make cloud logout clear auth and exit instead of switching in-place to a local session with confusing transcript backfill
- block `/logout` while the agent is running
- refine already-logged-out copy to point users to `/login`
- remove the extra low-level logout success log line so the command result owns user-facing messaging

## Test plan
- [x] `bun run check` passes
- [x] Local agent + `/logout` keeps the session open and shows the local-session confirmation message
- [x] Cloud agent + `/logout` logs out and exits cleanly
- [x] `/logout` while already logged out shows: `Already logged out. Run /login to sign into Constellation.`
- [x] `/logout` while busy is blocked

### DIFF
--- a/src/settings-manager.ts
+++ b/src/settings-manager.ts
@@ -1384,6 +1384,52 @@ class SettingsManager {
   }
 
   /**
+   * Get the last session for the currently configured local backend storage dir,
+   * regardless of the active runtime backend mode.
+   *
+   * This is useful when switching from cloud -> local (e.g. /logout), where we
+   * need to resume the user's last local session even though the current server
+   * key still points at the cloud backend.
+   */
+  getLastLocalBackendSession(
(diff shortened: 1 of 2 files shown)


src/settings-manager.ts
```diff
@@ -1383,6 +1383,52 @@ class SettingsManager {
     );
   }
 
+  /**
```
amelia-letta: `getLastLocalBackendSession` is defined but never called anywhere — dead code. Should this be removed, or is it used in a follow-up?

carenthomas: Good catch — it was written for the original cloud→local session-resume-on-logout approach, but that approach was replaced with exit-on-logout. Removed in the latest commit.

PR #2446, opened 2026-05-22
feat(tui): show recent agents when pins are empty

carenthomas:
## Summary
- fall back to a small list of recent agents in startup selection when there are no usable pinned or LRU agents
- make the `/agents` pinned tab show recent agents when pins are empty or all pinned agents are invalid
- allow recent-agent fallback rows in the pinned tab to be pinned with `Shift+P`
- keep the copy and spacing focused on pinning as the primary concept while making the empty state useful

## Test plan
- [x] `bun run check` passes
- [x] Logged-in startup with no local agents and no usable pinned/LRU agents shows recent Constellation agents instead of create-new only
- [x] `/agents` pinned tab with no usable pins shows 3 recent agents instead of a dead empty state
- [x] `Enter` selects a recent fallback agent and `Shift+P` pins it
- [x] Spacing/copy for pinned fallback state matches the intended quickstart layout

### DIFF
--- a/src/cli/profile-selection.tsx
+++ b/src/cli/profile-selection.tsx
@@ -299,9 +299,7 @@ function ProfileSelectionUI({
 
   const hasLocalDir = settingsManager.hasLocalLettaDir();
   const contextMessage = externalFreshRepoMode
-    ? options.length > 0
-      ? "0 pinned agents available."
-      : "0 pinned agents available."
+    ? `${options.length} pinned agent${options.length !== 1 ? "s" : ""} available.`
     : hasLocalDir
       ? "Existing `.letta` folder detected."
       : `${options.length} agent profile${options.length !== 1 ? "s" : ""} detected.`;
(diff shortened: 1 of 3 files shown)


src/cli/profile-selection.tsx
```diff
@@ -275,7 +299,9 @@ function ProfileSelectionUI({
 
   const hasLocalDir = settingsManager.hasLocalLettaDir();
   const contextMessage = externalFreshRepoMode
-    ? `${options.length} pinned agent${options.length !== 1 ? "s" : ""} available.`
+    ? options.length > 0
+      ? "0 pinned agents available."
+      : "0 pinned agents available."
```
amelia-letta: Both branches return `"0 pinned agents available."` — the ternary is dead code. When there are valid pinned agents this will incorrectly say 0.

```suggestion
    ? `${options.length} pinned agent${options.length !== 1 ? "s" : ""} available.`
    : "0 pinned agents available."
```

PR #2460, opened 2026-05-22
fix(local): Support reasoning selection for local models

sarahwooders:
## Summary
- Adds generic reasoning effort options for discovered local model handles like `ollama/*`, so the reasoning selector appears for local models.
- Carries the selected reasoning effort through custom model updates into OpenAI-compatible settings.
- Cleans local model labels in the selector by hiding provider prefixes for local handles.

Validated with `bun test src/agent/model-tier-selection.test.ts src/cli/components/ModelSelector-categories.test.tsx` and `bun run check`.

### DIFF
--- a/src/cli/components/ModelSelector.tsx
+++ b/src/cli/components/ModelSelector.tsx
@@ -103,6 +103,36 @@ type UiModel = {
   updateArgs?: Record<string, unknown>;
 };
 
+const LOCAL_MODEL_HANDLE_PREFIXES = [
+  "ollama/",
+  "ollama-cloud/",
+  "lmstudio/",
+  "llama.cpp/",
+  "llama-cpp/",
+];
+
+function isLocalModelHandle(handle: string): boolean {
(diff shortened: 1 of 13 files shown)


src/cli/components/ModelSelector.tsx
```diff
@@ -103,6 +103,36 @@ type UiModel = {
   updateArgs?: Record<string, unknown>;
 };
 
+const LOCAL_MODEL_HANDLE_PREFIXES = [
```
amelia-letta: `LOCAL_MODEL_HANDLE_PREFIXES` is duplicated here and in `src/agent/model.ts`. If someone adds a new local provider prefix to one but not the other, they diverge silently — one determines which models get reasoning options, the other determines which labels get cleaned up. Consider importing `isLocalModelHandle` from `@/agent/model` and either exporting the prefixes or a `localModelLabel` helper from there too.

cpacker (approved):
Addressed the duplicated local-model prefix logic by moving selector label stripping into the shared model utility module. Targeted tests and `bun run check` pass locally.

PR #2465, opened 2026-05-22
fix(startup): keep no-model label on local resume

carenthomas:
## Summary
- keep showing `No model selected` for local agents while the local backend still reports zero available models
- stop tying the startup model override to `agentProvenance.isNew`, so the label persists across immediate restarts of the same empty local agent
- add a focused helper test covering local/no-model, local/has-models, and non-local cases

## Test plan
- [x] `bun run check` passes
- [x] `bun test src/cli/helpers/startup-model-display.test.ts` passes
- [ ] Verify a fresh local no-model agent still shows `No model selected`
- [ ] Verify restarting the same local no-model agent still shows `No model selected`
- [ ] Verify once local models become available later, the UI returns to the real model label

### DIFF
--- a/src/cli/components/WelcomeScreen.tsx
+++ b/src/cli/components/WelcomeScreen.tsx
@@ -73,7 +73,7 @@ export function WelcomeScreen({
   loadingState,
   continueSession,
   agentState,
-  agentProvenance,
+  agentProvenance: _agentProvenance,
   startupHasAvailableLocalModels = true,
 }: {
   loadingState: LoadingState;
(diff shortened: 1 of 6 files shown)


src/cli/components/WelcomeScreen.tsx
```diff
@@ -73,7 +73,7 @@ export function WelcomeScreen({
   loadingState,
   continueSession,
   agentState,
-  agentProvenance,
+  agentProvenance: _agentProvenance,
```
amelia-letta: `agentProvenance` is now dead — destructured as `_agentProvenance` and never used. The `AgentProvenance` import (line 5) also only feeds the prop type. Both should be removed per clean-code mantra.

PR #2467, opened 2026-05-22
fix: clean up channel lifecycle errors

overlord-letta:
## Summary
- Replaces raw conversation-busy/conflict errors with shared user-facing copy and run IDs, while keeping automatic-retry wording opt-in only.
- Sanitizes channel lifecycle errors so OSC8 terminal hyperlinks and app chat URLs are not emitted into plain/channel surfaces.
- Adds Telegram inline "Report error" handling that submits minimal sanitized lifecycle metadata through the feedback endpoint; Slack/Discord now share the same formatter but do not yet expose report buttons.

## Before / after
Before, channel lifecycle failures could leak terminal hyperlink escapes and app URLs into user-visible copy:

```text
Turn failed:
Cannot send a new message: Another request (run_id=run-example) is currently being processed for this conversation. Please wait for it to complete.
View agent: \x1b]8;;https://app.letta.com/chat/agent-example\x1b\\agent-example\x1b]8;;\x1b\\ (run: run-example)
```

After, Slack and Discord get the shared sanitized message with only the useful action and generic run ID:

```text
Turn still running
Another request is already processing for this conversation. Please wait for it to finish, then try again.

Run ID: run-example
```

Telegram uses the same sanitized lifecycle copy and also renders a native inline "Report error" button. That button submits sanitized lifecycle metadata, not full channel transcripts.

## Validation
- `bun test src/channels/lifecycle-error.test.ts src/channels/telegram-adapter.test.ts src/cli/error-formatter.test.ts`
- `bun run lint`
- `bun run typecheck`
- `bun run check`

## Known limitations
- Report button parity is Telegram-only in this PR; Slack and Discord use the shared lifecycle formatter and can add native actions on top of `lifecycle-error-report.ts` later.


src/cli/helpers/error-formatter.ts
```diff
@@ -665,8 +688,10 @@ export function formatErrorDetails(
   e: unknown,
   agentId?: string,
   conversationId?: string,
+  options: FormatErrorDetailsOptions = {},
 ): string {
   let runId: string | undefined;
+  const surface = options.surface ?? "plain";
```
amelia-letta: Defaulting `surface` to `"plain"` drops OSC8 hyperlinks from all ~40 CLI terminal callers (use-conversation-loop, use-submit-handler, etc.) that don't pass options. The channels already use `formatChannelLifecycleErrorMessage` instead of `formatErrorDetails`, so they won't be affected by changing the default. Suggest defaulting to `"terminal"` since every current caller IS a terminal.

```suggestion
  const surface = options.surface ?? "terminal";
```

overlord-letta: Fixed in e0de700a. `formatErrorDetails` now defaults to `terminal`, and tests cover default OSC8 links plus explicit `plain` sanitization.

cpacker (approved):
Resolved the main merge conflict and verified the lifecycle-error cleanup against the current Telegram forum-topic/debounce path. Targeted lifecycle/Telegram/error-formatter tests and `bun run check` pass locally.

PR #2482, opened 2026-05-22
feat(statusline): add renderer foundation

carenthomas:
## Summary
- add a statusline renderer contract, shared display components, and segment helpers for the current statusline pieces
- add built-in `legacy` and `custom` renderers, with `legacy` remaining the default so user-visible behavior does not change
- route the existing InputRich footer through the legacy renderer instead of hardcoding the default statusline label inline
- add focused tests for renderer registry/defaults, legacy output, and the empty custom placeholder

## Test plan
- [x] `bun test src/cli/statusline-renderer.test.tsx` passes
- [x] `bun run check` passes
- [ ] Verify the default footer still matches the current legacy output in the TUI
- [ ] Verify the no-model placeholder still suppresses reasoning in the footer


src/cli/components/InputRich.tsx
```diff
@@ -580,43 +572,20 @@ const InputFooter = memo(function InputFooter({
...
+  const shouldRenderDefaultStatusline =
+    !hideFooterContent &&
+    !ctrlCPressed &&
+    !escapePressed &&
+    !isBashMode &&
+    !statusLineText &&
+    !(modeName && modeColor) &&
+    !footerNotification &&
+    !hasQueuedMessages &&
+    backgroundAgents.length === 0;
```
amelia-letta: `shouldRenderDefaultStatusline` is missing `!statusLineRight` — if `statusLineRight` is set without `statusLineText`, the early return fires and skips the `statusLineRight` rendering in the right column (line 655).

```suggestion
  const shouldRenderDefaultStatusline =
    !hideFooterContent &&
    !ctrlCPressed &&
    !escapePressed &&
    !isBashMode &&
    !statusLineText &&
    !statusLineRight &&
    !(modeName && modeColor) &&
    !footerNotification &&
    !hasQueuedMessages &&
    backgroundAgents.length === 0;
```

PR #2486, opened 2026-05-23
fix(local): drop empty provider text blocks

overlord-letta:
## Summary
- Problem: malformed local active context could carry empty text content blocks into strict Anthropic-backed provider requests.
- Fix: normalize local provider messages at the PiStreamAdapter boundary by dropping empty text blocks, skipping empty-only user/assistant messages, and preserving empty tool results with a non-empty placeholder.
- Risk/limits: this filters request context on use rather than rewriting existing transcript files.

## Test plan
- [x] bun test src/tests/backend/pi-stream-adapter.test.ts
- [x] bun run check


cpacker (approved):
Reviewed the Letta Code adapter changes against the local pi-ai provider converters. The normalization is at the right boundary, preserves tool-result validity, and keeps trailing assistant trimming after filtering.

PR #2487, opened 2026-05-23
feat(telegram): support group topics and mention routing

just-cameron:
## Summary
- Add Telegram group/supergroup/topic routing with per-topic route creation and replies back into the same topic.
- Add Telegram mention-only group mode, display-name mention detection, inbound debounce, and reply-context rendering for quoted Telegram messages.
- Allow forked subagents launched from channel turns to inherit scoped MessageChannel access back to the originating channel without broadening their toolset.

## Test plan
- [x] bun test src/channels/telegram-adapter.test.ts src/channels/telegram-registry.test.ts src/telegram-debounce.test.ts src/channels/xml.test.ts src/agent/subagent-env-composition.test.ts src/agent/subagent-model-resolution.test.ts
- [x] bun run typecheck


src/tools/toolset.ts
```diff
@@ -373,6 +376,26 @@ export async function prepareToolExecutionContextForScope(params: {
     }
   })();
 
+  const inheritedContextId = process.env.LETTA_INHERITED_TOOL_CONTEXT_ID;
+  const inheritedContext = inheritedContextId
+    ? getExecutionContextById(inheritedContextId)
```
amelia-letta: `getExecutionContextById` looks up from a `globalThis`-scoped Map — the child process spawned by `executeSubagent` won't have the parent's entries, so this always returns `undefined`. The `LETTA_INHERITED_TOOL_CONTEXT_ID` env var is set correctly but the lookup is cross-process and silently fails, making the inherited channel scope feature a no-op. Falls through to `resolveConversationChannelToolScope` which is safe, but the scoped inheritance described in the PR summary doesn't actually work.

cpacker (approved):
Conflict resolution verified, MessageChannel regression fixed, and CI is green.

cpacker (approved):
Re-verified after merging latest main; targeted tests and CI are green.

PR #2500, opened 2026-05-23
feat(statusline): add local extension runtime

carenthomas:
## Summary
- Add a global local extension loader/runtime for ~/.letta/extensions files, with status values and full-row custom statusline renderers.
- Move extension runtime ownership to AppCoordinator and thread it through to the input statusline slot.
- Preserve host-owned safety/transient hint preemption while letting custom renderers own idle space, including reload loading behavior to avoid default flashes.

## Test plan
- [x] bun test src/cli/statusline-renderer.test.tsx src/cli/extensions/local-extension-loader.test.ts src/cli/chdir-command.test.ts src/cli/statusline-config.test.ts src/cli/statusline-controller.test.ts src/cli/statusline-help.test.ts src/cli/statusline-runtime.test.ts
- [x] bun run check

### DIFF
--- a/src/cli/extensions/local-extension-loader.ts
+++ b/src/cli/extensions/local-extension-loader.ts
@@ -1,0 +1,363 @@
+import { createHash } from "node:crypto";
+import {
+  existsSync,
+  mkdirSync,
+  readdirSync,
+  readFileSync,
+  statSync,
+  symlinkSync,
+  unlinkSync,
+  writeFileSync,
+} from "node:fs";
+import { createRequire } from "node:module";
+import { homedir } from "node:os";
+import path from "node:path";
+import { pathToFileURL } from "node:url";
+import type {
+  StatuslineRenderContext,
+  StatuslineRenderer,
+  StatuslineRendererOutput,
+} from "@/cli/display/statusline/types";
+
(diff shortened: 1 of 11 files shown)


src/cli/extensions/local-extension-loader.ts
```diff
@@ -0,0 +1,363 @@
...
+  ".letta",
+  "extensions",
+);
+export const EXTENSION_CACHE_DIRECTORY = path.join(
+  homedir(),
+  ".letta",
+  "extension-cache",
+);
+
+const EXTENSION_FILE_EXTENSIONS = new Set([".js", ".mjs", ".ts", ".tsx"]);
```
carenthomas, line 34: This advertises `.ts`/`.tsx` extensions, but the cached file is imported directly by `node letta.js`; Node 18/20 will throw `ERR_UNKNOWN_FILE_EXTENSION` and silently skip them. We need transpile-to-JS in the cache or only accept JS here.

carenthomas: fixed, TS/TSX sources now transpile into cached .mjs files before dynamic import, and the cache always imports ESM under Node. also moved react to runtime deps because TSX output imports react/jsx-runtime.

PR #2505, opened 2026-05-23
fix(mcp): repair legacy server refresh

ZanzyTHEbar:
## Summary

- Repair legacy MCP server configs from `/mcp` manual refresh when the first refresh still returns zero tools
- Retry once with a minimal nested HTTP/SSE config patch before reloading tools
- Skip stdio, unusable URLs, and servers with visible auth/custom headers to avoid unsafe rewrites

## Context

This is a defensive Letta Code workaround for legacy/flattened MCP server rows while the server-side normalization issue is tracked in letta-ai/letta#3353.

## Validation

- `bun test src/cli/helpers/mcp-server-repair.test.ts` — PASS
- `bun run check` — PASS

## Risks

- The authoritative fix should still live server-side so all clients benefit.
- The CLI repair sends only `mcp_server_type` and `server_url` to avoid clearing auth/header fields, and only runs after a zero-tool manual refresh.

### DIFF
--- a/src/cli/components/McpSelector.tsx
+++ b/src/cli/components/McpSelector.tsx
@@ -7,6 +7,7 @@ import type { Tool } from "@letta-ai/letta-client/resources/tools";
 import { Box, useInput } from "ink";
 import { memo, useCallback, useEffect, useState } from "react";
 import { getClient } from "@/backend/api/client";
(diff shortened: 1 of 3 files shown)


ZanzyTHEbar:
determined this belongs in `letta-ai/letta` server refresh/config normalization

PR #2517, opened 2026-05-24
fix(nix): nix flake fixes and improvements

michaelBelsanti:
- Copy node_modules to fix build
- Wrap with bun, matches non-nix default
- Strip TSC type gen
- Commit flake.lock for reproducibility

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -33,9 +33,9 @@
           };
 
           nativeBuildInputs = [
+            pkgs.nodejs_22
             pkgs.bun
             pkgs.makeWrapper
-            pkgs.nodejs_22
             pkgs.pkg-config
             pkgs.python3
             pkgs.bun2nix.hook
@@ -54,6 +54,8 @@
 
           preBuild = ''
             export HOME="$TMPDIR"
+            substituteInPlace build.js \
+              --replace-fail 'await Bun.$`bunx tsc -p tsconfig.types.json`' 'true'
           '';
 
           buildPhase = ''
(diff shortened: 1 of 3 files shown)


flake.nix
```diff
@@ -54,6 +54,8 @@
 
           preBuild = ''
             export HOME="$TMPDIR"
+            substituteInPlace build.js \
+              --replace-fail 'await Bun.$`bunx tsc -p tsconfig.types.json`' 'true'
```
kianjones9, line 58: Doesn't this mean the types won't exist in the built package? it will no longer produce these right?

```
dist/types/protocol.d.ts
dist/types/protocol.d.ts.map
```

flake.nix
```diff
@@ -73,9 +75,12 @@
...
               --strip-components=1
 
-            makeWrapper ${pkgs.nodejs_22}/bin/node "$out/bin/letta" \
+            cp -rL node_modules "$out/lib/letta-code/node_modules"
+
+            makeWrapper ${pkgs.bun}/bin/bun "$out/bin/letta" \
               --add-flags "$out/lib/letta-code/letta.js" \
-              --prefix PATH : ${lib.makeBinPath [ pkgs.git pkgs.ripgrep ]}
+              --prefix PATH : ${lib.makeBinPath [ pkgs.git pkgs.ripgrep ]} \
+              --prefix LD_LIBRARY_PATH : ${pkgs.stdenv.cc.cc.lib}/lib # libstdc++.so.6 for sharp
```
kianjones9: this should be linux only right? wouldn't we want something like
```
${lib.optionalString pkgs.stdenv.hostPlatform.isLinux ''
  wrapProgram "$out/bin/letta" \
    --prefix LD_LIBRARY_PATH : ${pkgs.stdenv.cc.cc.lib}/lib
''}
```

flake.nix
michaelBelsanti: yeah. are they needed? just had it to speed up and shrink the build a little.

flake.nix
michaelBelsanti: mb, didnt think about the macos build. will add it.

michaelBelsanti:
i also updated bun.nix. without a way to keep it in sync with bun.lock, the package will fail to build.

kianjones9:
Looks good thanks for your contribution!

PR #2520, opened 2026-05-24
fix(cli): simplify agent info footer

cpacker:
## Summary
- Replace the slash-command footer alien block with a compact text layout that keeps the agent name accented.
- Hide ADE and usage shortcuts for local-backend agents while keeping them available for Constellation/API agents.
- Reduce slash-command autocomplete to five visible rows and wrap long agent/conversation identifiers instead of truncating them.

## Test plan
- [x] Render-checked local and API footer layouts manually.
- [x] bun run check

### DIFF
--- a/src/cli/components/AgentInfoBar.tsx
+++ b/src/cli/components/AgentInfoBar.tsx
@@ -3,9 +3,12 @@ import Link from "ink-link";
 import { memo, useMemo } from "react";
 import stringWidth from "string-width";
 import type { ModelReasoningEffort } from "@/agent/model";
-import { buildAppUrl, buildChatUrl } from "@/cli/helpers/app-urls";
+import {
+  buildAppUrl,
+  buildChatUrl,
+  isLocalAgentId,
+} from "@/cli/helpers/app-urls";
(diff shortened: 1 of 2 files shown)


src/cli/components/AgentInfoBar.tsx
```diff
@@ -85,100 +146,87 @@ export const AgentInfoBar = memo(function AgentInfoBar({
...
-                `Open in ADE: ${adeConversationUrl} · Usage: ${buildAppUrl("/settings/organization/usage")}`,
-                rightWidth,
-              )}
+          <Box width={contentWidth} flexShrink={1}>
+            <Text wrap="wrap">
+              <Text bold color={colors.footer.agentName}>
+                {agentNameLine}
+              </Text>
+              {modelLine ? <Text dimColor> · {modelLine}</Text> : null}
             </Text>
```
amelia-letta, line 176: tmux users lose the actual URLs — old code showed them as plain text so you could copy. now there's no way to get the URL at all in tmux.

cpacker: Fixed in aa3b4ea0 — tmux now renders copyable plain-text ADE and usage URLs, while non-tmux terminals keep the compact clickable labels.

PR #2523, opened 2026-05-25
refactor: extract extension host kernel

carenthomas:
## Summary
- Move extension source loading/activation/disposal into a runtime-independent `src/extensions` host kernel
- Add owner metadata, activation generation, abort signals, diagnostics, and stale-handle guards for registered capabilities
- Wrap the existing CLI extension loader/runtime around the host while preserving command/statusline/panel behavior

## Test plan
- [x] `bun test src/extensions/extension-host.test.ts src/cli/extensions/local-extension-loader.test.ts src/cli/extensions/command-runtime.test.ts`
- [x] `bun run check`


src/extensions/extension-host.ts
```diff
@@ -0,0 +1,919 @@
...
+            path: diagnostic.path ?? diagnostic.owner?.path ?? "",
+            phase: diagnostic.phase,
+          });
+        }
+        publish();
+      },
+    });
+    loadingRegistry = nextRegistry;
+    activeRegistry = nextRegistry;
+    publish();
```
amelia-letta, line 923: Concurrent `reload()` calls can race — a stale load overwrites the current registry here because there is no generation check. If reload A (gen 1) and reload B (gen 2) run concurrently and B resolves first, A final assignment clobbers B registry.

```suggestion
    loadingRegistry = nextRegistry;
    if (loadGeneration === generation) {
      activeRegistry = nextRegistry;
      publish();
    }
```

PR #2538, opened 2026-05-26
feat(extensions): expose backend primitives

carenthomas:
## Summary
- Add an optional `ExtensionBackendApi` with fork and message-stream primitives and expose it to extensions as `letta.backend`
- Wire the TUI extension runtime to resolve the active backend for extension backend calls
- Update extension authoring docs and the `/btw` recipe to use backend primitives instead of the server SDK

## Test plan
- [x] `bun run check`
- [x] `bun test src/extensions/extension-host.test.ts src/cli/extensions/local-extension-loader.test.ts`
- [x] `bun src/skills/builtin/creating-skills/scripts/validate-skill.ts src/skills/builtin/creating-extensions`

### DIFF
--- a/src/cli/extensions/use-local-extension-runtime.ts
+++ b/src/cli/extensions/use-local-extension-runtime.ts
@@ -66,14 +66,13 @@ export function useLocalExtensionRuntime(
   const getContext = useCallback(() => contextRef.current, []);
 
   const backend = useMemo<ExtensionBackendApi>(() => {
-    const activeBackend = getBackend();
     return {
       forkConversation(conversationId, options) {
-        return activeBackend.forkConversation(conversationId, options);
+        return getBackend().forkConversation(conversationId, options);
       },
       sendMessageStream(conversationId, messages, options, requestOptions) {
         return sendMessageStreamWithBackend(
-          activeBackend,
+          getBackend(),
           conversationId,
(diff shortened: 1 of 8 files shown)


src/cli/extensions/use-local-extension-runtime.ts
```diff
@@ -63,13 +65,32 @@ export function useLocalExtensionRuntime(
...
+        return sendMessageStreamWithBackend(
+          activeBackend,
+          conversationId,
+          messages,
+          options,
+          requestOptions,
+        );
+      },
+    };
+  }, []);
```
amelia-letta, line 83: `getBackend()` is captured once in the closure, but `configureBackendMode()` replaces the module-level `backend` variable — after a backend switch this object holds a stale reference. Call `getBackend()` inside each method body instead so it resolves at invocation time.

```suggestion
  const backend = useMemo<ExtensionBackendApi>(() => {
    return {
      forkConversation(conversationId, options) {
        return getBackend().forkConversation(conversationId, options);
      },
      sendMessageStream(conversationId, messages, options, requestOptions) {
        return sendMessageStreamWithBackend(
          getBackend(),
          conversationId,
          messages,
          options,
          requestOptions,
        );
      },
    };
  }, []);
```

PR #2544, opened 2026-05-26
feat(extensions): add lifecycle events API

carenthomas:
## Summary
- add a typed extension event API with `letta.events.on/off` and lifecycle event registration/diagnostics
- emit `conversation_open` and `conversation_close` events for startup, new/resume/fork, and conversation close paths while keeping existing hooks in parallel
- document lifecycle events in the extension authoring skill

## Test plan
- [x] `bun test src/extensions/extension-host.test.ts src/cli/extensions/local-extension-loader.test.ts`
- [x] `bun src/skills/builtin/creating-skills/scripts/validate-skill.ts src/skills/builtin/creating-extensions`
- [x] `bun run check`
- [x] manual smoke: global extension opens a panel on `conversation_open`


src/extensions/types.ts
```diff
@@ -118,8 +123,80 @@ export interface ExtensionOwner {
   generation: number;
 }
 
+export type ExtensionEventName = "session_start" | "session_shutdown";
+
+export type ExtensionSessionStartReason =
```
amelia-letta: `"reload"` is in both `ExtensionSessionStartReason` and `ExtensionSessionShutdownReason` but never emitted — `/reload` unmounts the entire App via key change without calling `runEndHooks("reload")`, and the remount fires `session_start` with `reason: "startup"`. Either emit these or remove the dead values from the union types (and the docs).

carenthomas: addressed: renamed the lifecycle events to `conversation_open` / `conversation_close` and removed the un-emitted `reload` reason values from the types/docs.

PR #2550, opened 2026-05-27
feat(extensions): add turn_start event

carenthomas:
## Summary
- add a `turn_start` extension event with raw `input` passed to handlers before user-message turns are sent
- allow handlers to mutate `event.input` or return `{ input }`, with later handlers seeing the current input
- emit `turn_start` from the common TUI send boundary plus headless one-shot/bidirectional paths, and document the API in the creating-extensions skill
- roll back partial `event.input` mutations when a handler throws so diagnostics stay non-blocking

## Test plan
- [x] `bun test src/extensions/extension-host.test.ts src/extensions/extension-runtime.test.ts src/headless-extension-lifecycle.test.ts`
- [x] `bun run check`


amelia-letta (approved):
Clean implementation. No issues to flag.

Key checks:
- `extensionRuntime` already in `useCallback` deps
- Handler errors non-blocking in both TUI and headless
- Per-event capability gating correct
- Results typing with `NonNullable` handles `undefined` result types properly
- Bidirectional mode context scoped correctly to available params

PR #2552, opened 2026-05-27
feat(local): allow extensions to register pi providers

### LINKED ISSUE #2541: Capability mismatch on local OpenAI-compatible providers: VLM images silently dropped, context window misreported, reasoning flag wrong
## Problem

When connecting a local OpenAI-compatible engine (oMLX, LM Studio, vLLM, llama.cpp, etc.) via `letta connect lmstudio`, letta-code constructs each model's capability contract from a substring match on the model ID plus a few hardcoded fallbacks. There is no probe of the engine for actual capabilities. Three distinct user-visible symptoms result, all from the same root cause:

1. **Vision input silently dropped.** Multimodal models whose IDs don't contain `llava`/`vision`/`vl` are treated as text-only. Image content blocks are replaced with placeholder text _before_ the request leaves letta-code, and the model responds as if it never saw an image.
2. **Context window misreported by default.** When `model_settings.context_window_limit` isn't manually set on an agent, the factory falls back to a hardcoded 128K, even when the engine has the model configured for a larger window. The statusline and `/context` then show 128K. There's a manual escape hatch (`context_window_limit`), but auto-detection is wrong.
3. **Reasoning flag wrong.** Reasoning models whose IDs don't contain `gpt-oss`/`qwen3`/`deepseek-r1` are treated as non-reasoning, and vice versa.

The first one is the loud failure (the model says it can't see the image you sent); the other two degrade quietly.

## Concrete setup that hit this

- **letta-code** v0.26.2 (installed via `pnpm`)
- **macOS**, Apple Silicon
- **Local engine**: oMLX 0.3.9 ([jundot/omlx](https://github.com/jundot/omlx)) on `http://localhost:8000`
- **Model**: `gemma-4-26B-A4B-it-oQ6` — a Gemma-4 multimodal VLM. oMLX has it configured with a 256K context window and `model_type: "vlm"`.
- **Provider connection**: `letta connect lmstudio --base-url http://localhost:8000/v1`, producing the following `~/.letta/lc-local-backend/providers/auth.json` entry:
  ```json
  {
    "id": "local-provider-lmstudio",
    "name": "lmstudio",
    "provider_type": "lmstudio_openai",
    "provider_category": "byok",
    "auth": { "type": "api", "key": "1234" },
    "base_url": "http://localhost:8000/v1"
  }
  ```
- **Agent**: created via `letta --new-agent --backend local --model lmstudio/gemma-4-26B-A4B-it-oQ6`. The agent's `model_settings.provider_type` is correctly stamped as `"lmstudio_openai"`.

## Reproduction

1. Start oMLX with a VLM whose ID doesn't match the heuristic — `gemma-4-26B-A4B-it-oQ6` is a clean example; any Gemma/Pixtral/Phi-Vision/MiniCPM-V variant works the same way.
2. Connect it: `letta connect lmstudio --base-url http://localhost:8000/v1`.
3. Create an agent: `letta --new-agent --backend local --model lmstudio/gemma-4-26B-A4B-it-oQ6`.
4. Drop any image into the chat with a prompt like "what's in this picture?"
5. **Observed**: the model replies "I cannot see the image, as the image data was omitted from your message." (or similar). The actual HTTP payload that reaches oMLX contains no image — only a text placeholder.
6. **Observed**: the statusline / `/context` shows the context window as `?/128K` even though the server has this model at 256K.

## Root cause trace

The contract is published by **letta-code** at model construction time, then enforced by **`@earendil-works/pi-ai`** at message-serialization time. The mismatch happens at the publish step.

### Who owns what

```
┌─ letta-code (this repo) ─────────────────────────┐  ┌─ @earendil-works/pi-ai (npm dep) ──┐
│ - provider config, discovery                     │  │ - Model<TApi> type definition       │
│ - customOpenAICompatibleModel factory            │──│ - transformMessages pipeline        │
│ - publishes Model contract                       │  │ - downgradeUnsupportedImages        │
│ - calls into pi-ai to send messages              │  │ - openai-completions serializer     │
└──────────────────────────────────────────────────┘  └─────────────────────────────────────┘
                                                                       │
                                                                       ▼
                                                      ┌─ oMLX / LM Studio over HTTP ────────┐
                                                      │  POST /v1/chat/completions          │
                                                      │  → gemma-4-26B-A4B-it-oQ6           │
                                                      └─────────────────────────────────────┘
```

**Direction**: letta-code **publishes** the capability contract; pi-ai **enforces** it. The mismatch happens at publish time, the user-visible failure happens at enforcement time — which is why the symptom (model "can't see" the image) looks like a model problem rather than a configuration problem.

### The trace, end to end

1. `letta connect lmstudio --base-url http://localhost:8000/v1`
   → letta-code: `src/backend/local/local-model-config.ts` runs `discoverOpenAICompatibleModelIds()` — hits `/v1/models`, gets back IDs only (OpenAI spec returns nothing else).

2. For each discovered ID, letta-code calls `customOpenAICompatibleModel(input)`
   → `src/backend/dev/pi-model-factory.ts` decides `input: ("text" | "image")[]` via the substring heuristic. `gemma-4-26B-A4B-it-oQ6` doesn't match `llava`/`vision`/`vl` → resolves to **`["text"]`**.

3. That Model object is constructed against `@earendil-works/pi-ai`'s contract
   → `pi/packages/ai/src/types.ts:550`: `input: ("text" | "image")[]` — letta-code now holds a Model claiming gemma is text-only.

4. User drops an image into the chat. letta-code hands `(Model, messages[])` to pi-ai.

5. pi-ai runs `transformMessages` before serializing
   → `pi/packages/ai/src/providers/transform-messages.ts:64` — which at line 71 calls `downgradeUnsupportedImages(messages, model)`.

6. That function checks `model.input.includes("image")` → false (we said text-only at step 2)
   → `transform-messages.ts:35-57`: walks every message, calls `replaceImagesWithPlaceholder(content, NON_VISION_USER_IMAGE_PLACEHOLDER)`.

7. The placeholder constant
   → `transform-messages.ts:12`: `NON_VISION_USER_IMAGE_PLACEHOLDER = "(image omitted: model does not support images)"`.

8. pi-ai's openai-completions serializer runs on the now-mutated message list
   → `pi/packages/ai/src/providers/openai-completions.ts:940`: the second gate, `if (hasImages && model.input.includes("image"))` — same check, second layer of defense. Tool-result images get stripped here too.

9. HTTP POST to oMLX with a text-only body. Gemma processes `"(image omitted: model does not support images)"` as literal user text.

10. Gemma — actually multimodal — replies as if no image arrived: "I cannot see the image, as the image data was omitted from your message." The captured HTTP body (Repro step 5) confirms the strip happened before the wire.

### letta-code side (the publisher)

[`src/backend/local/local-model-config.ts`](https://github.com/letta-ai/letta-code/blob/main/src/backend/local/local-model-config.ts) runs `discoverOpenAICompatibleModelIds()` against the engine's `/v1/models`. The OpenAI spec for that endpoint returns only identification fields (`id`, `object`, `created`, `owned_by`) — no capability metadata — so the discovery step only learns _which_ models exist, not _what they can do_.

[`src/backend/dev/pi-model-factory.ts`](https://github.com/letta-ai/letta-code/blob/main/src/backend/dev/pi-model-factory.ts) `customOpenAICompatibleModel(...)` then has to invent the rest of the `Model` object:

```ts
input:
  input.modelId.includes("llava") ||
  input.modelId.includes("vision") ||
  input.modelId.includes("vl")
    ? ["text", "image"]
    : ["text"],
contextWindow: input.contextWindow ?? 128000,
maxTokens: input.maxTokens ?? 32000,
reasoning:
  input.modelId.includes("gpt-oss") ||
  input.modelId.includes("qwen3") ||
  input.modelId.includes("deepseek-r1"),
```

The substring lists are an honest answer to a hard problem — but they miss every VLM that doesn't happen to advertise its modality in its name (Gemma, Pixtral, Phi-Vision, Llama-3.2-Vision, MiniCPM-V, …), and the 128K fallback silently misrepresents most non-OpenAI models.

### pi-ai side (the enforcer)

The published `input: ["text"]` then flows into `@earendil-works/pi-ai`, which enforces it strictly:

- [`packages/ai/src/types.ts:550`](https://github.com/earendil-works/pi/blob/main/packages/ai/src/types.ts#L550) — `Model.input: ("text" | "image")[]` is the contract.
- [`packages/ai/src/providers/transform-messages.ts:12`](https://github.com/earendil-works/pi/blob/main/packages/ai/src/providers/transform-messages.ts#L12) — `const NON_VISION_USER_IMAGE_PLACEHOLDER = "(image omitted: model does not support images)";`
- [`packages/ai/src/providers/transform-messages.ts:35-57`](https://github.com/earendil-works/pi/blob/main/packages/ai/src/providers/transform-messages.ts#L35-L57) — `downgradeUnsupportedImages()`: when `model.input.includes("image")` is false, walks every message and replaces image content blocks with that placeholder text.
- [`packages/ai/src/providers/openai-completions.ts:940`](https://github.com/earendil-works/pi/blob/main/packages/ai/src/providers/openai-completions.ts#L940) — a second-layer gate on the same check, for tool-result images.

That placeholder string is literally what the model reads. The "I cannot see the image, as the image data was omitted from your message" reply is consistent with gemma receiving no image; the captured HTTP body (Repro step 5) confirms the strip happens client-side before the wire.

## Candidate solutions

Both fit the user-facing problem; both have precedent. No preference here — flagging the trade-offs.

### Option A — make the built-in discovery smarter

Extend `discoverOpenAICompatibleModelIds()` to also probe well-known non-standard capability endpoints when present:

- **oMLX** — `GET /v1/models/status` returns per-model `model_type` (`llm`/`vlm`/`embeddings`), `max_context_window`, `max_tokens`, `thinking_default`.
- **LM Studio** — `GET /api/v0/models` ([docs](https://lmstudio.ai/docs/developer/rest/endpoints)) returns per-model `type` (`llm`/`vlm`/`embeddings`), `max_context_length`, `arch`, `state`, `quantization`.
- **llama.cpp** — `GET /v1/models` is gaining `architecture.input_modalities` (see [ggml-org/llama.cpp#22952](https://github.com/ggml-org/llama.cpp/pull/22952) referenced in [earendil-works/pi#3357](https://github.com/earendil-works/pi/issues/3357)).

Field names differ across engines (`model_type` vs `type`, `max_context_window` vs `max_context_length`) — the probe would need a thin normalizer. If no enrichment endpoint responds, fall back to the current heuristic + defaults.

Pair with a **user-facing escape hatch** for misdetection (per-agent override in `model_settings` to force `supportsImage: true` / `reasoning: true` / custom `contextWindow`), so the rare case where a probe is wrong doesn't trap users.

- **Pros**: minimal architectural change, ships behind existing discovery code, fixes the immediate case across every supported engine in one place.
- **Cons**: every new engine quirk (different field names, new metadata shapes) lands on letta-code maintainers. Re-invents work that the pi ecosystem has already done in extensions.

### Option B — let letta-code load pi provider extensions

The pi maintainers' [endorsed direction in earendil-works/pi#3357](https://github.com/earendil-works/pi/issues/3357) is per-engine custom-provider **extensions**, registered via `pi.registerProvider(name, config)`. Several already exist:

- [`huggingface/pi-llama`](https://github.com/huggingface/pi-llama) — semi-official llama.cpp extension (Julien Chaumond + Célina Hanouti)
- [`CaptCanadaMan/pi-ollama`](https://github.com/CaptCanadaMan/pi-ollama) — native `/api/chat` Ollama
- [`jgsheppa/pi-find-your-llama`](https://github.com/jgsheppa/pi-find-your-llama) — llama.cpp with model registration UX
- [`vampaz/pi-custom-provider-lmstudio`](https://github.com/vampaz/pi-custom-provider-lmstudio) — LM Studio (directly relevant here: the bug above is on `provider_type: "lmstudio_openai"`)
- [`adrianhoehne/pi-agent-llama-router-model-detector`](https://github.com/adrianhoehne/pi-agent-llama-router-model-detector) — llama.cpp router-mode

Letta-code today depends only on `@earendil-works/pi-ai` (the headless library) — _not_ `@earendil-works/pi-coding-agent` (which owns the `registerProvider` extension API). So none of these extensions can be installed in letta-code today, even though letta-code's own extension system (statusline / tool extensions in `~/.letta/extensions/`) is already in place and well-loved.

The change here is to wire pi's provider-extension lifecycle into letta-code's extension host, so users can `pnpm i -g huggingface/pi-llama` (or similar) and have it handle the model construction for their engine of choice.

- **Pros**: aligns letta-code with the maintainer-endorsed pi direction; per-engine quirks live with engine specialists; future-proof as `/v1/models` and friends expose more metadata; community can ship fixes faster than letta-code can.
- **Cons**: bigger lift; introduces dependency on `@earendil-works/pi-coding-agent` or a parallel loader; new surface area to maintain.

### Composite

(A) gets users unstuck quickly; (B) is the longer-term architectural fit. They aren't mutually exclusive — (A) could ship now as the default-good-experience while (B) lands as the extensibility story.

## Environment

- letta-code: v0.26.2
- pi-ai version: v0.75.5 (per letta-code `package.json` dependency range `^0.75.5`)
- OS: macOS, Apple Silicon
- Local engine: oMLX 0.3.9
- Reproduces with: `gemma-4-26B-A4B-it-oQ6` (Gemma-4 VLM)
- Should reproduce with any non-`llava`/`vision`/`vl`-named VLM served via any OpenAI-compat local engine

cpacker:
## Summary
- Add a pi-mono-style `letta.registerProvider()` / `letta.providers.register()` extension API for registering pi provider model contracts.
- Make local backend pi model resolution consume registered model capabilities (`input`, `reasoning`, `contextWindow`, `maxTokens`, headers/base URL) before falling back to heuristic custom OpenAI-compatible models.
- Surface registered local provider models and context windows through local model listing so `/model` and context-window defaults can use extension metadata.

## Fixes
- Fixes #2541

## Test plan
- [x] `bun test src/backend/pi-model-factory.test.ts src/backend/local-backend.test.ts src/extensions/extension-host.test.ts src/headless-extension-lifecycle.test.ts`
- [x] `bun run typecheck`
- [x] `bun run check`

### DIFF
--- a/src/backend/dev/pi-model-factory.ts
+++ b/src/backend/dev/pi-model-factory.ts
@@ -441,8 +441,8 @@ export async function resolvePiModelForAgent(
         timeout: resolveLocalProviderTimeout({ providerIds: [provider] }),
       };
   let baseURL =
-    registeredProvider?.config.baseUrl ??
     connection.baseURL ??
+    registeredProvider?.config.baseUrl ??
     spec?.baseUrlEnv?.() ??
     spec?.defaultBaseURL;
   let headers = mergeHeaders(
--- a/src/backend/dev/pi-model-factory.ts
+++ b/src/backend/dev/pi-model-factory.ts
@@ -15,6 +15,14 @@ import {
   resolveLocalProviderTimeout,
 } from "@/backend/local/local-provider-timeout";
 import {
+  getRegisteredPiProvider,
+  type PiProviderModelRegistration,
+  type PiProviderRegistration,
+  resolveRegisteredPiProviderApiKey,
+  resolveRegisteredPiProviderFromModelHandle,
+  stripRegisteredProviderHandlePrefix,
+} from "./pi-provider-extension-registry";
+import {
   expectedPiProviderList,
   getPiProviderSpec,
   isPiProvider,
@@ -93,6 +101,7 @@ export function resolvePiProvider(
     inferDefaultProviderFromStandardKeys(),
 ): PiProvider {
   if (isPiProvider(provider)) return provider;
+  if (getRegisteredPiProvider(provider)) return provider as PiProvider;
   throw new Error(
     `Unknown pi provider "${provider}". Expected ${expectedPiProviderList()}.`,
   );
@@ -103,6 +112,9 @@ export function resolvePiProviderFromAgent(
   modelSettings: PiModelSettings = {},
 ): PiProvider {
   return (
+    (resolveRegisteredPiProviderFromModelHandle(model) as
+      | PiProvider
+      | undefined) ??
     resolveProviderFromModelHandle(model) ??
     resolveProviderFromProviderType(modelSettings.provider_type) ??
     resolvePiProvider()
@@ -284,6 +296,61 @@ function withOverrides(
   };
 }
 
+function mergeHeaders(
+  ...headers: Array<Record<string, string> | undefined>
+): Record<string, string> | undefined {
+  const merged: Record<string, string> = {};
+  for (const header of headers) {
+    if (!header) continue;
+    Object.assign(merged, header);
+  }
+  return Object.keys(merged).length > 0 ? merged : undefined;
+}
+
+function withAuthHeader(
+  headers: Record<string, string> | undefined,
+  apiKey: string | undefined,
+  authHeader: boolean | undefined,
+): Record<string, string> | undefined {
+  if (!authHeader || !apiKey) return headers;
+  return {
+    ...headers,
+    Authorization: `Bearer ${apiKey}`,
+  };
+}
+
+function registeredModelToPiModel(input: {
+  providerName: string;
+  config: PiProviderRegistration;
+  model: PiProviderModelRegistration;
+  baseURL?: string;
+  headers?: Record<string, string>;
+}): Model<Api> {
+  const api = input.model.api ?? input.config.api;
+  if (!api) {
+    throw new Error(
+      `Provider "${input.providerName}" model "${input.model.id}" is missing an api`,
+    );
+  }
+  return {
+    id: input.model.id,
+    name: input.model.name,
+    api,
+    provider: input.providerName,
+    baseUrl: input.model.baseUrl ?? input.baseURL ?? "",
+    reasoning: input.model.reasoning,
+    ...(input.model.thinkingLevelMap
+      ? { thinkingLevelMap: input.model.thinkingLevelMap }
+      : {}),
+    input: input.model.input,
+    cost: input.model.cost,
+    contextWindow: input.model.contextWindow,
+    maxTokens: input.model.maxTokens,
+    ...(input.headers ? { headers: input.headers } : {}),
+    ...(input.model.compat ? { compat: input.model.compat } : {}),
+  } as Model<Api>;
+}
+
 function numericSetting(value: unknown): number | undefined {
   return typeof value === "number" && Number.isFinite(value)
     ? value
@@ -342,11 +409,16 @@ export async function resolvePiModelForAgent(
   const provider = options.provider
     ? resolvePiProvider(options.provider)
     : resolvePiProviderFromAgent(modelHandle, modelSettings);
-  const spec = getPiProviderSpec(provider);
+  const registeredProvider = getRegisteredPiProvider(provider);
+  const spec = isPiProvider(provider) ? getPiProviderSpec(provider) : undefined;
   const modelId =
     options.model ??
-    resolvePiModelFromAgent(modelHandle, provider) ??
-    resolvePiModelFromAgent(spec.defaultModel, provider) ??
+    (registeredProvider
+      ? stripRegisteredProviderHandlePrefix(modelHandle, provider)
+      : undefined) ??
+    (spec ? resolvePiModelFromAgent(modelHandle, spec.id) : undefined) ??
+    registeredProvider?.config.models?.[0]?.id ??
+    (spec ? resolvePiModelFromAgent(spec.defaultModel, spec.id) : undefined) ??
     process.env.LETTA_CODE_DEV_PI_MODEL ??
     "";
   const storageDir = options.localProviderAuthStorageDir;
@@ -355,14 +427,28 @@ export async function resolvePiModelForAgent(
       ? modelSettings.provider_type
       : options.preferredProviderType;
 
-  let connection = localProviderConnection(
-    spec.localProviderNames,
-    spec.apiKeyEnv?.() ?? spec.fallbackApiKey,
-    storageDir,
+  const registeredApiKey = resolveRegisteredPiProviderApiKey(
+    registeredProvider?.config.apiKey,
   );
+  let connection = spec
+    ? localProviderConnection(
+        spec.localProviderNames,
+        registeredApiKey ?? spec.apiKeyEnv?.() ?? spec.fallbackApiKey,
+        storageDir,
+      )
+    : {
+        apiKey: registeredApiKey,
+        timeout: resolveLocalProviderTimeout({ providerIds: [provider] }),
+      };
   let baseURL =
-    connection.baseURL ?? spec.baseUrlEnv?.() ?? spec.defaultBaseURL;
-  const headers = spec.headers?.();
+    registeredProvider?.config.baseUrl ??
+    connection.baseURL ??
+    spec?.baseUrlEnv?.() ??
+    spec?.defaultBaseURL;
+  let headers = mergeHeaders(
+    spec?.headers?.(),
+    registeredProvider?.config.headers,
+  );
   let providerOptions: Record<string, unknown> | undefined;
   let envOverrides: Record<string, string | undefined> | undefined;
   let oauthCredentials: OAuthCredentials | undefined;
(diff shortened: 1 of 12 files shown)


carenthomas:
overall this direction looks good to me. `letta.registerProvider()` / `letta.providers.register()` feels like the right API shape for letting extensions publish local provider/model capabilities instead of hardcoding every provider/model heuristic in app internals.

things I like:
- provider ownership/disposal is wired, so reload/dispose unregisters provider registrations
- local model resolution prefers registered metadata for vision/reasoning/context window/max tokens
- `listLocalModels()` surfaces registered models with context windows, so `/model` can benefit
- headless enables `providers: true`, which matters now that headless loads extensions before tool/model use
- tests cover registration, local listing, and pi model resolution

one substantive concern: I think configured local provider base URL should probably override the extension default. Current precedence looks like:

```ts
baseURL =
  registeredProvider?.config.baseUrl ??
  connection.baseURL ??
  spec?.baseUrlEnv?.() ??
  spec?.defaultBaseURL;
```

I think this should probably be:

```ts
baseURL =
  connection.baseURL ??
  registeredProvider?.config.baseUrl ??
  spec?.baseUrlEnv?.() ??
  spec?.defaultBaseURL;
```

The extension registration is provider/model metadata, but the user’s local provider connection is runtime configuration. If I configured LM Studio at `http://127.0.0.1:5678/v1`, installing an extension with LM Studio model metadata should not silently route me back to the extension’s default `baseUrl`.

The local listing test sets up a stored provider base URL, but it only exercises `listLocalModels()`, not `resolvePiModelForAgent()` with that storage dir, so I think this precedence issue is currently uncovered. I’d add a small test where registered `lmstudio` has `baseUrl: localhost:8000`, local provider record has `baseURL: 127.0.0.1:1234`, and the resolved model uses `127.0.0.1:1234`.

Issue #2572, opened 2026-05-27
Statusline shows 128K context for extension-registered providers (boot race + rejoin staleness)


jasikpark:
## Summary

When a provider is contributed via the `letta.registerProvider()` API (added in #2552), the statusline shows `128k` for that provider's models even when the extension declares a larger `contextWindow`. The chat-completion path correctly honors the registered value (routing works), but `projectLocalAgentState`'s `llm_config` shim hardcodes a 128K fallback that the UI surfaces in the statusline.

Reference extension used during investigation: https://gist.github.com/jasikpark/fd6683f497b278ad55838524f383f2f6 (oMLX provider — declares `gemma-4-26B-A4B-it-oQ6` at 262144).

## Local minimum fix

I'm running a local patch that adds a registry lookup in `projectLocalAgentState` (`src/backend/local/local-store.ts`), stripping the provider prefix from the model handle and using the registered `contextWindow` before falling back to 128000 (mirrors the lookup already in `pi-model-factory.ts`). Happy to share the diff if useful, but it only fixes the projection layer — see remaining bugs below.

## Remaining bug #1 — boot race

Cold-boot ordering puts the first agent-state projection **before** the extension's async `activate()` finishes:

```
T+0ms     [local-store] no provider matched      → 128000   ← initial agent load
T+66ms    [extension] activate called             (extension fires AFTER)
T+114ms   registerProvider returned successfully
T+1.9s    [local-store] lookup result            → 262144   ← subsequent reads
```

The UI snapshots `llmConfig.context_window = 128000` from the initial projection. Subsequent projections return the correct value, but `AppCoordinator.agentState` only refreshes when `initialAgentState` (the prop) changes — which doesn't happen after extension registration. Statusline shows `128k` until something triggers a fresh agent-state load (e.g., switching models).

## Remaining bug #2 — rejoin staleness

When rejoining an existing conversation, the statusline reads `0/128k` until the first request completes. May be partly fundamental (no live token count until model is hit), but the context-window denominator should be correct on mount, not require a request to refresh.

## Possible fix paths

**A. Block initial agent projection until extensions finish activating.** Simplest semantics, slows cold boot by however long the slowest extension takes (~70ms here).

[...]

PR #2575, opened 2026-05-27
fix(local): report registered model context windows

cpacker:
## Summary
- Thread extension-registered local model `contextWindow` / `maxTokens` metadata into local agent state so statusline and `/context` reporting do not get stuck on the legacy 128k fallback.
- Stop stamping new local default agents with an unconditional 128k context limit; use registered model metadata when available and keep 128k only as compatibility fallback.
- Clear the available-model cache when extension providers are registered/unregistered so `/model` and model switching see fresh provider metadata.

## Test plan
- [x] `bun test src/backend/local-backend.test.ts src/extensions/extension-host.test.ts src/backend/pi-model-factory.test.ts`
- [x] `bun run check`

### DIFF
--- a/src/backend/dev/pi-provider-extension-registry.ts
+++ b/src/backend/dev/pi-provider-extension-registry.ts
@@ -39,7 +39,30 @@ export interface RegisteredPiProvider {
   path?: string;
 }
 
+type PiProviderRegistryListener = () => void;
+
 const registeredProviders = new Map<string, RegisteredPiProvider>();
+const registryListeners = new Set<PiProviderRegistryListener>();
+
+function notifyRegistryListeners(): void {
+  for (const listener of [...registryListeners]) {
+    try {
+      listener();
+    } catch {
+      // Registry listeners are observers; a UI refresh failure should not make
+      // extension provider registration fail.
+    }
+  }
+}
+
+export function subscribePiProviderRegistry(
+  listener: PiProviderRegistryListener,
+): () => void {
+  registryListeners.add(listener);
+  return () => {
+    registryListeners.delete(listener);
+  };
+}
 
 function cloneHeaders(
   headers: Record<string, string> | undefined,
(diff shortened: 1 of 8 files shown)


jasikpark:
Followup from [#2576](https://github.com/letta-ai/letta-code/pull/2576) (the draft I opened from [#2572](https://github.com/letta-ai/letta-code/issues/2572) before seeing this PR — now closed in favor of yours).

Tested this branch locally against my oMLX extension (registers `lmstudio/gemma-4-26B-A4B-it-oQ6` with `contextWindow: 262144`). On cold boot the header still shows `0/128k` until `/reload`. Added trace logging in `projectableAgentRecord` and `localModelSettingsForHandle`:

```
T+0.000s  projectableAgentRecord → "no defaults (registry miss)" (agent #1)
T+0.049s  projectableAgentRecord → "no defaults (registry miss)" (agent #2, lmstudio/gemma-4-26B-A4B-it-oQ6)
T+0.122s  extension activate #1 starts
T+0.177s  extension activate #2 starts
T+0.180s  registerProvider(lmstudio) #1 returns
T+0.210s  registerProvider(lmstudio) #2 returns
── silence ── (no further retrieveAgent / projection calls; UI stays at 128k until manual /reload)
```

`hasModelSettingsForModel: true` in those entries, so the wiring is right — the registry is just empty at the moment projection runs. `projectLocalAgentState` finishes ~120ms before the extension's `activate()` registers.

Projection-time lookup here fixes `/reload` and stored legacy 128K (which #2576 didn't), so it's strictly better there. The remaining gap is the first-paint race. One option from #2576 that closes it: subscribe/notify on the pi provider registry (notify on register/unregister/dispose) + a `useEffect` in `LoadingApp` that re-calls `retrieveAgent` on mutation, so projection re-runs once registration lands.

cpacker:
@jasikpark just updated, can you try it out?

jasikpark:
@cpacker looking good, got the correct context on first startup:

image

cpacker:
> @cpacker looking good, got the correct context on first startup:
> 
> image

perfect, will merge once CI is clean

PR #2576, opened 2026-05-28
fix(local): re-project agent state on post-boot provider register

jasikpark:
## Summary

Follow-up to `fix(local): honor registered provider contextWindow in llm_config shim` (already on this branch).

Extensions that register Pi providers via the async `activate()` hook finish *after* the initial `projectLocalAgentState()` runs, so the projected `context_window` is computed before `registeredModelContextWindow()` has any registered provider to consult — UI falls back to the 128K default even when the registered model declares 262K.

This adds:

1. `subscribePiProviderRegistry(listener)` in `pi-provider-extension-registry.ts`, with `notifyPiProviderRegistry()` fired from `register`/`unregister`/`unregisterForOwner`/`clear` (guarded so no-op iterations don't notify).
2. A `useEffect` in `LoadingApp` that subscribes once `loadingState === "ready"` and re-runs `retrieveAgent` → `setAgentState` on every registry mutation, so the projected `context_window` re-resolves.

Refs https://github.com/letta-ai/letta-code/issues/2572.

## Status

Works on `/reload` paths (verified: UI flips from 128K to 262K once `omlx-provider.ts` finishes registering `lmstudio`).

**Cold boot is still broken** — even though `subscribePiProviderRegistry` fires correctly, a teardown immediately after the extension's second `activate()` empties the registry, so the re-projection sees no providers and falls back to 128K again. Two suspected culprits, neither investigated here:

- `omlx-provider.ts` activates twice on cold boot (same `scope:path` owner id, ~40ms apart) — unclear whether multiple sources resolve the same path or whether load-then-reload is the normal cold-boot sequence.
- `disposeLocalExtensions()` in `extension-host.ts` clears providers for every owner during the second activate's teardown phase, regardless of ownership.

Leaving cold-boot diagnosis to @cpacker (handoff comment incoming on #2572).

### DIFF
--- a/src/backend/dev/pi-provider-extension-registry.ts
+++ b/src/backend/dev/pi-provider-extension-registry.ts
@@ -41,6 +41,28 @@ export interface RegisteredPiProvider {
 
 const registeredProviders = new Map<string, RegisteredPiProvider>();
 
+type RegistryListener = () => void;
+const registryListeners = new Set<RegistryListener>();
+
+function notifyPiProviderRegistry(): void {
+  for (const listener of [...registryListeners]) {
+    try {
+      listener();
+    } catch {
+      // ignore listener errors so one bad listener can't block the rest
+    }
+  }
+}
+
+export function subscribePiProviderRegistry(
+  listener: RegistryListener,
+): () => void {
+  registryListeners.add(listener);
+  return () => {
+    registryListeners.delete(listener);
+  };
+}
+
 function cloneHeaders(
   headers: Record<string, string> | undefined,
 ): Record<string, string> | undefined {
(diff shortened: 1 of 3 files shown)


jasikpark:
Closing in favor of #2575 — @cpacker's approach (projection-time lookup inside `LocalStore.projectableAgentRecord` + cache clear on register/unregister) sidesteps the cold-boot race that this PR's reactive useEffect approach couldn't solve, and also upgrades the legacy 128K stamped onto existing stored agent records at projection time.

Leaving the trace from [#2572](https://github.com/letta-ai/letta-code/issues/2572#issuecomment-4559772072) in place as the diagnostic record.

PR #2583, opened 2026-05-28
fix(telemetry): reliably populate reflection_agent_id and persist events to disk

christinatong01:
## Summary
- Wait up to 30s (was 1s) for the spawned reflection subagents init event before emitting `reflection_start`, so `reflection_agent_id` lands populated. New exported constant `REFLECTION_AGENT_ID_WAIT_MS` in `src/tools/impl/task.ts` threaded through all four reflection launch sites (TUI manual + auto, websocket manual + auto).
- Emit a new `reflection_agent_id` field on `letta_code:reflection_start` and `letta_code:reflection_end`. Same value also emitted as `subagent_id` for back-compat with the existing PostHog dashboard.
- Persist telemetry events to `~/.letta/telemetry-queue.jsonl` on every `track()` and replay on next startup so reflection events survive abnormal exits / crashes. Path overridable via `LETTA_CODE_TELEM_QUEUE` (set to `0` to disable; tests do this).

## Test plan
- [x] `bun run check` (8/8 PASS)
- [x] `bun test src/telemetry/ src/tools/task-background-helper.test.ts src/tools/task-background.test.ts` (37/37 PASS)
- [ ] Manually verify a `reflection_start` event in PostHog includes `reflection_agent_id` after deploy
- [ ] Manually verify `reflection_end` matches the same `reflection_agent_id`
- [ ] Manually verify `~/.letta/telemetry-queue.jsonl` is created, drained on next startup, and removed after a successful flush
- [ ] Confirm the existing PostHog dashboard (keyed on `subagent_id`) still renders

### DIFF
--- a/src/tools/impl/task.ts
+++ b/src/tools/impl/task.ts
@@ -57,6 +57,20 @@ interface TaskArgs {
 const VALID_DEPLOY_TYPES = new Set(["general-purpose"]);
 const BACKGROUND_STARTUP_POLL_MS = 50;
 
+/**
+ * Generous upper bound (ms) used by reflection launch sites when awaiting
+ * the spawned subagent's Letta `agent_id` before emitting telemetry.
+ *
+ * Background: `reflection_start` telemetry needs `reflection_agent_id`
+ * populated so downstream pipelines (PostHog dashboards, letta-train)
+ * can join exactly instead of by ±15-min timestamp window. The init
+ * event that publishes `agent_id` normally lands in <500ms, but spawn
+ * failures or backend stalls can take longer. A bounded wait keeps
+ * telemetry reliable without risking a user-visible hang if the spawn
+ * never makes progress.
+ */
+export const REFLECTION_AGENT_ID_WAIT_MS = 30_000;
+
 type TaskRunResult = {
   agentId: string;
   conversationId?: string;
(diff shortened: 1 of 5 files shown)


src/tools/impl/task.ts
```diff
@@ -57,6 +57,20 @@ interface TaskArgs {
...
+ *
+ * Background: `reflection_start` telemetry needs `reflection_agent_id`
+ * populated so downstream pipelines (PostHog dashboards, letta-train)
+ * can join exactly instead of by ±15-min timestamp window. The init
+ * event that publishes `agent_id` normally lands in <500ms, but spawn
+ * failures or backend stalls can take longer. A bounded wait keeps
+ * telemetry reliable without risking a user-visible hang if the spawn
+ * never makes progress.
+ */
+export const REFLECTION_AGENT_ID_WAIT_MS = 30_000;
```
amelia-letta, line 72: 30s is a long blocking wait in the auto-trigger paths (step-count, compaction-event) — if the backend stalls, the user's next turn is delayed for half a minute. Normal case is <500ms so even 5-10s gives plenty of headroom without risking a user-visible hang on failure.

```suggestion
export const REFLECTION_AGENT_ID_WAIT_MS = 10_000;
```

PR #2588, opened 2026-05-29
feat(cli): add skill management commands

sarahwooders:
## Summary
- Add `letta install <skill>` to install skill folders into an agent memfs repository under `memory/skills/<skill-name>`.
- Add `letta skills list` and `letta skills delete <skill_name>` for installed agent skills.
- Reuse the interactive `/agents` selector UI for install/list when no agent id/name is provided, instead of dumping a numbered 100-agent list.
- Require an explicit `--agent`/`--agent-id` for destructive `skills delete` and emit a clear error if the user passes a source specifier instead of an installed skill directory name.
- Add compatibility source parsing for Hermes-style official skills, GitHub skill directories, and ClawHub registry skills.
- Support both cloud agents and local agents; local agent IDs install into the local backend memfs path instead of trying to clone a cloud git repo.

## Syntax and compatibility layer
The command intentionally mirrors the simple Hermes/OpenClaw skill-install shape while targeting Letta Code agent memory:

```bash
letta install <skill> [--agent <id> | --agent-id <id> | -n <agent name>] [--force]
letta skills list [--agent <id> | --agent-id <id> | -n <agent name>]
letta skills delete <skill_name> --agent <id>
```

Install/list resolve the target agent by explicit `--agent`/`--agent-id`, exact `-n/--name`, env (`LETTA_AGENT_ID`/`AGENT_ID`), then the shared Ink-based agent selector. The selector uses the same pinned/local/constellation tab UI as `/agents`, with skill-command-specific affordances: no New tab, no delete-agent shortcut, and no unpin shortcut.

Delete intentionally requires explicit `--agent`/`--agent-id`; it does not use env or selector fallback because it is destructive. Delete also expects the installed skill directory name (for example `meme-generation`), not an install source specifier like `official/creative/meme-generation`.

Supported install source syntaxes:

- Hermes official optional skills:
  - `official/finance/stocks`
  - Resolves to `https://github.com/NousResearch/hermes-agent.git`, subdirectory `optional-skills/finance/stocks`.
- GitHub repositories/directories:
  - `https://github.com/owner/repo`
  - `https://github.com/owner/repo/tree/main/path/to/skill`
  - `https://github.com/owner/repo/blob/main/path/to/skill/SKILL.md`
  - `owner/repo/path/to/skill`
  - Branch/path ambiguity is handled by checking remote branch names before choosing the skill subdirectory.
- ClawHub registry skills:
  - `clawhub/<slug>`
  - `clawhub:<slug>`
  - `clawhub:<slug>@<version>`
  - `https://clawhub.ai/skills/<slug>`
  - Resolves metadata from `https://clawhub.ai/api/v1/skills/<slug>`, picks `latestVersion.version`/`tags.latest` when no version is specified, then downloads the ZIP bundle from `/api/v1/download`.

After fetching any supported source, the installer normalizes everything into the same compatibility layer: a temporary local directory containing `SKILL.md` and supporting files. The final install path is determined from `SKILL.md` frontmatter `name` when available, falling back to the source directory name. This keeps the memfs layout source-agnostic and allows existing skills to be replaced only with `--force`.

Cloud agents use the git-backed cloud memfs checkout. Local agents use `~/.letta/lc-local-backend/memfs/<agent-id>/memory`, initializing a local memory repo if needed, so `agent-local-*` IDs do not hit the cloud git endpoint.

`letta skills list` reads installed directories under `memory/skills`, parses `SKILL.md` frontmatter when available, and emits JSON with names, descriptions, and paths. `letta skills delete <skill_name> --agent <id>` removes the matching `memory/skills/<skill_name>` directory after validating the skill name and path containment.

Note: the old nested `letta skills install ...` route is still accepted as a compatibility alias, but help text and examples advertise `letta install ...`.

## Test plan
- [x] `bun test src/cli/subcommands/skills.test.ts`
- [x] `bun run check`
- [x] `letta skills delete official/creative/meme-generation` fails before selecting an agent and asks for `--agent`.
- [x] `letta skills delete official/creative/meme-generation --agent <id>` fails with a clear installed-name-vs-source-specifier error.
- [x] Live-installed `official/finance/stocks` via `letta install ...` into a cloud agent memfs and verified `skills/stocks/SKILL.md` plus script files.
- [x] Live-installed `official/finance/stocks` via `letta install ...` into `agent-local-59e6bd2b-0696-4b8a-92bf-2e6e2ffbfb40` and verified it writes to the local backend memfs path.
- [x] Live-installed `clawhub/nano-banana-pro` into agent memfs and verified `skills/nano-banana-pro/SKILL.md`, `_meta.json`, `skill-card.md`, and script files.
- [x] Live-ran `letta skills list --agent <id>` and verified installed skills are returned as JSON.

### DIFF
--- a/src/cli/subcommands/skills.ts
+++ b/src/cli/subcommands/skills.ts
@@ -41,7 +41,7 @@ function printUsage(): void {
   console.log(
     `
 Usage:
-  letta skills install <skill> [--agent <id> | -n <name>] [--force]
+  letta install <skill> [--agent <id> | -n <name>] [--force]
 
 Sources:
   official/<path>         Hermes official optional skill, e.g. official/finance/stocks
@@ -546,7 +546,7 @@ async function installSkill(
   }
 }
 
-export async function runSkillsSubcommand(argv: string[]): Promise<number> {
+async function runInstall(argv: string[]): Promise<number> {
   let parsed: ReturnType<typeof parseSkillsArgs>;
   try {
     parsed = parseSkillsArgs(argv);
@@ -558,20 +558,14 @@ export async function runSkillsSubcommand(argv: string[]): Promise<number> {
     return 1;
   }
 
-  const [action, specifier] = parsed.positionals;
-  if (parsed.values.help || !action || action === "help") {
+  const [specifier] = parsed.positionals;
+  if (parsed.values.help || !specifier || specifier === "help") {
     printUsage();
     return 0;
   }
 
-  if (action !== "install") {
-    console.error(`Unknown action: ${action}`);
-    printUsage();
-    return 1;
-  }
-
-  if (!specifier) {
-    console.error("Missing skill identifier or GitHub URL.");
+  if (parsed.positionals.length > 1) {
+    console.error(`Unexpected argument: ${parsed.positionals[1]}`);
     printUsage();
     return 1;
   }
@@ -591,4 +585,22 @@ export async function runSkillsSubcommand(argv: string[]): Promise<number> {
     console.error(error instanceof Error ? error.message : String(error));
     return 1;
   }
+}
+
+export async function runInstallSubcommand(argv: string[]): Promise<number> {
+  return runInstall(argv);
+}
+
+export async function runSkillsSubcommand(argv: string[]): Promise<number> {
+  const [action, ...rest] = argv;
+  if (action === "install") {
+    return runInstall(rest);
+  }
+  if (!action || action === "help" || action === "--help" || action === "-h") {
+    printUsage();
+    return 0;
+  }
+  console.error(`Unknown action: ${action}`);
+  printUsage();
+  return 1;
 }
--- a/src/cli/subcommands/skills.ts
+++ b/src/cli/subcommands/skills.ts
@@ -29,6 +29,8 @@ interface InstallResult {
   name: string;
   path: string;
   source: string;
+  committed?: boolean;
+  commitSha?: string;
 }
 
 interface SkillListItem {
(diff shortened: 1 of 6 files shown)


src/cli/subcommands/skills.ts
```diff
@@ -0,0 +1,594 @@
...
+          clawHubSource as ClawHubSourceLocation,
+        );
+    tmpDir = downloaded.tmpDir;
+    const sourceDir = resolve(downloaded.sourceDir);
+    assertInside(tmpDir, sourceDir);
+    if (!existsSync(sourceDir)) {
+      const missingPath = gitSource?.subdir || clawHubSource?.slug || ".";
+      throw new Error(`Skill path not found: ${missingPath}`);
+    }
+    const result = await installSkillDirectory({ sourceDir, memoryDir, force });
```
amelia-letta, line 646: Files are copied to disk but never committed/pushed to the memfs repo. `commitAndSyncMemoryWrite` exists in `memory-git.ts` — without it the skill is local-only and won't persist across machines or survive a memfs reset.

src/cli/subcommands/skills.ts
```diff
@@ -0,0 +1,784 @@
...
+  const skills = await listSkillDirectories({ memoryDir });
+  return { agentId, skills };
+}
+
+async function deleteSkill(
+  skillName: string,
+  agentId: string,
+): Promise<DeleteResult> {
+  const memoryDir = await getAgentMemoryDir(agentId);
+  const result = await deleteSkillDirectory({ memoryDir, name: skillName });
```
amelia-letta, line 730: Same as install — `rmSync` removes files from disk but never commits the deletion to the memfs repo. On next pull the deleted skill could reappear if the server still has it.

PR #2592, opened 2026-05-29
fix(auth): fix stale credential reauth flow

cpacker:
## Summary
- Separate OAuth device-code login from the "already signed in" check so startup reauth can replace stale stored credentials.
- Start failed cloud startup auth directly in device-code reauth, while explicitly blocking invalid `LETTA_API_KEY` env overrides and server reachability failures.
- Add loading/cancel states around device-code setup and regression coverage for the stale-credential path.

## Test plan
- [x] `bun test src/cli/local-first-setup-wiring.test.ts src/auth/oauth-network-errors.test.ts`
- [x] `bun run check`
- [x] Manual headless invalid env check: `LETTA_API_KEY=bad ... letta -p hello`

### DIFF
--- a/src/auth/ConstellationLoginView.tsx
+++ b/src/auth/ConstellationLoginView.tsx
@@ -84,10 +84,7 @@ export function ConstellationLoginView({
 
         const now = Date.now();
         settingsManager.updateSettings({
-          env: {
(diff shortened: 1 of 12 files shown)


ThatOneCalculator:
seems to have fixed the issue for me!!

PR #2599, opened 2026-05-29
feat(listener): add typed Letta Code upgrade command

kianjones9:
## Summary
- Add an allowlisted `upgrade-letta-code` remote command advertised through listener supported commands.
- Run the existing manual updater inside the listener, then restart `letta remote` with the same environment name after a successful update.
- Include supported command metadata at environment registration so clients can capability-gate upgrade affordances.

## Test plan
- [x] `git diff --check`
- [x] `npm run typecheck`


src/updater/auto-update.ts
```diff
@@ -107,6 +115,61 @@ export function buildInstallCommand(
...
+      writable: false,
+      reason: "Self-update is disabled for development/source checkouts.",
+      manual_command: manualCommand,
+    };
+  }
+
+  const packageParentPath = dirname(installPath);
+  const writable = canWritePath(installPath) && canWritePath(packageParentPath);
+  return {
+    supported: true,
```
kianjones9, line 183: upgrade specific metadata

carenthomas (approved):

PR #2600, opened 2026-05-29
feat(reflection): support manual multi-transcript reflection

kl2806:
## Summary
- add manual multi-transcript reflection with `/reflect --recent N` and repeated `/reflect --conversation <id>`
- add `/reflect --auto`, which builds transcript candidates, asks a selector reflection subagent to choose relevant conversations, then launches one multi-transcript reflection pass
- support focused reflection instructions via `--instruction` / `-i` / `--`, threading the user instruction through auto candidate selection, multi-reflection manifests, and final reflection prompts
- build multi-transcript manifests with unreflected/replay slices, selector reasons, auto-selected candidate metadata, and optional `user_instruction`
- advance cursors only for unreflected slices after successful multi-reflection
- teach reflection subagents to read multi-transcript manifests and auto transcript candidate manifests
- filter weak `/reflect --auto` candidates so one-turn leftovers and reflected recent-only transcripts do not dominate selection

## Test plan
- `bun test src/cli/reflection-transcript.test.ts`
- `bun run check`

### DIFF
--- a/src/cli/app/use-submit-handler.ts
+++ b/src/cli/app/use-submit-handler.ts
@@ -22,6 +22,7 @@ import {
   STALE_APPROVAL_RECOVERY_DENIAL_REASON,
 } from "@/agent/approval-recovery";
 import { getResumeDataFromBackend } from "@/agent/check-approval";
+import { ISOLATED_BLOCK_LABELS } from "@/agent/memory";
 import {
   ensureMemoryFilesystemDirs,
   getScopedMemoryFilesystemRoot,
@@ -31,7 +32,6 @@ import {
   isActiveMemfsEnabled,
   isLocalMemfsActive,
 } from "@/agent/memory-runtime";
-import { sendMessageStreamWithBackend } from "@/agent/message";
 import {
   detectPersonalityFromPersonaFile,
   type PersonalityId,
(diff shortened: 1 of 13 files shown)


src/cli/app/use-submit-handler.ts
```diff
@@ -2750,47 +2882,314 @@ export function useSubmitHandler(ctx: SubmitHandlerContext) {
...
+                conversationId: reflectionConversationId,
+              });
+              cmd.finish(
+                `Reviewing ${autoPayload.candidates.candidates.length} candidate transcript(s) for reflection. View the transcript candidates here: ${autoPayload.candidatesPath}`,
+                true,
+              );
+              return { submitted: true };
+            }
+
+            const reflectionPayload = await buildMultiReflectionPayload({
```
amelia-letta, line 3350: No active-subagent guard for `recent`/`conversations`/`auto` paths — the `single` kind delegates to `launchReflectionSubagent` which checks `isReflectionSubagentActive`, but the multi-transcript paths skip straight to `buildMultiReflectionPayload` + `spawnBackgroundSubagentTask`. Running `/reflect --recent 5` while a single `/reflect` is already active will spawn a second reflection subagent that races on memory writes.

just-cameron:
## Silent transcript dropping when `maxTotalChars` budget is exceeded

When `buildMultiReflectionPayload` iterates through conversations, transcripts that would push the total past `maxTotalChars` (150K) are silently skipped — the loop just `continue`s. A user running `/reflect --recent 10` might get 3 conversations reflected with no indication that 7 were dropped.

The manifest does track `approx_chars` per slice, so the info is available — it's just not surfaced to the user or the subagent.

A few options to consider:

1. **Warn in the terminal** — after launch, print something like `⚠ Reflecting on 3 of 10 requested conversations (150K char budget reached). Skipped: conv-abc, conv-def, ...`
2. **Include skipped conversations in the manifest** — so the reflection subagent can acknowledge what it didn't get
3. **Progressive summarization** — if a transcript won't fit raw, summarize it first (via a quick subagent or truncation), then include the summary. Every requested conversation contributes, just at different fidelity levels

Option 3 is the most principled (mirrors how memory consolidation actually works) but also the most complex. At minimum, option 1 would prevent silent data loss in a memory system.

Two minor notes as well:
- The first transcript always bypasses the budget check (`transcripts.length > 0` guard), so a single 200K-char conversation would blow the budget for everything else
- `reservedReflectionAgentIds` is an in-memory `Set` — fine for CLI, but in server mode a process restart loses the reservation

cpacker (approved):

PR #2625, opened 2026-05-30
fix(cli): declutter slash command autocomplete

kianjones9:
## Summary
- Make slash autocomplete height viewport-aware
- Hide bundled skill commands only when a builtin command already covers the same workflow
- Keep non-aliased bundled and agent skills visible

## Test plan
- [x] bunx --bun @biomejs/biome@2.2.5 check --write src/cli/components/SlashCommandAutocomplete.tsx
- [x] bun run typecheck


carenthomas:
so just to confirm, the new visible count will be in the range of 3-12 based on the terminal height?

carenthomas (approved):

kianjones9:
yes. In other apps they actually take like half the terminal I think.

PR #2631, opened 2026-05-30
fix(headless): handle bidirectional interrupts

codeluggage:
## Summary

Fixes interrupt handling in bidirectional headless `stream-json` mode:

- Handles `control_request: interrupt` synchronously in the readline handler so long-running turns can be cancelled while the main loop is blocked draining the model/tool stream.
- Latches interrupts that arrive in the narrow window after a user message is handed to the main loop but before the turn AbortController exists.
- Avoids latching idle interrupts, so an interrupt sent between turns does not poison the next user message.
- Wakes pending `getNextLine()` waiters after fast-path interrupt aborts an active turn, so interrupts during `can_use_tool` permission waits do not hang.
- Clears dangling pending approvals before the next turn if the prior turn ended interrupted.

## Verification

- `bun test src/headless-interrupt-latch.test.ts src/headless-interrupt-recovery.test.ts`
- `bunx --bun @biomejs/biome@2.2.5 check src/headless.ts src/headless-interrupt-latch.test.ts`
- `bun run typecheck`
- Live smoke: headless bidirectional `--permission-mode standard`, Bash tool triggers `can_use_tool`; sending `control_request: interrupt` during the permission wait returns `result.subtype = "interrupted"` and does not execute the tool.

### DIFF
--- a/src/headless-interrupt-latch.test.ts
+++ b/src/headless-interrupt-latch.test.ts
@@ -143,5 +143,10 @@ describe("headless interrupt latch wiring", () => {
     expect(source).toContain(
       "// Controller now exists — close the pre-controller race window.",
     );
+    // Interrupts must wake permission/external-tool waits that are blocked in
+    // getNextLine(), otherwise the fast path consumes the interrupt and hangs.
+    expect(source).toContain(
+      "Wake that waiter so the aborted turn can unwind.",
+    );
   });
 });
--- a/src/headless-interrupt-latch.test.ts
+++ b/src/headless-interrupt-latch.test.ts
@@ -1,0 +1,147 @@
+import { describe, expect, test } from "bun:test";
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { decideInterruptAction } from "@/headless";
+
+describe("decideInterruptAction", () => {
+  test("aborts the active turn when a controller exists", () => {
+    expect(
+      decideInterruptAction({ hasActiveController: true, turnStarting: false }),
+    ).toBe("abort-active");
+    // An active controller takes precedence even mid-startup.
+    expect(
+      decideInterruptAction({ hasActiveController: true, turnStarting: true }),
+    ).toBe("abort-active");
+  });
+
+  test("latches when a turn is starting but its controller does not exist yet", () => {
+    // Narrow pre-controller race: a user message was just dispatched.
+    expect(
+      decideInterruptAction({ hasActiveController: false, turnStarting: true }),
+    ).toBe("latch");
+  });
+
+  test("is a no-op when idle (no active or starting turn)", () => {
+    // The regression: an idle interrupt must NOT latch, or it poisons the
+    // next user turn.
+    expect(
+      decideInterruptAction({
+        hasActiveController: false,
+        turnStarting: false,
+      }),
+    ).toBe("noop");
+  });
+});
+
+/**
+ * Lifecycle simulation mirroring the headless main loop's interrupt handling,
+ * driven purely by decideInterruptAction. Proves the two behaviors cpacker
+ * asked us to cover in PR #2631:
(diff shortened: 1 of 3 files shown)


cpacker:
Thanks for the detailed writeup — the tool-return emission piece looks directionally good, and I verified a merge with current `origin/main` passes `bun run check` locally.

I found one blocking issue in the interrupt latch behavior:

```ts
if (currentAbortController !== null) {
  currentAbortController.abort();
} else {
  pendingInterrupt = true;
}
```

This catches the intended race window, but it also catches idle interrupts. If the headless session is idle and waiting in `getNextLine()`, an incoming `control_request: interrupt` will be consumed by the fast path, return success, and set `pendingInterrupt = true`. The next normal user prompt then creates an AbortController and immediately aborts itself:

```ts
currentAbortController = new AbortController();
if (pendingInterrupt) {
  pendingInterrupt = false;
  currentAbortController.abort();
}
```

So an interrupt sent while idle can poison the next turn. Existing behavior was effectively no-op success when there was no current operation, which seems correct.

I think the fix is to distinguish “idle, no active turn” from the narrow “user message was just dequeued and controller is about to be created” window. For example, add a short-lived `turnStarting` / `creatingTurnController` flag around the user-message setup and only latch in that specific pre-controller window. Otherwise, if no controller exists, respond success but do not set `pendingInterrupt`.

Could you also add coverage for both cases?

- idle interrupt does not abort the next user message
- interrupt immediately after a user message still aborts that turn via the latch

Non-blocking: the tool-return forwarding callback is duplicated in two call sites; a helper would be cleaner but is not required for this fix.

codeluggage:
@cpacker Thank you for this - and the patience!

Useful pointers and advice. The issue you brought up has been fixed and I've been running it myself for a while. Covered with tests but hard for me to know how complete they are as I am not that deep in this codebase.

Brought it up to date with `main`. 

Extracting duplicated tool-return callback into a helper sounds right - prefer that here in this PR or separately? I expect to be back with more PRs as I continue using this and expanding on it.

A few more specifics from Claude Fable:

Fixed in e2bd9e1 using the approach you suggested: a turnStarting flag scoped to the pre-controller window. Interrupt handling now distinguishes three states (see decideInterruptAction): active controller → abort; turnStarting with no controller yet → latch; idle → no-op success without latching. While testing this I also found that an interrupt landing mid-approval could leave a dangling approval that blocked the next turn — fixed in 7e1057e.

Test coverage added in headless-interrupt-latch.test.ts for both cases you asked for (idle interrupt does not abort the next user message; interrupt immediately after a user message still aborts via the latch), plus an idle→turn→idle sequence test and a source-level check that the latch window closes after controller creation.

devanshrj:
Hey @codeluggage — quick update on this PR. First, thanks again for driving this and for explaining the real agent-shell/headless use case behind it.

I updated the branch to get it back into a focused/mergeable shape. The original PR had two concerns: local tool-return stream output and interrupt/cancel handling. We split those now:

- The local tool event stream work moved to #3242. That PR now emits paired local `tool_call_message` → `tool_return_message` events, including denied tools as error returns, which should cover the tool-output needs from this PR.
- This PR now keeps the interrupt/cancel correctness pieces from your work:
  - fast-path `control_request: interrupt`
  - pre-controller interrupt latch
  - idle interrupt no-op behavior
  - dangling approval recovery after interrupt

I also added one follow-up fix: if an interrupt arrives while headless is waiting for a `can_use_tool` permission response, we now wake the pending `getNextLine()` waiter so the turn returns `interrupted` instead of hanging.

I updated the PR title/body to reflect the narrower scope, but wanted to call out the split explicitly since the original description mentioned your tool/integration use case. If #3242’s stream shape doesn’t cover what your integration needed from the original tool-return changes, please let us know; we can cover any remaining integration gaps in follow-up PRs.

Verified with focused tests + live smoke for interrupt-during-permission-wait.

PR #2634, opened 2026-05-31
fix(cli): restore on-demand file autocomplete

cpacker:
## Summary
- Replace the custom PR2 implementation with a close port of local Pi TUI `packages/tui/src/autocomplete.ts` @ file autocomplete behavior.
- Use Pi-style fd-backed fuzzy file suggestions, scoped path queries, quoting for paths with spaces, directory trailing-slash completion, and file trailing-space completion.
- Keep Letta-specific changes limited to Ink rendering/input wiring and fd path resolution.

## Test plan
- [x] `bun run check`
- [x] `bun test src/cli/pi-file-autocomplete.test.ts src/tools/create-worktree.test.ts`

Stacked on #2633.


src/cli/components/InputRich.tsx
```diff
@@ -1706,6 +1706,44 @@ export function Input({
     onSubmit,
   ]);
 
+  // Handle file selection from autocomplete
+  const insertFileReference = useCallback(
+    (selectedPath: string, appendSpace: boolean) => {
+      const atIndex = value.lastIndexOf("@");
```
amelia-letta: `lastIndexOf("@")` replaces the wrong `@` when there are multiple in the input. `extractSearchQuery` in FileAutocomplete correctly identifies the active `@` by cursor position, but `atIndex` is lost at the callback boundary — `onSelect` only passes the path. If input is `@first @sec` and the cursor is near `@first`, this replaces `@sec` instead.

PR #2641, opened 2026-05-31
refactor(perf): avoid blocking local startup on model discovery

cpacker:
## Summary
- Keep local disk-backed resume off the model discovery critical path by backgrounding local `backend.listModels()` startup hints.
- Avoid redundant startup settings writes by skipping unchanged session persistence and removing the extra early `lastAgent` write.
- Overlap terminal theme and Kitty preflight probes with React/Ink/App imports while still completing them before Ink render.

## Test plan
- [x] `bun test src/settings-manager.test.ts`
- [x] `bun run typecheck`
- [x] `bun run check`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -691,9 +691,8 @@ async function main(): Promise<void> {
     process.exit(subcommandResult);
   }
 
-  // Everything below only runs for interactive TUI mode
+  // Everything below only runs for interactive/headless agent mode
   await settingsManager.initialize();
-  await initTerminalTheme();
 
   const settings = await settingsManager.getSettingsWithSecureTokens();
   markMilestone("SETTINGS_LOADED");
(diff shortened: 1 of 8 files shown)


src/index.ts
```diff
@@ -1266,61 +1285,82 @@ async function main(): Promise<void> {
...
-      if (isValid) {
-        const { bootstrapBaseToolsIfNeeded } = await import(
-          "@/agent/bootstrap-tools"
-        );
-        await bootstrapBaseToolsIfNeeded();
-      }
+        return credentialValidation;
+      };
+
+      if (!isHeadless) {
```
amelia-letta: Deferring credential validation drops the old interactive-mode error handling — specifically the `LETTA_API_KEY` set+invalid exit and the network-error setup redirect. With this change, the TUI renders with bad credentials and the user gets a generic API error instead of the targeted "unset LETTA_API_KEY" message or setup flow. Intentional?

cpacker: Good catch — not intentional. I restored the pre-TUI interactive credential validation path so invalid `LETTA_API_KEY` and network/setup diagnostics are preserved before API startup. The perf part that remains is that terminal preflight starts earlier and base-tool bootstrap is backgrounded after successful validation. Pushed in a8697f75.

PR #2654, opened 2026-06-01
fix(bash): add opt-in strict shell mode

kl2806:
## Summary
- Add `LETTA_BASH_STRICT=1` to opt into running Unix shell commands with `set -euo pipefail`.
- Apply the strict prelude through the shared shell launcher path so `Bash`, `shell_command`, and `run_shell_command` all get the same behavior while default interactive behavior remains unchanged.
- Add regression/parity tests for the rollout bug where an intermediate shell failure can be masked by a later successful command.

## Context
While debugging Tinker skill-reflection rollouts, we found trajectories where the agent attempted to write a skill file to a missing directory, then printed a success message. Because the shell command ended with a successful command, Letta Code returned a successful tool result even though the actual write failed. Minimal example:

```bash
cat missing-file
echo "SKILL.md written successfully"
```

Without strict mode, the final `echo` can make the overall shell exit code `0`, so shell tools report success even though an earlier required operation failed. In the rollout, the observed feedback looked like:

```text
SKILL.md written successfully
bash: .../SKILL.md: No such file or directory
```

but the tool status was still `success`. That can corrupt eval/training signal because the policy sees a successful tool call for a failed file write. Harnesses can now set `LETTA_BASH_STRICT=1` so these intermediate shell failures propagate through the tool output.

## Test plan
- [x] `bun run lint`
- [x] `bun run typecheck`
- [x] `bun test src/tools/bash.test.ts --test-name-pattern "strict mode"`
- [x] `bun test src/tools/shell-command.test.ts --test-name-pattern "strict mode"`
- [x] `bun test src/tools/run-shell-command.test.ts --test-name-pattern "strict mode"`
- [x] `bun test src/tools/shell-launchers.test.ts --test-name-pattern "strict shell"`

### DIFF
--- a/src/tools/impl/bash.ts
+++ b/src/tools/impl/bash.ts
@@ -21,6 +21,19 @@ import { validateRequiredParams } from "./validation.js";
 // Cache the working shell launcher after first successful spawn
 let cachedWorkingLauncher: string[] | null = null;
 
+const STRICT_BASH_ENV_VAR = "LETTA_BASH_STRICT";
+const STRICT_BASH_PRELUDE = "set -euo pipefail";
+
+function isTruthyEnvValue(value: string | undefined): boolean {
+  return value === "1" || value?.toLowerCase() === "true";
+}
+
+function withStrictBashPrelude(command: string, env: NodeJS.ProcessEnv): string {
+  if (process.platform === "win32") return command;
+  if (!isTruthyEnvValue(env[STRICT_BASH_ENV_VAR])) return command;
+  return `${STRICT_BASH_PRELUDE}\n${command}`;
+}
+
 function rebuildCachedLauncher(
   command: string,
   secretEnv?: Record<string, string>,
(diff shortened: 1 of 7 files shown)


src/tools/impl/bash.ts
```diff
@@ -21,6 +21,22 @@ import { validateRequiredParams } from "./validation.js";
...
+  return value === "1" || value?.toLowerCase() === "true";
+}
+
+function withStrictBashPrelude(
+  command: string,
+  env: NodeJS.ProcessEnv,
+): string {
+  if (process.platform === "win32") return command;
+  if (!isTruthyEnvValue(env[STRICT_BASH_ENV_VAR])) return command;
+  return `${STRICT_BASH_PRELUDE}\n${command}`;
```
amelia-letta: Shell tool parity gap — `shell_command` and `run_shell_command` also execute multi-line shell scripts but don't get this strict mode protection. Same intermediate-failure-masking bug could happen there. Consider applying the prelude in `shell-launchers.ts` or extracting `withStrictBashPrelude` to a shared helper.

kl2806:
Addressed the shell tool parity review comment in d32ca066.

Changes:
- Moved `LETTA_BASH_STRICT` handling into `shell-launchers.ts` via shared `withStrictShellPrelude`.
- Kept `Bash` using the shared helper.
- Wired `shell_command` to pass its effective env into `buildShellLaunchers` so `secretEnv`/harness env also work.
- Added targeted parity tests for `shell_command`, `run_shell_command`, and `buildShellLaunchers`, in addition to the existing Bash strict-mode test.

devanshrj (approved):

PR #2670, opened 2026-06-02
feat(cli): add Codex-style terminal title activity

cpacker:
## Summary

- Adds Codex-style terminal title activity with a Letta-specific default of `activity` + `agent-name` so terminal tabs reflect the long-lived agent you are talking to.
- Keeps `project-name` and the other Codex-style title items selectable/reorderable through `/title`, including aliases and ordered rendering.
- Adds a managed OSC title writer with sanitization, duplicate suppression, BEL termination, and clear-on-unmount behavior.
- Drives title activity from TUI runtime state, including the 100ms braille spinner while working and 1s Action Required blink while approval is pending.
- Validated with focused title tests and `bun run check`.

### DIFF
--- a/src/cli/app/AppCoordinator.tsx
+++ b/src/cli/app/AppCoordinator.tsx
@@ -55,6 +55,7 @@ import {
 } from "@/cli/commands/runner";
 import type { BtwState } from "@/cli/components/BtwPane";
 import type { ModelSelectorSelection } from "@/cli/components/ModelSelector";
+import { useFrameCycle } from "@/cli/components/spinners/use-frame-cycle";
 import { buildStatuslineRenderContext } from "@/cli/display/statusline/context";
 import type { ExtensionConversationCloseReason } from "@/cli/extensions/types";
 import {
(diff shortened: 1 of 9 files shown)


src/cli/app/AppCoordinator.tsx
```diff
@@ -4634,6 +4641,145 @@ export function App({
...
+    !terminalTitleShowsActionRequired && terminalTitleTaskRunning;
+  const terminalTitleRunState =
+    loadingState !== "ready"
+      ? "Starting"
+      : !terminalTitleTaskRunning
+        ? "Ready"
+        : executionPhase === "thinking"
+          ? "Thinking"
+          : "Working";
+  const terminalTitleSpinnerFrameIndex = useFrameCycle(
```
amelia-letta: `useFrameCycle` here is App-level `useState` — every 100ms tick re-renders the entire App + AppView tree while the spinner is active. The existing `StreamingStatusSpinner` avoids this by being a `memo` component. Consider extracting a `TerminalTitleWriter` component that owns the frame cycle + title computation + stdout write, so the 10Hz re-render stays local instead of hitting the whole tree.

PR #2678, opened 2026-06-02
fix(telemetry): populated subagent_id on start + fix flush

christinatong01:
## Summary

Two fixes for missing reflection telemetry in PostHog.

### 1. `reflection_start` now carries `subagent_id`

The old code awaited the agent-creation ID lookup inline with a 1s timeout before emitting `reflection_start`. Agent creation takes 1–10s, so it timed out ~100% of the time, leaving `subagent_id` empty. Now the wait runs in the background (up to 30s) and emits when the ID resolves; the launcher returns immediately. Timeout → emit with `undefined` instead of dropping the event.

### 2. Telemetry flush no longer drops the final batch on shutdown

Two flushes racing ~22ms apart on exit hit 429, dropping the second batch (typically late `reflection_end` events). Fixes:

- `inflightFlush` guard — concurrent callers share one POST
- `flush()` split into guard + `performFlush()`
- New `drain()` (3s cap) — exit handlers `await drain()` instead of fire-and-forget
- Flush interval: 5min → 2min, batch size: 100 → 50

## Test plan

- [x] `bun test src/telemetry/` — 6/6 pass
- [x] `bun run typecheck` — clean
- [x] `biome check` on changed files — clean

## Notes

- Ctrl+C within ~1–3s of reflection launch may lose `reflection_start` — accepted, reflection itself also wouldn't have run
- No new event types, no schema-breaking changes

### DIFF
--- a/src/cli/helpers/reflection-launcher.ts
+++ b/src/cli/helpers/reflection-launcher.ts
@@ -15,6 +15,27 @@ import { debugLog, debugWarn } from "@/utils/debug";
 
 export const AUTO_REFLECTION_DESCRIPTION = "Reflect on recent conversations";
 
+/**
+ * Maximum time to wait (in the background, non-blocking) for the spawned
+ * reflection subagent to be assigned a concrete agent ID before we emit the
+ * `reflection_start` telemetry event with whatever we have.
+ *
+ * Background: `spawnBackgroundSubagentTask` returns immediately with a local
+ * task id, but the actual Letta agent is created asynchronously via an HTTP
+ * POST that typically takes 1–10 seconds (longer under congestion or cold
+ * start). Previously this wait was capped at 1s AND awaited inline, which
+ * (a) blocked the reflection-launch path for up to 1s, and (b) almost always
+ * timed out, causing `reflection_start.subagent_id` to be empty in 100% of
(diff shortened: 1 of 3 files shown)


src/cli/helpers/reflection-launcher.ts
```diff
@@ -143,6 +146,17 @@ export async function launchReflectionSubagent(
 
     const { spawnBackgroundSubagentTask, waitForBackgroundSubagentAgentId } =
       await import("@/tools/impl/task");
+
+    // Defer `reflection_start` until the agent ID resolves (background, bounded by REFLECTION_AGENT_ID_WAIT_MS).
+    const emitReflectionStart = (resolvedAgentId: string | null) => {
+      telemetry.trackReflectionStart(triggerSource, {
+        subagentId: resolvedAgentId ?? undefined,
```
sarahwooders, line 153: is this the `agent-...` id, or the local subagent id?

sarahwooders (approved):

src/cli/helpers/reflection-launcher.ts
christinatong01: this is the letta agent id (not local runtime subagent id)

PR #2679, opened 2026-06-03
refactor(prompts): reuse memfs prompts for local backend

devanshrj:
## Summary
- remove non-recall local backend MemFS prompt forks
- make local backend system prompts reuse the shared MemFS prompt
- make init/memory/reflection built-ins backend-neutral with `$MEMORY_DIR` and commit-only wording
- keep the local recall prompt separate because local transcript search has distinct behavior

## Tests
- `bun test src/agent/prompt-assets.test.ts src/agent/subagent-builtins.test.ts src/backend/local-system-prompt-compilation.test.ts`
- `bun run check`

## Stack
- Stacked on #2677 / `letta/memfs-post-turn-push-319c3461`


src/agent/subagents/builtin/memory.md
```diff
@@ -219,24 +214,7 @@ index), wait and retry up to 3 times with backoff (sleep 2,
...
-git pull --rebase
-```
-
-If rebase has conflicts, resolve them autonomously to
-stabilize local `main` against remote `main` first. In this
-step, prefer **remote main** content for conflicting files,
-then run `git rebase --continue`.
-
-Now merge the defrag branch:
+Merge the defrag branch:
```
amelia-letta, line 217: Dropping `git pull` before merge means remote MemFS agents lose the pull-before-merge safety step. If the harness pushed changes after a turn and the defrag subagent merges without pulling, the local main can be behind remote — subsequent harness push fails or creates divergence. Is the stacked PR (#2677) handling this differently, or should the pull step stay for remote MemFS agents?

devanshrj: No but we'll figure out a better pull mechanism too like we did for push in #2677

PR #2680, opened 2026-06-03
feat: pass response state ids through letta-code streams

jnjpng:
## Goal
Implement the letta-code side of approval-boundary response state reuse. The client should remember response-state ids emitted by tmp cloud and send the previous id back only on the next eligible approval continuation, so cloud can later reuse prepared approval-boundary context instead of rebuilding it from the DB.

This is intentionally scoped to the hot path we can reason about safely: pure approval continuations that were fully auto-handled by the client. That includes auto-allowed tool execution and auto-denied permission results, regardless of permission mode. If a human reviewed any approval, the client does not send the response-state header because the user-visible pause creates more opportunity for agent/conversation state to change.

## What changed
- Capture `response_state` SSE metadata chunks and store the latest `response_id` by conversation/agent scope.
- Suppress `response_state` metadata chunks from normal CLI output.
- Send `X-Letta-Response-State` only when there is an actual `previous_response_id` and the approval continuation was explicitly marked as fully auto-handled.
- Keep the header minimal: `v`, `cache_scope`, and `previous_response_id`; no client-local tool context id is sent.
- Clear stale response-state ids when leaving the eligible path so ids are not reused across non-approval or manually reviewed flows.

## Notes
- This PR is safe to publish before the cloud PR: older servers ignore the custom header, and if cloud does not emit `response_state`, the client sends nothing.
- Companion cloud PR: letta-ai/letta-cloud#12046.

## Tests
- `bun test src/agent/send-message-stream-acting-user.test.ts`
- `bun test src/cli/approval-classification.test.ts`
- `bun test src/websocket/listen-client-concurrency.test.ts --test-name-pattern \"approval continuation\"`
- `bun run typecheck`
- `bun run check:boundaries`
- commit hook: biome/lint-staged + typecheck

### DIFF
--- a/src/agent/message.ts
+++ b/src/agent/message.ts
@@ -11,6 +11,7 @@ import type {
 } from "@letta-ai/letta-client/resources/agents/messages";
 import type { MessageCreateParams as ConversationMessageCreateParams } from "@letta-ai/letta-client/resources/conversations/messages";
 import { type Backend, getBackend } from "@/backend";
+import { permissionMode } from "@/permissions/mode";
 import {
   type ClientTool,
   type PermissionModeState,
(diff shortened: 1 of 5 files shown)


carenthomas (approved):

kianjones9 (approved):
awesome!

PR #2711, opened 2026-06-04
fix(tools): guide Windows Bash usage

sarahwooders:
## Summary
- Clarifies Bash tool guidance/schema so Windows agents know the tool executes through PowerShell, not Unix bash.
- Adds Windows-specific prompt guidance for common failures seen in D. Attridge payloads: `&&` / `||`, Unix utilities like `head`/`grep`, curl-style headers, and fragile one-line PowerShell quoting.
- Appends a recovery hint to Bash error output when common Windows shell syntax failures occur so the next model turn can adapt instead of repeating the same broken command style.

## Test plan
- [x] `bun test src/tools/bash.test.ts src/tools/codex-unified-exec-toolset.test.ts`
- [x] `bunx --bun @biomejs/biome@2.2.5 check src/tools/bash.test.ts src/tools/impl/bash.ts src/tools/tool-definitions.ts src/tools/schemas/Bash.json src/tools/descriptions/Bash.md`
- [ ] `bun run check` (fails in TypeScript on pre-existing missing `@pierre/diffs` / `@pierre/diffs/ssr` declarations in `src/web/generate-diff-viewer.ts`, unrelated to this PR)

Context: investigation of downloaded D. Attridge ClickHouse payloads showed GLM-5.1 tool calls were structurally valid, but many tool results failed at runtime because agents used Unix/bash idioms in Windows PowerShell sessions.

### DIFF
--- a/src/tools/impl/bash.ts
+++ b/src/tools/impl/bash.ts
@@ -25,6 +25,38 @@ import { validateRequiredParams } from "./validation.js";
 // Cache the working shell launcher after first successful spawn
 let cachedWorkingLauncher: string[] | null = null;
 
+const WINDOWS_SHELL_FAILURE_HINT = [
+  "",
+  "Windows shell note: Bash runs through PowerShell on Windows, not Unix bash.",
+  "Use PowerShell-compatible syntax or separate tool calls. Common fixes: use `;` plus explicit `$LASTEXITCODE` checks instead of `&&` / `||`, `Get-Content -TotalCount` instead of `head`, `Select-String` instead of `grep`, PowerShell hashtables for HTTP headers, and a temporary `.ps1` script for complex quoting.",
+].join("\n");
+
+const WINDOWS_SHELL_ERROR_PATTERNS = [
+  /The token '&&' is not a valid statement separator/i,
+  /The token '\|\|' is not a valid statement separator/i,
+  /\b(head|grep|sed|awk|cat) : The term '\1' is not recognized/i,
+  /The string is missing the terminator/i,
+  /Cannot bind parameter 'Headers'/i,
+  /Missing '\(' after 'if' in if statement/i,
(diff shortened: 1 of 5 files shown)


src/tools/impl/bash.ts
```diff
@@ -25,6 +25,38 @@ import { validateRequiredParams } from "./validation.js";
...
+  /Cannot bind parameter 'Headers'/i,
+  /Missing '\(' after 'if' in if statement/i,
+  /& was unexpected at this time/i,
+  /The '<' operator is reserved for future use/i,
+  /Missing property name after reference operator/i,
+  /Missing argument in parameter list/i,
+  /Unexpected token/i,
+];
+
+export function appendWindowsShellFailureHint(
```
amelia-letta, line 48: `exec_command` also runs through PowerShell on Windows (line 360 of exec-command.ts) and can produce the same error patterns in its output. Shell parity gap — same class as PR #1711 (Bash had truncation, shell_command didn't). Worth applying `appendWindowsShellFailureHint` to `formatExecOutput` too.

cpacker:
Closing this as superseded by #2712.

#2712 fixed the core issue we wanted to address: the Claude/default `Bash` tool now has platform-aware Windows guidance explaining that the runtime uses the same Codex-derived native Windows launcher strategy: PowerShell Core first, Windows PowerShell fallback, then `cmd.exe`.

This PR also adds runtime recovery hints for PowerShell syntax failures, which may still be worth doing, but not in this shape:
- it conflicts with #2712,
- it hardcodes Windows guidance into the static `Bash.json` schema,
- and it only applies the recovery hint to `Bash`, while `exec_command` uses the same Windows launcher path and can hit the same error patterns.

If we want adaptive Windows shell failure hints, we should do that as a separate follow-up PR that applies consistently across `Bash` and `exec_command` without changing the static schema.

Issue #2713, opened 2026-06-04
Stuck when creating agent using CLI with local backend


KY64:
When I try to run `letta --backend local agents create` I got stuck forever. No error, no warning. This happens after I have signed in to Letta from the CLI. However, if I haven't signed in, it doesn't get stuck and instead give me this error

```
Missing LETTA_API_KEY
Run 'letta' to configure authentication, or set LETTA_API_KEY to your API key
Error: getClient() called without credentials
```

but the local agent is successfully created.

### Version

`0.27.0` and I even cloned the repo then build it, it's still happening.

### Repro steps

1. Install CLI
2. Run `letta` then follow the authentication process
3. Run `letta --backend local agents create`

### Expected behavior

It should just show a JSON output of the agent and continue the shell so user can run other command.

### Actual behavior

It is stuck forever. No warning, no error.

### Known workarounds

Just press CTRL+C or terminate the process after few seconds, you can run `letta` again and the agent is already created.

### Related information

* Operating system: GNU/Linux

### Do you have time to contribute to fix?

Yes, I have forked the repo and pushed a fix [here](https://github.com/KY64/letta-code/tree/fix/hang-local-agent-creation). Will create a PR if this issue hasn't been solved or anyone is not working on it.

PR #2723, opened 2026-06-04
fix: hang local agent creation

KY64:
### Summary

This PR will solve issue https://github.com/letta-ai/letta-code/issues/2713 by ignoring memFS sync to Letta Cloud API when the backend mode is set to `local`.

### Changes

- Add check of `getBackend().capabilities.remoteMemfs` in `enableMemfsForCreatedAgent` before executing memFS cloud API sync
- Move `enableMemfsForCreatedAgent` from `src/agent/personality.ts` to `src/agent/create.ts` to prevent circular dependency
- Removing check of `personality` in `src/cli/subcommands/agents.ts` which is used to determine whether to call `enableMemfsForCreatedAgent`
- Add test on `src/agent/personality.test.ts` to ensure the cloud sync is skipped when backend mode is `local`

### DIFF
--- a/src/agent/create.ts
+++ b/src/agent/create.ts
@@ -13,6 +13,7 @@ import { DEFAULT_AGENT_NAME, DEFAULT_SUMMARIZATION_MODEL } from "@/constants";
 import { settingsManager } from "@/settings-manager";
 import { getModelContextWindow } from "./available-models";
 import { getDefaultMemoryBlocks } from "./memory";
+import { GIT_MEMORY_ENABLED_TAG } from "./memory-git";
 import {
   formatAvailableModels,
   getDefaultModel,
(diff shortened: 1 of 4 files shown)


cpacker:
Thanks for opening this and tracking down the right failure mode. I opened a clean replacement PR at #2744 because this branch had gone stale against `main` and the implementation had a couple of issues after recent changes:

- it no longer merged cleanly (`src/agent/personality.ts` conflicts with newer tag-preservation logic on `main`)
- moving `enableMemfsForCreatedAgent` into generic `createAgent()` made MemFS enablement happen for every agent creation path, including callers that intentionally compute/skip MemFS mode themselves
- the older tag update path could overwrite existing agent tags instead of preserving the newer `origin:letta-code` tag handling

#2744 keeps your core fix — skip remote MemFS API calls when the active backend does not support remote MemFS — but applies it narrowly inside the existing MemFS enable helper and adds regression coverage for local backend creation. I included you as a co-author on the replacement commit.

cpacker:
Closing in favor of #2744, which keeps the same fix direction but rebases it cleanly onto current `main` and avoids broadening MemFS enablement to all `createAgent()` callers.

KY64:
Hi @cpacker thanks for looking into this!
I've checked the PR and it's much cleaner.
I did intentionally edited suggestion from Letta which was only changing `enableMemfsForCreatedAgent`, I thought it would be cleaner to just move to `createAgent` since I find this

```ts
    if (!personality) {
      await enableMemfsForCreatedAgent({
        agentId,
        agentTags: result.agent.tags,
      });
    }
```

can be simplified due to `createAgentForPersonality` also call it inside its function. Yet I missed the part where there is a case where memFS is disabled by the user and my code introduce another issue with it. I guess will need to ask Letta to review my change first before pushing. Thanks for merging the fix!

PR #2728, opened 2026-06-04
fix(agent): preserve Letta Code origin tags

sarahwooders:
## Summary
- keep `origin:letta-code` when desktop/personality-created agents enable git memory
- ensure resumed TUI/headless agents are tagged before managed system prompt version checks
- ensure websocket/listener turns tag agents as Letta Code agents and run managed prompt updates

## Testing
- `bun test src/agent/system-prompt-versioning.test.ts src/agent/personality.test.ts`
- `bunx biome check src/agent/system-prompt-versioning.ts src/agent/personality.ts src/agent/create.ts src/index.ts src/headless.ts src/websocket/listener/turn.ts`

Note: `bun run typecheck` currently fails on unrelated existing missing `@pierre/diffs` types in `src/web/generate-diff-viewer.ts`.

### DIFF
--- a/src/agent/personality.ts
+++ b/src/agent/personality.ts
@@ -483,9 +483,12 @@ export async function enableMemfsForCreatedAgent(params: {
       }
     }
     const tags = Array.from(new Set([...currentTags, LETTA_CODE_ORIGIN_TAG]));
-    if (!tags.includes(GIT_MEMORY_ENABLED_TAG)) {
+    if (
+      !tags.includes(GIT_MEMORY_ENABLED_TAG) ||
+      !currentTags.includes(LETTA_CODE_ORIGIN_TAG)
+    ) {
       await client.agents.update(agentId, {
-        tags: [...tags, GIT_MEMORY_ENABLED_TAG],
+        tags: Array.from(new Set([...tags, GIT_MEMORY_ENABLED_TAG])),
       });
     }
     settingsManager.setMemfsEnabled(agentId, true);
--- a/src/agent/personality.ts
+++ b/src/agent/personality.ts
@@ -2,8 +2,8 @@ import { execFile as execFileCb } from "node:child_process";
 import { existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { dirname, join } from "node:path";
 import { promisify } from "node:util";
-import { getBackend } from "@/backend";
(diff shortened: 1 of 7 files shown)


src/agent/personality.ts
```diff
@@ -470,7 +471,18 @@ export async function enableMemfsForCreatedAgent(params: {
...
+      try {
+        const agent = await client.agents.retrieve(agentId, {
+          include: ["agent.tags"],
+        });
+        currentTags = agent.tags ?? [];
+      } catch {
+        currentTags = [];
+      }
+    }
+    const tags = Array.from(new Set([...currentTags, LETTA_CODE_ORIGIN_TAG]));
```
amelia-letta, line 485: if the agent already has `git-memory-enabled` but not `origin:letta-code`, the `if` block is skipped and the origin tag is never persisted. need to also update when `LETTA_CODE_ORIGIN_TAG` is missing, even if `GIT_MEMORY_ENABLED_TAG` is already there.

src/headless.ts
```diff
@@ -1499,8 +1499,21 @@ export async function handleHeadlessCommand(
...
   if (isResumingAgent && !systemPromptPreset) {
-    const { getMemoryPromptModeForAgent, scheduleManagedSystemPromptUpdate } =
-      await import("@/agent/system-prompt-versioning");
+    const {
+      ensureLettaCodeOriginTag,
+      getMemoryPromptModeForAgent,
+      scheduleManagedSystemPromptUpdate,
+    } = await import("@/agent/system-prompt-versioning");
+    try {
+      agent = await ensureLettaCodeOriginTag(agent);
```
carenthomas: This tags any resumed headless agent as `origin:letta-code`, then the same tagged `agent` is passed into `clearPersistedClientToolRules` below. If that tag means “created/owned by Letta Code” rather than “touched by Letta Code,” this can make non-Code agents eligible for tool-rule cleanup.

src/agent/personality.ts
sarahwooders: Fixed in 23bd79875 — the update now also runs when `origin:letta-code` is missing, even if `git-memory-enabled` is already present, and writes a deduped tag set.

src/headless.ts
sarahwooders: Fixed in 23bd79875 — headless now keeps the original `agent` object for downstream startup cleanup and uses a separate `taggedAgent` only for managed prompt versioning.

PR #2741, opened 2026-06-05
Force continuous rendering mode

overlord-letta:
## Summary
- Remove the old token streaming display toggle from the TUI runtime so streaming transcript lines always render continuously.
- Force stored `tokenStreaming` settings to `true`, including legacy disabled values and future attempted disables.
- Keep the legacy `/stream` command as a compatibility no-op that reasserts continuous rendering.

## Test plan
- [x] `LETTA_BASE_URL= LETTA_SETTINGS_BASE_URL= bun test src/settings-manager.test.ts`
- [x] `bun run check`


overlord-letta:
Closing this PR because it targeted the wrong repo/surface for the request.

Issue #2761, opened 2026-06-06
[bug]  unable to connect to some provider using `letta connect`


KY64:
I'm unable to use `letta connect` for connecting to provider. Here is the result:

```
letta connect openai --api-key a
Validating OpenAI API API key...
Failed to connect OpenAI API: Settings not initialized. Call settingsManager.initialize() first.
```

```
letta connect openai
OpenAI API API key: 
Validating OpenAI API API key...
Failed to connect OpenAI API: Settings not initialized. Call settingsManager.initialize() first.
```

```
letta connect openrouter
OpenRouter API API key: 
Validating OpenRouter API API key...
Failed to connect OpenRouter API: Settings not initialized. Call settingsManager.initialize() first.
```

However, it works for `chatgpt`

```
letta connect chatgpt
Already connected to ChatGPT via OAuth. Use /connect in the TUI and select ChatGPT / Codex plan to disconnect or re-authenticate.
```

and if I pass `--backend local`

```
letta --backend local connect openai
OpenAI API key: 
Validating OpenAI API key...
Saving provider...
Connected OpenAI (openai) in local storage.
```

```
letta --backend local connect openrouter
OpenRouter API key: 
Validating OpenRouter API key...
Saving provider...
Connected OpenRouter (openrouter) in local storage.
```

### Version

0.27.5 and also tested commit https://github.com/letta-ai/letta-code/commit/33a21140af0c854597636eb24f29a7a2756f848e

### Repro steps

1. Install Letta Code CLI
2. Login
3. Run `letta connect openai` or `letta connect openrouter`

### Expected behavior

It should be able to store the API key and connect to the provider.

### Actual behavior

Unable to connect to provider and returns

```
letta connect openai
OpenAI API API key: 
Validating OpenAI API API key...
Failed to connect OpenAI API: Settings not initialized. Call settingsManager.initialize() first.
```

### Known workarounds

We can pass `--backend local` or we can connect it by entering the TUI and type `/connect`

### Related information

- Operating System: GNU/Linux

### Do you have time to contribute to fix?

Yes, will spend my time to look into this issue.

cpacker:
Fixed by #2762

PR #2762, opened 2026-06-06
fix(cli): initialize settings before provider validation

KY64:
## Summary

This PR will solve https://github.com/letta-ai/letta-code/issues/2761 where there is missing setting configuration when using API Key with `--backend api`. This PR will call `ensureSettingsReady` whenever the backend target is not `local`

## Changes

- **src/cli/subcommands/connect.test.ts** Added test to ensure setting is properly initialized respecting the backend target
- **src/cli/subcommands/connect.ts** Initialize settings if backend target is `api`

### DIFF
--- a/src/cli/subcommands/connect.test.ts
+++ b/src/cli/subcommands/connect.test.ts
@@ -110,6 +110,54 @@ describe("connect subcommand", () => {
     );
   });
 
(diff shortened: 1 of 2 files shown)


KY64:
The failure check is due to Github CLI

image

cpacker (approved):

Issue #2764, opened 2026-06-06
[bug] saved backend preference is not applied before CLI subcommands


KY64:
`letta backend local` saves local mode as the default backend, but `letta connect ...` does not consistently respect that saved preference.

For example, after setting the default backend to local:

```bash
letta backend local
```

I would expect:

```bash
letta connect xiaomi
```

to behave like:

```
letta --backend local connect xiaomi
```

However, it returns this

```sh
letta connect xiaomi                                                                            
Unknown provider: xiaomi. Supported providers: chatgpt (alias: codex), anthropic, openai, zai, zai-coding, minimax, gemini, moonshot, kimi-code, openrouter, bedrock, codex
```

It seems the list of model only appear if we use `--backend api`. This creates an inconsistency of the saved backend preference since the subcommand only read from explicit backend flag otherwise it assumes the `backend` is set to `api` by default instead of reading it from preferred settings.

### Version

Commit https://github.com/letta-ai/letta-code/commit/33a21140af0c854597636eb24f29a7a2756f848e

### Repro steps

1. Set default backend to local:

```bash
   letta backend local
```

2. Run connect without an explicit backend flag:

```bash
   letta connect xiaomi
```

3. Compare with explicit backend override:

```bash
   letta --backend local connect xiaomi
```

### Expected behavior

Saved backend preference should be used as the default backend before subcommands run.

In other words:

```bash
letta backend local
letta connect xiaomi
```

should behave consistently with:

```bash
letta --backend local connect xiaomi
```

unless the user explicitly overrides it with `--backend api`

### Actual behavior

```sh
letta connect xiaomi                                                                            
Unknown provider: xiaomi. Supported providers: chatgpt (alias: codex), anthropic, openai, zai, zai-coding, minimax, gemini, moonshot, kimi-code, openrouter, bedrock, codex
```

`letta connect xiaomi` runs before normal startup initializes settings and applies `settings.preferredBackendMode`

So connect resolves providers using the default/API backend instead of the saved local backend preference. Depending on the provider path and nearby initialization behavior, this can either:

- save the provider credential in the wrong backend, which is API store instead of local storage
- fail with unrelated API/settings errors even though the user selected local as the default.

### Known workarounds

Pass the backend explicitly:

```
letta --backend local connect xiaomi
```

or use the TUI flow where backend startup has already been initialized.

### Related information

- Operating system: GNU/Linux

### Do you have time to contribute to fix?

Yes

PR #2766, opened 2026-06-06
fix(cli): apply saved preferredBackendMode before routing subcommands

KY64:
## Summary
 
This PR fixes issue https://github.com/letta-ai/letta-code/issues/2764 where `letta connect` ignores the saved backend preference. Running `letta backend local;
letta connect xiaomi` would resolve providers to the API storage target instead of local, while `letta --backend
local connect xiaomi` worked correctly.
 
The root cause: the subcommand early-exit in `main()` happened before `settingsManager.initialize()`, so
`defaultProviderStorageTarget()` always saw the API backend — the saved `preferredBackendMode` was never applied. This causes 2 issues:

1. The list of provider will always take from `api` backend even if the preference is `local`, so if user tries to connect provider that is not supported on `api`, it will fail.
2. The PR https://github.com/letta-ai/letta-code/pull/2762 solve the initialization issue, and if the PR is merged, there is a subtle inconsistent behavior. Which is the `connect` subcommand still not respecting the user preferred backend settings. If user runs `letta connect openai`, the command will succeed regardless the preference because `openai` exist on both `api` and `local`. However, *where* the credential stored is not like what we expect. When the preferred settings is `local`, the credential is still stored in the cloud since the connection option is still set as empty if user does not pass `--backend` flag. We can see from this line

https://github.com/letta-ai/letta-code/blob/33a21140af0c854597636eb24f29a7a2756f848e/src/cli/subcommands/connect.ts#L392-L408

which I assume it will use the byok-provider that checks if the connection options is empty, it will not store in local auth storage.

https://github.com/letta-ai/letta-code/blob/33a21140af0c854597636eb24f29a7a2756f848e/src/providers/byok-providers.ts#L725-L745

 
## Changes
 
- **src/settings-manager.ts** Added `readPreferredBackendModeSync()` method to `SettingsManager` — reads
`preferredBackendMode` directly from `~/.letta/settings.json` via `readJsonObjectSync()` and `getSettingsPath()`
without triggering full initialization (default creation, dirty-key tracking, rollback writes, etc.)
- **src/index.ts** Before routing subcommands, if no explicit `--backend` flag is given, call
`settingsManager.readPreferredBackendModeSync()` and `configureBackendMode()` with the result
- **src/settings-manager.test.ts** Added 6 tests for `readPreferredBackendModeSync`: missing file, no preference
set, reads `"local"`, reads `"api"`, invalid value, malformed JSON

## Concern

This fix may be simple but I also got feedback from Letta to consider the backward compatibility with `letta local-backend migrate-transcripts` since it may cause error. I need feedback on how best to approach this since this tool also in beta I guess (not yet 1.0) so breaking change is expected.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -84,6 +84,7 @@ import {
 } from "./permissions/mode";
 import {
   type Settings,
+  readPreferredBackendModeSync,
   settingsManager,
   shouldPersistSessionState,
 } from "./settings-manager";
@@ -690,6 +691,20 @@ async function main(): Promise<void> {
       error instanceof Error ? `Error: ${error.message}` : String(error),
     );
     process.exit(1);
+  }
+
+  // Configure backend from saved preference before routing subcommands.
+  // Without this, subcommands like `connect` that depend on
+  // defaultProviderStorageTarget() would always see "api" when no
+  // explicit --backend flag is given, even if the user previously ran
+  // `letta backend local`.  Uses a lightweight file read to avoid
+  // triggering full SettingsManager initialization (default creation,
+  // dirty-key tracking, rollback writes, etc.).
+  if (!explicitBackendMode) {
+    const saved = readPreferredBackendModeSync();
+    if (saved) {
+      configureBackendMode(saved);
+    }
   }
 
   // Early exit for CLI subcommands (e.g., `letta server`, `letta memory`).
--- a/src/index.ts
+++ b/src/index.ts
@@ -84,7 +84,6 @@ import {
 } from "./permissions/mode";
 import {
   type Settings,
-  readPreferredBackendModeSync,
   settingsManager,
   shouldPersistSessionState,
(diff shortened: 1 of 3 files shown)


cpacker:
Thanks for putting this together — the core bug and approach were right.

I opened a replacement PR at #2772 because the fix needed a small precedence adjustment before merge: saved `preferredBackendMode` should apply to backend-aware subcommands like `connect`, but it should not override stronger selectors such as explicit `--backend`, `LETTA_LOCAL_BACKEND_EXPERIMENTAL`, or a custom `LETTA_BASE_URL`/self-hosted API target.

The replacement PR keeps your lightweight pre-subcommand settings read idea, adds the precedence handling/tests, and I added you as a co-author on the commit so you get credit for the fix.

KY64:
@cpacker got it. Thank you for reviewing this PR!

PR #2767, opened 2026-06-07
fix(local): sync model token limits from pi catalog

cpacker:
## Summary

- Derive local model context/output-token settings from the pi-ai catalog for configured local providers.
- Reset model-specific local settings on model switches so stale output-token caps cannot persist across DeepSeek → Kimi-style changes.
- Cover agent and conversation-scoped local model updates with regressions for pi catalog token limits.

## Verification

- `bun test src/backend/local-backend.test.ts src/agent/modify-local.test.ts`
- `bun run check`

### DIFF
--- a/src/backend/local/local-store.ts
+++ b/src/backend/local/local-store.ts
@@ -191,9 +191,6 @@ function supportedConversationModelSettingsFromBody(
   if (modelSettings === null) return null;
 
   const next = modelSettings ?? {};
-  if (typeof bodyRecord.context_window_limit === "number") {
-    next.context_window_limit = bodyRecord.context_window_limit;
-  }
   if (
     typeof bodyRecord.max_tokens === "number" ||
     bodyRecord.max_tokens === null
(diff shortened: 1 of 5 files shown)


src/backend/local/local-store.ts
```diff
@@ -1515,6 +1538,26 @@ export class LocalStore {
     return updated;
   }
 
+  private applyConversationModelDefaults(
+    conversation: StoredConversation,
+    body: ConversationUpdateBody,
+  ): StoredConversation {
+    const requestedModel = (body as Record<string, unknown>).model;
```
amelia-letta, line 1548: `applyConversationModelDefaults` applies catalog defaults whenever `body.model` is a string, even if the model hasn't changed. The agent update path (L1315) has a `modelChanged` guard that only applies defaults on model switches. This means a same-model conversation update (e.g. just toggling `parallel_tool_calls`) will re-apply catalog defaults and overwrite any user-customized `context_window_limit`/`max_tokens`, while the agent path preserves them. Intentional difference or missing guard?

cpacker: Good catch — that was a missing guard. I updated the conversation path to only apply pi catalog defaults when the conversation model actually changes, and added a regression that same-model conversation updates preserve explicit context/max-token overrides.

PR #2769, opened 2026-06-07
fix(tools): manage ripgrep resolution

### LINKED ISSUE #2759: **Bug:  @vscode/ripgrep platform binary mismatch on Windows
Location: node_modules/@vscode/ripgrep/bin/rg
Symptom: "How do you want to open this file?" dialog when Grep tool is invoked
Root cause: Postinstall script checks if (binExists) return without checking platform compatibility
Workaround: Delete the bin/ folder and reinstall, or download the correct Windows binary manually

Details:

The `@vscode/ripgrep` package's postinstall script (line 94-98 of `lib/postinstall.js`) checks if `bin/` folder exists and exits without downloading if it does. This causes issues when:

1. Letta Code is installed on one platform (e.g., macOS) with the `bin/rg` binary
2. Then moved/used on Windows
3. The wrong platform binary persists and Windows prompts "How do you want to open this file?" when trying to execute `rg`

**Fix needed**: Either check platform compatibility before skipping download, or provide a `--force` flag that users can invoke to refresh the binary.

**Workaround**: Delete `node_modules/@vscode/ripgrep/bin/` and run `node lib/postinstall.js --force` or manually download the correct Windows binary.

cpacker:
## Summary

- Replace direct `@vscode/ripgrep` usage in Grep/Glob with a Pi-style resolver that validates `rg --version` before using a binary.
- Prefer a Letta-managed `~/.letta/bin/rg` cache, then system `rg`, then a validated bundled fallback; download the correct BurntSushi/ripgrep asset when no valid binary exists.
- Use the same validated resolver for shell PATH injection so stale platform-mismatched bundled binaries are not prepended.
- Add regression coverage for preferring valid managed ripgrep and skipping stale managed binaries.

### DIFF
--- a/src/tools/impl/ripgrep-manager.ts
+++ b/src/tools/impl/ripgrep-manager.ts
@@ -10,7 +10,7 @@ import {
 } from "node:fs";
 import { createRequire } from "node:module";
 import { arch, homedir, platform } from "node:os";
-import { dirname, join } from "node:path";
+import { dirname, isAbsolute, join } from "node:path";
 import { Readable } from "node:stream";
 import { pipeline } from "node:stream/promises";
 import { fileURLToPath } from "node:url";
@@ -58,8 +58,8 @@ const RIPGREP_CONFIG: ToolConfig = {
   },
 };
 
-function isOfflineModeEnabled(): boolean {
-  const value = process.env[OFFLINE_ENV];
+function isOfflineModeEnabled(env: NodeJS.ProcessEnv = process.env): boolean {
+  const value = env[OFFLINE_ENV];
   if (!value) return false;
   return (
     value === "1" ||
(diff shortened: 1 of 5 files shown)


src/tools/impl/ripgrep-manager.ts
```diff
@@ -0,0 +1,407 @@
...
+      );
+    }
+    return undefined;
+  }
+}
+
+export function getRipgrepBinDir(): string | undefined {
+  const rgPath = getRipgrepPath();
+  return rgPath ? dirname(rgPath) : undefined;
+}
```
amelia-letta: When `getRipgrepPath()` returns a bare command name like `"rg"` (system binary found via PATH), `dirname("rg")` returns `"."` — this gets prepended to the shell PATH in `shell-env.ts`. Adding `"."` to PATH is a security risk and unexpected behavior.

```suggestion
export function getRipgrepBinDir(): string | undefined {
  const rgPath = getRipgrepPath();
  if (!rgPath || !path.isAbsolute(rgPath)) return undefined;
  return dirname(rgPath);
}
```

src/tools/impl/ripgrep-manager.ts
```diff
@@ -0,0 +1,407 @@
...
+  createWriteStream,
+  existsSync,
+  mkdirSync,
+  readdirSync,
+  renameSync,
+  rmSync,
+} from "node:fs";
+import { createRequire } from "node:module";
+import { arch, homedir, platform } from "node:os";
+import { dirname, join } from "node:path";
```
amelia-letta: Need `isAbsolute` in the named import for the `getRipgrepBinDir` fix to work:

```suggestion
import { dirname, isAbsolute, join } from "node:path";
```

src/tools/impl/ripgrep-manager.ts
amelia-letta: Updated suggestion (with `isAbsolute` added to the import on line 13):

```suggestion
export function getRipgrepBinDir(): string | undefined {
  const rgPath = getRipgrepPath();
  if (!rgPath || !isAbsolute(rgPath)) return undefined;
  return dirname(rgPath);
}
```

cpacker: Fixed — getRipgrepBinDir now only returns a directory for absolute rg paths, so a system PATH hit like `rg` no longer contributes `.` to PATH. I also moved the ripgrep-manager tests to pass an explicit env into the resolver instead of mutating process.env/PATH globally; that should fix the Linux/macOS full-suite failure from test interference.

PR #2776, opened 2026-06-08
fix: declare yaml runtime dependency for skill scripts

### LINKED ISSUE #2775: creating-skills validator imports yaml but desktop app bundle cannot resolve it with Bun
## Summary

The `creating-skills` bundled validator/packager scripts import `yaml`, but the Letta Code desktop app bundle does not appear to include that dependency where Bun can resolve it. Running the validator directly with Bun fails before it can validate a skill.

## Environment

- Letta Code desktop app on macOS
- Bun 1.3.14
- GitHub CLI auth account: `jigonr`
- Script path:

```text
/Applications/Letta.app/Contents/Resources/app.asar.unpacked/node_modules/@letta-ai/letta-code/skills/creating-skills/scripts/validate-skill.ts
```

## Repro

Using a skill directory with a valid `SKILL.md`, run:

```bash
SK=/Applications/Letta.app/Contents/Resources/app.asar.unpacked/node_modules/@letta-ai/letta-code/skills/creating-skills/scripts

bun "$SK/validate-skill.ts" skills/planning-course-materials
```

The same happens when using `bun x tsx` without extra module resolution:

```bash
bun x tsx "$SK/validate-skill.ts" skills/planning-course-materials
```

## Actual behaviour

```text
error: Cannot find package 'yaml' from '/Applications/Letta.app/Contents/Resources/app.asar.unpacked/node_modules/@letta-ai/letta-code/skills/creating-skills/scripts/validate-skill.ts'

Bun v1.3.14 (macOS arm64)
```

The import in `validate-skill.ts` is:

```ts
import { parse as parseYaml } from "yaml";
```

I also checked the app-bundled `@letta-ai/letta-code/package.json`; `yaml` did not appear in the package dependencies, and there was no nearby `yaml` directory under the app-bundled `node_modules` in my install.

## Expected behaviour

The bundled skill validator and packager should run from the Letta Code app bundle without requiring the user to have a separate user-level `yaml` dependency installed or to manually set `NODE_PATH`.

## Workaround

If `yaml` is installed in the user's local `node_modules`, this works:

```bash
NODE_PATH="$(dirname "$(bun pm bin)")" bunx tsx \
  "$SK/validate-skill.ts" \
  skills/planning-course-materials
```

Packaging also works with:

```bash
NODE_PATH="$(dirname "$(bun pm bin)")" bun \
  "$SK/package-skill.ts" \
  skills/planning-course-materials \
  /tmp/skill-validation
```

With that workaround, the validator reports:

```text
Skill is valid!
```

and packaging succeeds.

## Notes

There was a separate skill-authoring issue in my first test: an unquoted colon in the YAML frontmatter `description` field. After quoting the description, validation succeeded with the `NODE_PATH` workaround. That is separate from this dependency-resolution issue.

## Suggested fix

Either:

1. add `yaml` to the relevant `@letta-ai/letta-code` package dependencies and ensure it is included in the desktop app bundle;
2. bundle/vendor the validator dependencies with the app;
3. avoid the external `yaml` dependency in `validate-skill.ts`; or
4. document the Bun-compatible workaround in the `creating-skills` skill docs.

xianzuyang9-blip:
## Summary
- add yaml to the published package dependencies so bundled creating-skills scripts can resolve it from the desktop/app package
- sync the Bun lock root dependency entry
- add a regression guard that checks bundled skill scripts do not import undeclared external runtime packages

## Verification
- 
ode static dependency guard: bundled creating-skills scripts external imports = ["yaml"], missing dependencies = []

Note: I could not run the repository's Bun test suite in this Windows environment because un is not installed here.

Fixes #2775

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -16,6 +16,7 @@
         "shiki": "^4.0.2",
         "strip-ansi": "^7.2.0",
         "ws": "^8.19.0",
+        "yaml": "^2.8.1",
       },
       "devDependencies": {
         "@slack/bolt": "^4.7.0",
--- a/package.json
+++ b/package.json
@@ -46,7 +46,8 @@
     "sharp": "^0.34.5",
     "shiki": "^4.0.2",
     "strip-ansi": "^7.2.0",
-    "ws": "^8.19.0"
+    "ws": "^8.19.0",
+    "yaml": "^2.8.1"
   },
   "optionalDependencies": {
     "@vscode/ripgrep": "^1.17.0"
--- a/src/skills/skill-creator-scripts.test.ts
+++ b/src/skills/skill-creator-scripts.test.ts
@@ -3,7 +3,14 @@
  */
 
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
-import { existsSync, mkdirSync, rmSync, writeFileSync } from "node:fs";
+import {
+  existsSync,
+  mkdirSync,
+  readdirSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
 import { join } from "node:path";
 import {
   initSkill,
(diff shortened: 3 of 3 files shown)


cpacker:
Thanks for the fix here — this is the same underlying issue as #2775. We opened #2778 with a slightly different approach: instead of adding `yaml` as a runtime package dependency, it makes the bundled skill scripts self-contained and adds a CI guard to prevent future bundled skill script package dependencies/manifests.

I added you as a co-author on the #2778 commit for credit since your PR identified and addressed the same bug path.

PR #2777, opened 2026-06-08
fix(agent/import): path traversal when importing agent files

KY64:
## Summary

Fixes an AgentFile (`.af`) import path traversal issue in skill extraction. When importing skills from an AgentFile, the importer previously trusted both:

- `skills[].name`
- keys inside `skills[].files`

These values were used to build local filesystem paths with `resolve()` without validating that the resulting
paths stayed inside the intended imported skill directory. So instead of writing a new skills in `~/.letta/agets/<agent-id>/memory/skills`, it is possible to **write** files on **ANY** directory that normal user is allowed to. This PR adds validation for imported skill names and confines embedded skill file writes to the target skill directory.

## Security impact

Before this change, a malicious AgentFile could write files outside the intended skill directory during `letta
--import`.

Example malicious embedded skill file entry:

```json
{
  "skills": [
    {
      "name": "safe-skill",
      "files": {
        "../../outside-marker.txt": "written outside skill dir"
      }
    }
  ]
}
```

The old import path resolved this relative to `<agent-memory>/skills/safe-skill/` which allowed the write to escape into a parent directory. A malicious `skills[].name` could also be used as a path segment:

```json
{
  "skills": [
    {
      "name": "../../outside-skill-dir",
      "files": {
        "SKILL.md": "# escaped"
      }
    }
  ]
}
```

## POC

The POC agent file can be found here https://github.com/KY64/letta-af-path-traversal-poc
We can clone it and import it using `letta --import` for example.

- [poc-write-file-outside.af](https://github.com/KY64/letta-af-path-traversal-poc/blob/main/poc-write-file-outside.af) is POC which uses a valid skill name but the skill file is written outside the skill directory. Also it is possible to **overwrite** any file with a content that is described in the value.
- [poc-skill-name-traversal.af](https://github.com/KY64/letta-af-path-traversal-poc/blob/main/poc-skill-name-traversal.af) is POC which uses filepath as the skill name


cpacker (changes requested):
Thanks for fixing this — the path traversal protection itself looks directionally right, and I validated the branch in a temp merge worktree with:

- `bun run check`
- `bun test ./src/agent/import-skills.test.ts`

One blocker before merge: `MAX_IMPORTED_SKILL_NAME_LENGTH` is set to 60, but our skill spec/creator validator allows names up to 64 characters. That makes some valid Letta Code skills fail to round-trip through `.af` export/import. Please align this limit with the existing 64-character skill-name limit and update the regression test accordingly.

Small cleanup while touching that function: consider rejecting names where `name !== name.trim()` instead of silently normalizing, so imported names cannot collide after trimming.

The failed `review` check appears unrelated to this PR code: the action failed because `gh` did not have `GH_TOKEN` configured.

KY64:
Sure, let me update

KY64:
Just updated, can you review it @cpacker ?
I decided to export the constant from `src/skills/builtin/creating-skills/scripts/validate-skill.ts` though I guess there should be a global constant for this shared constraint.

cpacker:
Can you resolve the conflict against main?

KY64:
> Can you resolve the conflict against main?

Sure, let me look into it

KY64:
@cpacker Resolved

cpacker (approved):
Re-reviewed the updated PR. The previous blocker is addressed: the import path now uses the shared 64-character limit, rejects trimmed-name mismatches, and the branch is cleanly merged with current `main`.

Validated locally in a fresh temp merge worktree:

- `bun run check`
- `bun test ./src/agent/import-skills.test.ts ./src/skills/skill-creator-scripts.test.ts`

The remaining failed `review` check is still the workflow `GH_TOKEN` configuration issue, not a code failure.

PR #2782, opened 2026-06-08
fix(tools): support worktree creation from explicit repo path

kianjones9:
## Summary
- Add `repo_path` to CreateWorktree so sessions started outside a git repo can still target the intended repository.
- Return an actionable outside-repo error instead of raw `git rev-parse` failure output.
- Cover the new repo_path behavior in the existing CreateWorktree test suite.

## Test plan
- [x] `bun test src/tools/create-worktree.test.ts`
- [x] `bun run check`

### DIFF
--- a/src/tools/create-worktree.test.ts
+++ b/src/tools/create-worktree.test.ts
@@ -320,6 +320,36 @@ describe("CreateWorktree tool", () => {
     expect(remoteFile.replace(/\r\n/g, "\n")).toBe("latest remote\n");
   });
 
+  test("creates a worktree from repo_path when current cwd is outside a git repository", async () => {
+    const repo = await trackRepo();
+    const dir = await mkdtemp(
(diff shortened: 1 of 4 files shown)


cpacker (approved):
Reviewed the implementation and this looks good to merge.

What I checked:
- `repo_path` is added to the tool schema/description and resolved relative to the current cwd when provided.
- Existing cwd-based behavior is preserved when `repo_path` is omitted.
- Outside-repo failures now return an actionable message telling the model to pass `repo_path` rather than surfacing only raw `git rev-parse` output.
- The new test covers creating a worktree from `repo_path` while the runtime cwd is outside a git repository.

Validated locally in a fresh temp merge worktree:
- `bun run check`
- `bun test src/tools/create-worktree.test.ts`

cpacker (approved):

PR #2790, opened 2026-06-09
docs: tighten plan mode extension example

carenthomas:
## Summary
- make the plan-mode example require full-plan AskUserQuestion approval text
- expand read-only tool aliases and strengthen tool-name normalization
- document recall-only subagent behavior and clearer denial recovery guidance

## Test
- git diff --check

### DIFF
--- a/src/skills/builtin/creating-extensions/references/plan-mode.md
+++ b/src/skills/builtin/creating-extensions/references/plan-mode.md
@@ -24,12 +24,14 @@ This is a pattern reference, not a full product implementation. Keep local exten
 -> remind the agent that only read-only tools and plan-file writes are allowed
 -> permission overlay denies mutations outside ~/.letta/plans/*.md
 -> agent writes the plan with normal Write/Edit/ApplyPatch tools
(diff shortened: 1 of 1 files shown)


src/skills/builtin/creating-extensions/references/plan-mode.md
```diff
@@ -181,30 +185,46 @@ Use a permission overlay, not `tool_start`, for policy. Normalize tool names by
...
+  "readmanyfiles",
   "search",
-  "search_file_content",
+  "searchfilecontent",
+  "searchfiles",
   "skill",
   "taskoutput",
-  "update_plan",
-  "view_image",
+  "todowrite",
```
amelia-letta, line 213: `todowrite` and `writetodos` are write tools — they create/modify todo items, not read. Including them in the "read-only" set lets the agent mutate todos in plan mode, which contradicts the "only read-only tools and plan-file writes" policy. Should these move out and be handled by a separate allowlist (like `updateplan` is effectively a plan-file write)?

PR #2794, opened 2026-06-09
feat(models): add Claude Fable 5

overlord-letta:
## Summary
- Adds a Fable 5 model preset for the Anthropic handle exposed by Core: anthropic/claude-fable-5.
- Keeps the Letta Code preset conservative at 200k context, matching the existing Opus 4.8 registry behavior even though the model advertises a larger server-side context window.
- Adds a registry test so the short fable id resolves to the expected label, handle, context, max output, and tool settings.

Cloud/Core support:
- https://github.com/letta-ai/letta-cloud/pull/12189

## Validation
- bun test src/agent/model-tier-selection.test.ts
- bun run check

### DIFF
--- a/src/models.json
+++ b/src/models.json
@@ -11,7 +11,8 @@
         "context_window": 140000,
         "max_output_tokens": 28000,
         "parallel_tool_calls": true
-      }
+      },
+      "isFeatured": true
     },
     {
       "id": "auto-fast",
@@ -23,7 +24,8 @@
         "context_window": 140000,
         "max_output_tokens": 28000,
         "parallel_tool_calls": true
-      }
+      },
+      "isFeatured": true
     },
     {
       "id": "auto-chat",
@@ -35,7 +37,8 @@
         "context_window": 140000,
         "max_output_tokens": 28000,
         "parallel_tool_calls": true
-      }
+      },
+      "isFeatured": true
     },
     {
       "id": "fable",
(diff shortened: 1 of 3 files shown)


src/models.json
```diff
@@ -244,6 +244,19 @@
...
+    {
+      "id": "fable",
+      "handle": "anthropic/claude-fable-5",
+      "label": "Fable 5",
+      "description": "Anthropic's most capable model",
+      "isFeatured": true,
+      "updateArgs": {
+        "context_window": 200000,
+        "max_output_tokens": 128000,
+        "enable_reasoner": true,
```
amelia-letta: Missing `reasoning_effort` — every other Anthropic model with `enable_reasoner: true` also sets this. Sonnet and Opus default to `"high"`.

```suggestion
        "enable_reasoner": true,
        "reasoning_effort": "high",
        "parallel_tool_calls": true
```

src/models.json
```diff
@@ -32,19 +30,235 @@
...
+      "updateArgs": {
+        "context_window": 200000,
+        "max_output_tokens": 128000,
+        "reasoning_effort": "max",
+        "enable_reasoner": true,
+        "parallel_tool_calls": true
+      }
+    },
+    {
+      "id": "opus-1m",
```
amelia-letta: Handle changed from `claude-opus-4-6` to `claude-opus-4-8` — existing users with `opus-1m` selected will silently get a different model on next message. Intentional upgrade or should this be a new ID?

PR #2800, opened 2026-06-09
refactor(local): route model catalog through Pi registry

cpacker:
## Summary
- Stack on #2797 and add a Pi Coding Agent `ModelRegistry`/`AuthStorage` adapter for local backend model catalog metadata.
- Route built-in provider runtime lookup, `/model` listing metadata, and context/max-token settings through Pi’s registry instead of LC reading `pi-ai` catalogs directly.
- Keep LC-specific responsibilities limited to local auth storage mapping, endpoint discovery, handle/provider aliases, and API-shaped local backend projection.

## Review focus
- Please check for any unnecessary drift from Pi TUI / Pi Coding Agent semantics in `pi-ai`-related local backend code.
- Remaining LC-specific logic should be adapter glue only: storage location, auth UI records, endpoint discovery, aliases, and API-shaped projection.
- This PR intentionally does not deep-import Pi’s unexported `findInitialModel` resolver; completing that cleanup should happen after Pi exposes that selection API publicly.

### DIFF
--- a/src/backend/dev/pi-model-registry-adapter.ts
+++ b/src/backend/dev/pi-model-registry-adapter.ts
@@ -1,0 +1,158 @@
+import type { Api, Model } from "@earendil-works/pi-ai";
+import {
+  getOAuthProvider,
+  type OAuthCredentials,
+} from "@earendil-works/pi-ai/oauth";
+import {
+  type AuthCredential,
+  AuthStorage,
+  ModelRegistry,
+} from "@earendil-works/pi-coding-agent";
+import {
+  type LocalProviderRecord,
+  listLocalProviderRecords,
+} from "@/backend/local/local-provider-auth-store";
+import {
+  getPiProviderSpec,
+  isPiProvider,
+  type PiProvider,
+  resolveProviderFromModelHandle,
+  resolveProviderFromProviderType,
+  stripProviderHandlePrefix,
+} from "./pi-provider-registry";
+
+// Adapter boundary: LC owns local storage paths, auth UI records, endpoint
+// discovery, and API-shaped projection. Built-in provider catalog/auth model
+// metadata should come from Pi's ModelRegistry so local backend behavior stays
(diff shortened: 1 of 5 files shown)


src/backend/local/local-model-config.ts
```diff
@@ -14,7 +17,6 @@ import {
 import {
```
amelia-letta, line 17: Since listCatalogModelsForProvider is no longer imported here, the export in pi-provider-registry.ts is dead code. Should remove it from there too (along with the getModels call at L393 if that was its last consumer in that file).

src/backend/dev/pi-model-registry-adapter.ts
```diff
@@ -0,0 +1,158 @@
...
+  for (const record of listLocalProviderRecords(storageDir)) {
+    const providerId = piProviderIdForRecord(record);
+    const credential = credentialForRecord(record);
+    if (!providerId || !credential) continue;
+    credentials[providerId] = credential;
+  }
+  return AuthStorage.inMemory(credentials);
+}
+
+export function createLocalPiModelRegistry(storageDir?: string): ModelRegistry {
```
amelia-letta, line 79: createLocalPiModelRegistry reads all provider records from disk and builds a new registry on every call. In listLocalModels this gets called once per provider (listPiRegistryCatalogModelsForProvider) plus once per model handle (via addModel -> localModelSettingsForHandle -> findPiRegistryModelForHandle). The old getModels() was a static import. Worth caching the registry for the duration of a single listLocalModels call?

cpacker:
Closing this stacked refactor. The direction is right in spirit, but importing `@earendil-works/pi-coding-agent` as a Letta Code runtime dependency is too broad for the local backend adapter and pulls in unrelated CLI/TUI surface area. We should only do this after Pi exposes a narrow model registry/resolver package or public export that LC can depend on without the coding-agent dependency blast radius.

PR #2820, opened 2026-06-10
fix(memfs): push immediately after non-turn memory writes

christinatong01:
Companion fix to #2819.

## Problem

The June 4 refactor (`4f851eff`) moved the push out of `commitMemoryWrite` and into `syncPendingMemoryCommitsAfterTurn`, which only fires after a turn. Non-turn writes like **profile picture uploads** (direct WS command) never trigger post-turn sync, so commits sit locally indefinitely.

This caused profile pictures to appear briefly (optimistic UI) then disappear on refetch — the commit existed locally but was never pushed to the server.

## Fix

Fire `syncPendingMemoryCommitsAfterTurn` in the background immediately after a successful commit in the `write_memory_file` WS handler. The response is sent to the client immediately (non-blocking); the push happens async.

Skipped for local-only backends (`syncMode: "local"`).

## Also

Added `console.log`/`console.warn` across the write path (checkout, commit, push) so failures are visible in desktop debug logs.

### DIFF
--- a/src/websocket/listener/commands/memory.ts
+++ b/src/websocket/listener/commands/memory.ts
@@ -743,6 +743,21 @@ export function handleMemoryProtocolCommand(
           ...(memorySyncMode ? { syncMode: memorySyncMode } : {}),
         });
 
+        // ── Push immediately (non-turn writes don't get post-turn sync) ─
+        if (commitResult.committed && !memorySyncMode) {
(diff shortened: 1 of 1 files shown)


amelia-letta (commented):
same gap in `delete_memory_file` — L925 does a commitMemoryWrite with no immediate push, and the comment on L922 ("post-turn harness sync handles remote push") is the same assumption this PR is fixing for writes.

PR #2837, opened 2026-06-11
fix(tui): add right padding to tool output cards

kianjones9:
## Summary
- reserve a right-padding column for tool result text content
- apply the same width budget to collapsed and streaming shell output

## Tests
- bun test src/cli/create-worktree-result-renderer.test.tsx src/cli/markdown-display-code-highlighting.test.tsx
- git diff --check

Note: `bun run typecheck` is currently blocked by an unrelated `src/backend/dev/pi-provider-registry.ts` KnownProvider type error on main.


kianjones9:
Closing: this fixed the TUI rendering path, but the requested issue is in the Desktop/web tool call card padding.

PR #2854, opened 2026-06-11
feat(mods): add web search examples for local agents

overlord-letta:
## Summary
- Add a bundled `web-search-mods` skill with local mod examples for Perplexity, Exa, and Parallel.
- Link the skill from the existing mod-writing guidance and README local-mode section so local agents can discover the path when they need web search.
- Cover the bundled skill references in the existing skill-discovery test.

## Why
Local agents do not get hosted web-search tools automatically. The current extension surface is called mods, and model-callable mod tools are already the right local mechanism. This PR adds copy-pasteable provider examples instead of adding a new first-class built-in search tool or taking a dependency on provider SDKs.

## Behavior
After this change, a user can ask Letta Code for web-search mods and get provider-specific templates for:
- Perplexity via `PERPLEXITY_API_KEY`
- Exa via `EXA_API_KEY`
- Parallel via `PARALLEL_API_KEY`

Each template registers one local tool and uses `fetch`, so the examples stay dependency-free.

## Limits
I did not make live provider calls because provider API keys are not available in this environment. The validation here proves discovery and packaging, not external provider connectivity.

## Test plan
- [x] `bun test ./src/agent/skills-discovery.test.ts`
- [x] `bun run check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,8 @@ letta --backend local --new-agent
 
 Local agents do not appear in the Constellation, but their memory is still a normal git repository under `~/.letta/lc-local-backend/memfs/<agent-id>/memory`.
 
+Local agents can add model-callable web search through trusted local mods in `~/.letta/mods/`. The bundled [`web-search-mods` skill](src/skills/builtin/web-search-mods/SKILL.md) includes Perplexity, Exa, and Parallel examples.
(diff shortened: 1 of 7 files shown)


sarahwooders:

PR #2857, opened 2026-06-11
feat: add app-server external tool callbacks

carenthomas:
## Summary
- add app-server protocol frames for external tool registration and callback execution
- expose scoped external tools per input turn via external_tool_scope_ids
- add listener-side request/response bridge and app-server client helpers

## Tests
- bun run check
- bun test src/websocket/listener/external-tools.test.ts src/tools/tool-execution-context.test.ts
- bun test src/app-server-client.test.ts src/websocket/listener/protocol-inbound.test.ts src/websocket/listen-client-protocol.test.ts src/types/app-server-protocol.test.ts

### DIFF
--- a/src/app-server-client.test.ts
+++ b/src/app-server-client.test.ts
@@ -194,6 +194,69 @@ describe("app-server client", () => {
     expect(sent).toEqual(["sync", "abort_message", "input"]);
   });
 
+  test("registers external tools and responds to external tool calls", async () => {
+    const { client, control, stream } = createFakeClient();
+    control.open();
+    stream.open();
+    await client.connect();
+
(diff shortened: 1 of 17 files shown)


carenthomas:
Closing this because #2860 landed as the runtime_start-shaped version of app-server external tools.

The protocol direction from this PR, a separate `external_tools_register` command, ended up being less clean than registering controller-owned tools during `runtime_start` after the runtime is resolved. #2860 keeps the callback flow and scoped per-turn exposure, but binds tools to the resolved `{agent_id, conversation_id}` lifecycle instead of a separate side channel.

Thanks for the comparison branch, it helped validate the API shape.

PR #2858, opened 2026-06-11
Add app-server tool policy support

carenthomas:
## Summary
- add runtime/input `tool_policy` with all/none/allow/deny modes
- apply runtime policy as the baseline and input policy / legacy allowlist as narrowing filters
- store runtime default policy per conversation and apply it during tool preparation

Stacked on #2857.

## Tests
- bun run check
- bun test src/tools/tool-execution-context.test.ts
- bun test src/websocket/listener/protocol-inbound.test.ts src/websocket/listen-client-protocol.test.ts

### DIFF
--- a/src/tools/manager.ts
+++ b/src/tools/manager.ts
@@ -58,6 +58,7 @@ import {
 } from "@/runtime-context";
 import { settingsManager } from "@/settings-manager";
 import { telemetry } from "@/telemetry";
+import type { ClientToolPolicy } from "@/types/protocol_v2";
 import { debugLog } from "@/utils/debug";
 import { isRecord } from "@/utils/type-guards";
 import {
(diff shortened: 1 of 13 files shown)


carenthomas:
Closing this out after #2860 landed.

The tool policy idea is still potentially useful, but it is not needed for the current app-server external tools MVP. Codex app-server parity is covered by runtime/thread-start dynamic tools plus per-turn selection, and we already have `client_tool_allowlist` / `external_tool_scope_ids` for the immediate visibility controls.

This PR is also still stacked on the superseded #2857 explicit-register design, so keeping it open would risk pulling that abandoned protocol shape back into the diff. We can revisit a smaller runtime-level tool visibility control later if we get a concrete product need.

PR #2880, opened 2026-06-13
feat(cli): add statusline to include help command and newline keybind

KY64:
### Summary

This PR will only change TUI appearance to add simple text below input text box.

Screenshot_2026-06-13_11-58-05

I added this because `/help` seems also contain many shortcuts which TUI user may not aware.

For the newline, I tried Shift+Enter yet it sends the input, CTRL+Enter also send the input then realised it was using ALT+Enter to create newline. So I think putting below textbox would be helpful since user may need to make some bullet points to explain something or answer multiple questions from the models.

Also `/help` seems not a valid command but it works if I press enter afterward, tested this with latest commit https://github.com/letta-ai/letta-code/commit/bdd0e56ed6111d6f23150007170a898a6961fe68

image

### Concern

- I don't have Apple device on my end so I can't test, but I added the shortcut to be platform aware so for Mac it should be `Opt+Enter` but others would be `Alt+Enter`
- I changed the size of right column width under the input text box to give space for the text on the left column. Not sure what is the consequence, so far I test by writing prompt and input command.

### Changes

- **src/cli/components/HelpDialog.tsx**: Update newline shortcut to be platform aware
- **src/constants.ts**: Added constant for the shortcut text
- **src/cli/display/statusline/renderers/Default.tsx**: Added text below input text box

### DIFF
--- a/src/cli/components/HelpDialog.tsx
+++ b/src/cli/components/HelpDialog.tsx
@@ -1,6 +1,7 @@
 import { Box, useInput } from "ink";
 import { useCallback, useEffect, useMemo, useState } from "react";
 import { commands } from "@/cli/commands/registry";
+import { INPUT_NEWLINE_MODIFIER_LABEL } from "@/constants";
 import { getVersion } from "@/version";
 import { colors } from "./colors";
 import { OverlayShell } from "./OverlayShell";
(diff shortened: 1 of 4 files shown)


KY64:
The failure CI for Code Review is due to GH CLI token, while the `CI / macOS` is due to timeout.

PR #2887, opened 2026-06-13
feat(input): rotate conversation threads with Ctrl+PageUp/PageDown

seb-veto:
## Summary

Adds browser-tab-style rotation through an agent's conversation threads, driven from the main input:

- **Ctrl+PageUp** → previous thread
- **Ctrl+PageDown** → next thread

Rotation cycles across the most recently active conversations (same ordering as the `/resume` selector) with wraparound, and always includes the current thread even if it is not on the first page. The switch is performed by running the existing `/resume <id>` command, so transcript backfill, goal pausing, and the conversation-switch alert behave **exactly** like a manual resume — no new switch path to keep in sync.

The target/order math lives in a small pure, unit-tested helper (`conversation-rotation.ts`).

## Why Ctrl+PageUp/PageDown (and not Cmd+Shift+[ / ])

The original request was Chrome-style `Cmd+Shift+[ / ]`. Terminals do **not** deliver the `Cmd` (super) modifier to TUI programs — the terminal/OS intercepts it — so a literal `Cmd`-based binding cannot be captured by a terminal app like letta-code. `Ctrl+PageUp/PageDown` is the terminal-deliverable equivalent and is the same shortcut many browsers use for tab switching. The vendored Ink keypress parser already decodes `\e[5;5~` / `\e[6;5~` into `key.pageUp/pageDown + key.ctrl`, so detection is reliable. A host application (e.g. Letta Desktop) can later map `Cmd+Shift+[/]` onto these.

## Behavior details

- Fires only while the input is interactive (idle), so it never interferes with a streaming turn.
- If there are no other threads, shows a brief "No other threads to switch to" status instead of switching.
- On a list-fetch error, surfaces a clear failure message.
- Listed in the Help dialog (Shortcuts tab).

## Files changed

- `src/cli/helpers/conversation-rotation.ts` — pure `computeRotatedConversationId` + `withCurrentConversation` helpers
- `src/cli/helpers/conversation-rotation.test.ts` — 11 unit tests
- `src/cli/app/AppCoordinator.tsx` — `handleRotateConversation` (fetch recent threads → compute target → run `/resume <id>`)
- `src/cli/app/AppView.tsx` — thread the handler to the input
- `src/cli/components/InputRich.tsx` — Ctrl+PageUp/PageDown key handling
- `src/cli/components/HelpDialog.tsx` — document the shortcut

## Test plan

- [x] `bun test src/cli/helpers/conversation-rotation.test.ts` (11 pass) — next/prev, wraparound both ends, single-thread/empty/missing-current → null, duplicate-id dedupe, and current-thread inclusion
- [x] `bun run typecheck`
- [x] `biome check` on changed files
- [x] `check:boundaries`, `check:exported-functions`, `check:filename-casing`, circular-deps — clean
- [ ] Manual: with 2+ threads, Ctrl+PageDown/PageUp cycles and the transcript backfills; wraps at both ends; single-thread shows the no-op status

### DIFF
--- a/src/cli/app/AppCoordinator.tsx
+++ b/src/cli/app/AppCoordinator.tsx
@@ -75,6 +75,11 @@ import {
   resetContextHistory,
 } from "@/cli/helpers/context-tracker";
 import {
+  type ConversationRotationDirection,
+  computeRotatedConversationId,
+  withCurrentConversation,
+} from "@/cli/helpers/conversation-rotation";
+import {
   generateConversationTitleFromSummary,
   getConversationTitleSettings,
   listConversationTitleMessages,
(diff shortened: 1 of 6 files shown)


sarahwooders:

PR #2889, opened 2026-06-13
fix(listener): re-enter the goal loop on the Desktop/API surface

seb-veto:
## Summary

The autonomous `/goal` continuation loop only re-entered in the **TUI** (`src/cli/app/use-conversation-loop.ts`, via `handleGoalContinuation` → `processConversation(allowReentry)`). The **websocket listener** (Desktop/API surface) activated durable goal state and injected a **single** continuation prompt, then stopped — so a goal never iterated autonomously outside the terminal.

This wires the same end-turn re-entry into the listener path.

## Approach

- `handleIncomingMessage` is now a thin **trampoline**: the existing turn body is `runSingleTurn`, and after each clean `end_turn` the wrapper asks `buildGoalContinuationTurn` whether to take another autonomous goal turn. No recursion (a `while` loop), so no stack growth; each iteration is a full, normal turn — exactly mirroring the TUI.
- Stop conditions (in `src/websocket/listener/goal-continuation.ts`, pure + unit-tested `decideGoalContinuation`):
  - turn did not end cleanly (`lastStopReason !== "end_turn"`, i.e. cancel/error) → stop
  - goal no longer `active` (complete / blocked / paused — the per-conversation stored status set by the `update_goal` tool) → stop + restore standard permission mode
  - **hard iteration cap** `MAX_AUTONOMOUS_GOAL_ITERATIONS = 50` → stop (runaway safety net, independent of any user-facing step cap)
  - token budget exhausted → stop
- `/goal` now starts the first turn on the conversation **message queue** (`startGoalLoopTurnInBackground`) instead of `await`ing it inline, so the command returns its confirmation immediately while the loop runs serialized in the background (also prevents concurrent turns on the same conversation).
- Honest label: `(iter 1/∞)` → `(iter 1/50)`.

## Why a hard cap

This adds an autonomous loop to a non-interactive surface, so a goal the model never marks complete/blocked must not run forever. The cap guarantees termination; `update_goal` (complete/blocked) is still the normal exit, and the goal can be resumed with `/goal resume`.

## Known limitation (follow-up)

`goalLoopMode` is a process-global singleton (pre-existing). On a single listener serving **multiple** conversations with concurrent goals, the global iteration/active state can interfere across conversations. The authoritative complete/blocked signal is the per-conversation stored goal status, so completion is correctly scoped; the global iteration counter is not. A per-conversation goal-loop state is the right follow-up (tracked in a separate issue). The common Desktop case (one active goal) is unaffected.

## Test plan

- [x] `bun test src/websocket/listener/goal-continuation.test.ts` — 8 pass (continue path; stop on inactive/unclean/complete/blocked/paused/cap/budget; unclean prioritized over status)
- [x] `bun test src/websocket/listener/` — 79 pass (no regressions from the `handleIncomingMessage` rename/wrapper)
- [x] `bun run typecheck`
- [x] `biome check`, `check:boundaries`, `check:exported-functions`, `check:filename-casing`, circular-deps — clean
- [ ] Manual (Desktop): set `/goal <objective>`; confirm the agent keeps taking turns until it calls `update_goal complete`/`blocked`, the cap is hit, or the turn is cancelled; confirm permission mode returns to standard on stop

## Repro (issue anchor)

On Desktop/API, set a `/goal`; observe the agent runs exactly one turn then idles. In the TUI the same goal iterates. Root cause: re-entry exists only in `use-conversation-loop.ts`; the listener (`src/websocket/listener/commands.ts`) injects one prompt with no end-turn continuation.

### DIFF
--- a/src/websocket/listener/commands.ts
+++ b/src/websocket/listener/commands.ts
@@ -45,6 +45,7 @@ import type {
 import { debugLog } from "@/utils/debug";
 import { markSecretsReminderRefreshPending } from "./commands/secrets";
 import { getConversationWorkingDirectory } from "./cwd";
+import { MAX_AUTONOMOUS_GOAL_ITERATIONS } from "./goal-continuation";
 import { reloadListenerModAdapter } from "./mod-adapter";
 import {
   getOrCreateConversationPermissionModeStateRef,
@@ -63,7 +64,11 @@ import {
   buildMaybeLaunchReflectionSubagent,
   handleIncomingMessage,
 } from "./turn";
-import type { ConversationRuntime, StartListenerOptions } from "./types";
+import type {
+  ConversationRuntime,
+  IncomingMessage,
+  StartListenerOptions,
+} from "./types";
 
 export { SUPPORTED_REMOTE_COMMANDS } from "./listener-constants";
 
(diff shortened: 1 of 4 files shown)


sarahwooders:

PR #2892, opened 2026-06-13
fix(desktop): keep project chat grouping after first message

seb-veto:
## Summary
- Always include the listener `cwd_map` and boot working directory in device status updates, including scoped runtime updates
- Ensures Desktop can immediately retain project-folder grouping for newly-created project chats after the first message
- Adds coverage for runtime-start replay and scoped device status snapshots

## Why
Creating a new chat from a project folder can briefly show the chat under that project, but after the first message it falls back to the default/Documents bucket even though the bottom bar still shows the correct cwd. The listener has the mapping persisted, but scoped device status omitted `cwd_map`, so the client did not receive the map during scoped updates.

## Test plan
- `bun test src/websocket/listen-client-protocol.test.ts`
- `bun run typecheck`
- `bun run check`


sarahwooders:

PR #2893, opened 2026-06-13
fix(app-server): let a reconnecting control socket take over instead of rejecting it

seb-veto:
## Summary
- A new chat (and any message) **silently does nothing** after the desktop reconnects its control socket. The conversation is created but its first message never runs a turn, and no error is surfaced.
- Root cause: `startAppServer`’s control-channel handler **hard-rejected** a second control connection (`closeSocket(1008, "control channel already connected")`). When the desktop reconnects while the previous control socket is still registered (half-open after sleep/wake, or the close not yet observed by the server), the new socket cannot start runtimes or send messages.
- Notably, `startControlSession` **already** performs runtime takeover for this case, but that path was unreachable because the connection was rejected first.

## Fix
- On a new control connection, **supersede** the stale session ("newest control wins") instead of rejecting: tear down the previous runtime/sockets and let the new socket take over. This mirrors the takeover `startControlSession` already does.
- Safe against stale-close races: the superseded control socket’s `close` handler is already guarded by an active-runtime check, so it cannot clobber the new session.

## How it was found
Built a ground-truth repro driving the real `startAppServer` + `LocalBackend` over real control/stream WebSockets:
- ✅ single-socket new chats, ✅ conversation-create-without-runtime_start, ✅ graceful reconnect (close→reopen) — all already worked.
- ❌ **overlap reconnect** (old control still registered) — `runtime_start` on the new socket got **no response**, matching the symptom (conversation created on disk, empty `messages.jsonl`, no error). Fixed by this change.

## Test plan
- [x] `bun test src/websocket/app-server-control-reconnect.test.ts` (new): overlap + graceful reconnect both run a first-message turn
- [x] `bun test src/websocket/app-server.test.ts` (existing) green
- [x] `bun run typecheck`, `biome check`, `check-layer-boundaries`, `check-exported-functions`, `check-filename-casing`, `madge --circular` all clean
- [ ] Manual: on Desktop, reconnect (sleep/wake) then open a new chat and send a message — should respond

Note: `src/websocket/listen-client-concurrency.test.ts` has 19 pre-existing failures on `main` (reproduce with this change stashed); unrelated to this fix.

### DIFF
--- a/src/websocket/app-server-control-reconnect.test.ts
+++ b/src/websocket/app-server-control-reconnect.test.ts
@@ -1,0 +1,319 @@
+import { afterEach, describe, expect, test } from "bun:test";
+import { mkdtemp, rm } from "node:fs/promises";
+import os from "node:os";
+import { join } from "node:path";
+import WebSocket from "ws";
+import type { AgentCreateBody } from "@/backend";
+import { __testSetBackend } from "@/backend";
+import { LocalBackend } from "@/backend/local";
+import { type AppServerHandle, startAppServer } from "@/websocket/app-server";
+
+const TEST_TIMEOUT_MS = 8000;
+
+function waitForOpen(socket: WebSocket): Promise<void> {
+  if (socket.readyState === WebSocket.OPEN) {
+    return Promise.resolve();
+  }
+  return new Promise((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


sarahwooders:

PR #2894, opened 2026-06-13
fix(listener): apply the per-agent default working directory to new chats

seb-veto:
## Summary
- The desktop **"Default working directory"** setting (per agent, per device) is silently ignored by new chats: a new conversation starts in whatever directory the listener booted in (e.g. `Documents`/`Stack`) regardless of the saved default.
- Root cause: the default is stored under the `agent:<id>::conversation:default` scope key, but `getConversationWorkingDirectory` only consulted that key when the conversation id was literally `"default"` (the pre-send "New conversation" preview). Once a real conversation id is created, its scope key has no entry, so resolution fell straight back to `bootWorkingDirectory` — skipping the per-agent default entirely.

## Fix
- Add a `getInheritedWorkingDirectory` tier: a concrete conversation with no explicit per-conversation override now inherits the **per-agent default** (when set) before `bootWorkingDirectory`. The agent-default scope itself still inherits boot (no higher tier).
- Apply the same inheritance in `setConversationWorkingDirectory` when deciding whether to persist or drop an override, so explicit per-conversation choices that differ from the inherited default are preserved (and redundant entries are still pruned).

Resolution precedence is now: `conversation:<id>` override → `agent:<id>::conversation:default` → boot cwd.

## Repro (before)
1. Set a per-agent **Default working directory** (e.g. `/Users/me/Code/foo`).
2. Start a **new chat** and ask the agent what folder it is in.
3. It reports the listener boot dir (e.g. `~/Documents`), not the saved default.

## Test plan
- [x] `bun test src/websocket/listener/cwd.test.ts` (new): override / per-agent-default / boot tiers + persist-vs-drop logic
- [x] `bun test src/websocket/listen-client-protocol.test.ts` (161) green
- [x] `bun run typecheck`, `biome check`, layer-boundaries, exported-functions, filename-casing, `madge --circular` all clean
- [ ] Manual: set a default working dir, open a new chat, confirm the agent runs in that dir

### DIFF
--- a/src/websocket/listener/cwd.test.ts
+++ b/src/websocket/listener/cwd.test.ts
@@ -1,0 +1,117 @@
+import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { mkdtempSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import {
+  getConversationWorkingDirectory,
+  getWorkingDirectoryScopeKey,
+  setConversationWorkingDirectory,
+} from "./cwd";
+import type { ListenerRuntime } from "./types";
+
+const BOOT = "/boot/dir";
+const AGENT = "agent-local-test";
+const AGENT_DEFAULT = "/agent/default/dir";
+const CONV = "local-conv-abc";
+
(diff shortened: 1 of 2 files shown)


sarahwooders:

PR #2895, opened 2026-06-13
fix(local-backend): fork conversations in-process instead of via the proxy

seb-veto:
## Summary
- Forking a conversation on a **local agent** fails with `501 LOCAL_BACKEND_UNSUPPORTED_ENDPOINT` ("This local backend endpoint is not implemented in the Desktop proxy yet"), including forking the `default` conversation.
- Root cause: `LocalBackend` never overrode `forkConversation`, so it inherited the default `Backend` implementation that issues `POST /v1/conversations/{id}/fork`. The desktop local-backend proxy does not implement that route, so every local fork 501s — even though the local store already supports forking in-process.

## Fix
- Override `forkConversation` on `LocalBackend` to delegate to `this.store.forkConversation(...)` directly (mirroring the dev fake-headless backend), so it never touches the HTTP proxy. Option types already match (`{ agentId?, hidden? }`).

## Test plan
- [x] `bun test src/backend/local-backend.test.ts` (32) green, incl. new regression: a local fork returns a new **listed** conversation that carries the source summary
- [x] `bun run typecheck`, `biome check`, layer-boundaries, exported-functions, filename-casing all clean
- [ ] Manual: on a local agent, fork a conversation from the desktop sidebar — should create a forked chat instead of erroring

### DIFF
--- a/src/backend/local-backend.test.ts
+++ b/src/backend/local-backend.test.ts
@@ -233,6 +233,38 @@ describe("local backend pi transcript", () => {
     }
   });
 
+  // Regression: forking must run in-process via the local store. The default
+  // Backend.forkConversation posts to /v1/conversations/{id}/fork, which the
+  // desktop local-backend proxy returns 501 for, so without the override every
+  // fork attempt failed with LOCAL_BACKEND_UNSUPPORTED_ENDPOINT.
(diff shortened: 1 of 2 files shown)


sarahwooders:

PR #2908, opened 2026-06-15
fix(channels): normalize DM thread routing

overlord-letta:
## Summary

Fixes channel DM routing regressions from Ezra Support:

- Discord direct-message routes now send through the DM channel even when the route has no stored thread id.
- Telegram private-chat sends now drop stale forum thread ids before calling the Telegram Bot API, while group/forum topic sends keep their thread id.

## Why

Both bugs had the same shape: direct/private channel routes were carrying thread metadata in a way the provider rejected. Discord DMs need the DM channel id as the thread target in our adapter call path. Telegram private chats must not send a message_thread_id at all because Telegram only accepts it for forum topics in groups/supergroups.

## Validation

- bun test src/channels/message-channel.test.ts src/channels/telegram-adapter.test.ts
- bunx --bun @biomejs/biome@2.2.5 check src/channels/discord/message-actions.ts src/channels/message-channel.test.ts src/channels/telegram-adapter.test.ts src/channels/telegram/adapter.ts src/channels/telegram/message-actions.ts
- pre-commit hook ran layer-boundary/export/filename checks, Biome write, and tsc --noEmit

## Linear

LET-9192
LET-9188
LET-9193

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,58 +109,6 @@ jobs:
       - name: Lint & Type Check
         run: bun run check
 
-  telegram-live-smoke:
-    needs: [classify, check]
-    name: Telegram Live Smoke
-    runs-on: ubuntu-24.04
-    timeout-minutes: 10
-    if: >
-      needs.classify.outputs.channels_changed == 'true' &&
-      (github.event_name == 'push' ||
-      (github.event_name == 'pull_request' &&
-      github.actor != 'dependabot[bot]' &&
-      github.event.pull_request.head.repo.full_name == github.repository))
-    permissions:
-      contents: read
-    steps:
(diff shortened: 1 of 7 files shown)


overlord-letta:
Follow-up from internal review: tightened the same Telegram stale-thread guard for lifecycle error replies and channel control prompts, not just normal MessageChannel sends. Private Telegram chats now omit `message_thread_id` on those paths too, while group/forum topics still keep their topic thread targeting.

Additional validation:
- `bun test src/channels/message-channel.test.ts src/channels/telegram-adapter.test.ts`
- pre-commit hooks: layer-boundary/export/filename checks, Biome on staged files, `tsc --noEmit`

just-cameron (approved):

PR #2909, opened 2026-06-15
fix(transcripts): persist approval-request tool calls

overlord-letta:
## Summary

Fixes LET-9191 by writing local reflection transcript rows when a turn pauses for tool approval.

Before this, transcript deltas were appended only for normal end_turn completions. If a stream emitted approval_request_message/tool-call rows and then the stop reason was coerced to requires_approval, those approval-ready tool calls stayed visible in the live chunks/API but never reached transcript.jsonl.

## Approach

- Route transcript appends through a stop-reason-aware helper that records both end_turn and requires_approval turns.
- Keep reflection launch itself gated on normal end_turn completions.
- Merge transcript rows by source line id so the later approved tool result updates the earlier approval-ready tool-call row instead of duplicating it.

## Validation

- bun test src/cli/reflection-transcript.test.ts src/cli/app/reflection-transcript-wiring.test.ts src/websocket/listener/turn-reflection.test.ts src/headless-reminder.test.ts
- bunx --bun @biomejs/biome@2.2.5 check src/cli/helpers/reflection-transcript.ts src/cli/reflection-transcript.test.ts src/cli/app/use-conversation-loop.ts src/cli/app/reflection-transcript-wiring.test.ts src/headless.ts src/headless-reminder.test.ts src/websocket/listener/turn.ts src/websocket/listener/turn-reflection.test.ts
- pre-commit hook ran layer-boundary/export/filename checks, Biome write, and tsc --noEmit

## Linear

LET-9191

### DIFF
--- a/src/cli/app/reflection-transcript-wiring.test.ts
+++ b/src/cli/app/reflection-transcript-wiring.test.ts
@@ -15,26 +15,20 @@ describe("interactive reflection transcript wiring", () => {
     expect(source).toContain("description: AUTO_REFLECTION_DESCRIPTION");
   });
 
-  test("conversation loop evaluates reflection after the transcript append on end_turn", () => {
+  test("conversation loop evaluates reflection after the transcript append", () => {
     const conversationLoopPath = fileURLToPath(
       new URL("./use-conversation-loop.ts", import.meta.url),
     );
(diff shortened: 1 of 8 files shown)


sarahwooders:

PR #2910, opened 2026-06-15
fix(cron): broadcast crons_updated when crons.json changes on disk

christinatong01:
## Summary
- The Schedules UI only refetches on `crons_updated` events, but those were emitted **only** from the WS cron command handlers (UI-initiated add/update/delete). Crons created by the agent via `letta cron add` write directly to `~/.letta/crons.json` and never touch the listener socket, so the UI never learned about them until a manual reload.
- Add a connection-scoped watcher on `~/.letta/crons.json` that broadcasts a scope-less `crons_updated` whenever the file changes on disk. Bound to the listener **connection** (not the scheduler lease) so every connected listener notifies its own UI regardless of which process owns the scheduler — important for multi-process / cross-env setups.
- Watches the **directory** rather than the file so it survives the temp-file+rename atomic write (which swaps the inode and silently kills a file-level `fs.watch`), and debounces 250ms to coalesce the rapid events a single write produces. Bonus: the scheduler’s own run-status writes now surface live too.

## Implementation
- `src/websocket/listener/cron-file-watcher.ts` (new) — the watcher + debounced broadcast, with abort-signal teardown.
- `src/websocket/listener/lifecycle.ts` — start on connect, stop on disconnect, init in `createRuntime`.
- `src/websocket/listener/types.ts` — `cronFileWatcher` runtime field.
- `src/cron/cron-file.ts` / `src/cron/index.ts` — export `getCronDir()` + `CRON_FILE_BASENAME`.
- `src/websocket/listener/cron-file-watcher.test.ts` (new) — broadcasts on change, debounces, and stops cleanly.

## Test plan
- [x] `bun test src/websocket/listener/cron-file-watcher.test.ts` (3 pass)
- [x] `bun test src/cron/` (93 pass) and `bun test src/websocket/listener/` (74 pass)
- [x] `bun run typecheck`, `bun run lint`, `check:boundaries`, `check:cycles` all clean
- [ ] Manual: agent runs `letta cron add` → schedule appears in the desktop UI without reload


sarahwooders:

PR #2911, opened 2026-06-15
fix(local-backend): support OpenAI-compatible local provider models

overlord-letta:
## Summary
- Allow OpenAI-compatible local providers to use arbitrary model IDs after catalog and alias lookup, covering proxy providers with model handles outside the Pi catalog.
- Normalize DeepSeek built-in models to chat completions and bump `@earendil-works/pi-ai` to 0.79.4 for current provider/catalog metadata.
- Extract provider-only mod loading so local connect/startup/listener paths can register provider mods without exposing broader mod capabilities.

## Test plan
- [x] `bun run check`
- [x] `bun test src/backend/pi-model-factory.test.ts src/mods/provider-mod-adapter.test.ts src/websocket/listener/mod-adapter.test.ts src/cli/subcommands/router.test.ts`

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -5,7 +5,7 @@
     "": {
       "name": "@letta-ai/letta-code",
       "dependencies": {
-        "@earendil-works/pi-ai": "^0.79.1",
+        "@earendil-works/pi-ai": "^0.79.4",
         "@letta-ai/letta-client": "^1.10.2",
         "@pierre/diffs": "1.2.2",
         "glob": "^13.0.0",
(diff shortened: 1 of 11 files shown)


carenthomas:
I think we should close this PR for LET-9162.

The reported bug is specifically Desktop/listener not surfacing a provider mod in the local provider page. But the listener already reloads provider-capable mods/extensions on startup before handling messages, and the provider list path already reads from `listRegisteredPiProviders()`. Adding a new process-wide provider-only mod adapter and wiring it into CLI startup / `letta connect` doesn’t address the Desktop provider-list lifecycle directly.

My concern is that this broadens when trusted local mod code executes and adds early local backend reconfiguration (`configureBackendMode("local")` after provider mod load) without being necessary for the reported bug. If the issue is that the listener stays alive while a user adds/edits provider mods, the targeted fix should be a Desktop/listener refresh path: reload provider mods on explicit provider-page refresh or after a cheap source dirty check, then list providers. If the mod fails to load, surface diagnostics instead of silently omitting the provider.

The arbitrary OpenAI-compatible model handle and DeepSeek changes may be useful separately, but they don’t address the mod-loading bug reported in LET-9162.

overlord-letta:
Closing this based on review feedback. The LET-9162 fix should stay focused on the provider-extension listing/refresh path for Desktop/local provider surfaces. The arbitrary OpenAI-compatible model-handle support and DeepSeek/Pi catalog update may still be useful, but they should be split into separate work if we want them.

PR #2925, opened 2026-06-15
fix(cron): preserve channel scope for scheduled runs

overlord-letta:

## Problem
Scheduled runs are not channel-originated turns, so they have historically depended on whatever channel routes happened to exist for the conversation. The cron file also did not remember which outbound destinations were intended, and the scheduled prompt did not tell the agent how to message an allowed channel.

## What changed
- Cron tasks now persist normalized `channel_targets`.
- `letta cron add` captures the current routed channel context when invoked from a channel turn, and the WS cron protocol can also add/update explicit targets.
- The listener resolves stored targets against live routes/adapters at fire time, injects that explicit MessageChannel scope into the queued scheduled turn, and tells the agent the exact authorized `MessageChannel(...)` call shape.
- Scheduled runs with no stored targets get an explicit empty channel scope instead of falling back to all conversation routes.
- MessageChannel route resolution now honors `threadId` when one is supplied, so stored thread-scoped schedule destinations do not become ambiguous with other routes in the same chat.

### Schedule a thread-visible follow-up from Slack
If a user asks from a routed Slack thread:

```text
Remind this thread tomorrow morning to check PR 2925.
```

The agent can create the schedule from that turn:

```sh
letta cron add \
  --name pr-2925-followup \
  --description "Check PR 2925 and report back to this Slack thread" \
  --prompt "Check PR 2925. If there is a meaningful update, send it back to the originating Slack thread." \
  --cron "0 9 * * *" \
  --agent "$AGENT_ID" \
  --conversation "$CONVERSATION_ID"
```

Because the command runs inside a routed Slack turn, the persisted cron task includes the current channel destination, e.g.:

```json
{
  "channel_targets": [
    {
      "channel": "slack",
      "account_id": "5f7559d8-1a52-47a8-96dc-cfb004f80652",
      "chat_id": "C0B36B9QW7K",
      "chat_type": "channel",
      "thread_id": "1781565691.348579"
    }
  ]
}
```

When the schedule fires, the prompt shown to the agent includes the exact authorized call:

```text
Available outbound channel destinations for this scheduled run:
- MessageChannel(action="send", channel="slack", chat_id="C0B36B9QW7K", accountId="5f7559d8-1a52-47a8-96dc-cfb004f80652", threadId="1781565691.348579", message="...")
Only these destinations are authorized for MessageChannel.
```

### Create a schedule over the listener protocol
Clients that already know the intended target can pass it explicitly with `cron_add`:

```json
{
  "type": "cron_add",
  "request_id": "req-1",
  "agent_id": "agent-123",
  "conversation_id": "conv-123",
  "name": "daily-support-digest",
  "description": "Post support digest to the ops thread",
  "cron": "0 17 * * 1-5",
  "recurring": true,
  "prompt": "Summarize today's support issues and post the digest to the configured channel.",
  "channel_targets": [
    {
      "channel": "slack",
      "account_id": "acct-slack",
      "chat_id": "C123",
      "chat_type": "channel",
      "thread_id": "1712790000.000050"
    }
  ]
}
```

### Stored target is unavailable at fire time
If the stored route is disabled or the adapter is not running, the scheduled prompt says no stored destination is currently available and the run does not get MessageChannel scope for that stale target.

## Validation
- `bun run check`
- `env -u USER_CWD bun test src/cron/cron-file.test.ts src/cron/scheduler.test.ts src/cron/channel-targets.test.ts src/tools/shell-env.test.ts src/tools/tool-execution-context.test.ts src/tools/impl/message-channel.test.ts src/channels/message-channel.test.ts src/websocket/listen-client-protocol.test.ts`

## Notes
This covers listener/WS-scheduled cron fires and CLI-created schedules from routed channel turns. It does not add a new manual CLI flag for arbitrary proactive destinations.

### DIFF
--- a/src/channels/registry.ts
+++ b/src/channels/registry.ts
@@ -819,6 +819,7 @@ export class ChannelRegistry {
     agentId: string,
     conversationId: string,
     accountId?: string,
+    threadId?: string | null,
   ): ChannelRoute | null {
     const normalizedAccountId = accountId?.trim();
     const matches = getRoutesForChannel(channel).filter(
@@ -826,6 +827,9 @@ export class ChannelRegistry {
         route.chatId === chatId &&
         route.agentId === agentId &&
         route.conversationId === conversationId &&
+        (threadId === undefined
+          ? true
+          : (route.threadId ?? null) === (threadId ?? null)) &&
         (!normalizedAccountId ||
           (route.accountId ?? LEGACY_CHANNEL_ACCOUNT_ID) ===
             normalizedAccountId) &&
--- a/src/cli/subcommands/cron.ts
+++ b/src/cli/subcommands/cron.ts
@@ -25,6 +25,8 @@ import {
   parseEvery,
   readCronRunLogEntriesPage,
 } from "@/cron";
+import { extractCronChannelTargetsFromInheritedContext } from "@/cron/channel-targets";
+import { LETTA_INHERITED_CHANNEL_CONTEXT_ENV } from "@/runtime-context";
 
 // ── Usage ───────────────────────────────────────────────────────────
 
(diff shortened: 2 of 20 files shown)


just-cameron (approved):

sarahwooders:

PR #2926, opened 2026-06-16
fix(discord): keep DM replies out of message threads

ezra-letta:
## Summary
- stop using inbound message IDs as implicit thread IDs for non-Slack channel turns
- preserve Discord DM `threadId: null` so MessageChannel replies send to the DM channel instead of fetching the inbound message ID as a channel
- add a regression test for Discord DM MessageChannel replies

## Testing
- `bun test src/channels/message-channel.test.ts`
- `bun run typecheck`
- `bun run check`

## Context
Live Discord DM replies were failing with `Error sending message to discord: Unknown Channel` while inbound delivery worked. The shared MessageChannel inference path was converting the active Discord turn source message ID into a thread ID. That fallback is Slack-specific because Slack thread IDs are message timestamps; Discord message IDs are not channel/thread IDs.


sarahwooders:

PR #2931, opened 2026-06-16
feat(channels): add first-party Signal channel

overlord-letta:
## Summary

Adds a first-party Signal channel backed by `signal-cli` JSON-RPC/SSE, with runtime support, onboarding, media handling, and MessageChannel integration.

Major pieces:
- Add Signal channel plugin, REST client, SSE adapter, target parsing, message actions, and account config adapter.
- Wire Signal through channel account/config snapshots, first-party plugin discovery, routing, pairing, protocol validation, MessageChannel reminders, and formatting.
- Add interactive `letta channels configure signal` setup for local daemon probing, Docker bootstrap, native `signal-cli` link/register/verify, captcha handling, QR rendering, and already-linked account recovery.
- Support inbound images, attachments, Signal text styles, reactions, typing indicators, and voice/audio transcription with ffmpeg normalization for unsupported audio.
- Add `self_chat_mode` for personal-number / Note to Self usage, including inbound self-message handling and outbound non-self target blocking.
- Add Signal `recipient_aliases` so UUID-only inbound senders can be mapped to replyable phone-number targets.
- Guard against multiple enabled Signal accounts sharing one native daemon/base URL, since native event streams are daemon-scoped.
- Improve Signal diagnostics without logging raw SSE message bodies.
- Document setup, safety model, troubleshooting, media support, and current native daemon constraints.

## Live validation

Validated locally with native `signal-cli daemon` using:

```bash
signal-cli -c /Users/cameron/.local/share/signal-cli-letta-pr2931 daemon \
  --http 127.0.0.1:8080 \
  --receive-mode on-connection \
  --ignore-stories
```

Confirmed:
- Fresh daemon `/api/v1/check` and `/api/v1/events` work.
- JSON-RPC `send` succeeds after daemon restart.
- Inbound Signal messages route through Letta local backend.
- UUID-only inbound sender was mapped via `recipient_aliases` to `signal:+15036195666` for replyability.
- Agent successfully tested Signal text, images/media, and richer send cases through the channel.

## Test plan

Focused suite run after final review fixes:

```bash
bun test \
  src/channels/signal/setup.test.ts \
  src/channels/signal-service.test.ts \
  src/channels/signal/adapter.test.ts \
  src/channels/signal/client.test.ts \
  src/channels/signal/target.test.ts \
  src/channels/signal-registry.test.ts \
  src/channels/signal-config.test.ts \
  src/channels/signal-protocol.test.ts \
  src/channels/transcription.test.ts \
  src/channels/xml.test.ts \
  src/channels/message-channel.test.ts \
  src/tools/impl/message-channel-formatting.test.ts \
  src/tools/impl/message-channel.test.ts \
  src/channels/plugin-registry.test.ts \
  src/channels/runtime-deps.test.ts \
  src/channels/registry.test.ts
```

Result: 173 pass, 0 fail.

Additional checks:

```bash
bun run build
bunx --bun @biomejs/biome@2.2.5 check <touched files>
bun run check:exported-functions
bun run check:filename-casing
node scripts/check-layer-boundaries.js
```

All passed.

## Review notes

Pre-draft review fixes addressed:
- Dynamic Signal account starts now reject shared daemon/base URL conflicts, not just startup initialization.
- Legacy Signal config/account loading now preserves the documented default of `download_media: true` unless explicitly disabled.
- SSE handler failure diagnostics no longer include raw Signal event data.
- Native `signal-cli` config-dir detection handles quoted paths with spaces.
- Health checks fall back to JSON-RPC `version` if `/api/v1/check` is unavailable.

## Known constraints

- Letta does not install or supervise `signal-cli`; a daemon/wrapper must be running.
- Native daemon event streams are daemon-scoped. Multiple enabled Signal accounts require separate daemon instances with separate config dirs/ports.
- Some native daemon/account-state changes require restarting the daemon; stale daemon processes can receive while JSON-RPC sends fail.

> "Signals are only useful when they arrive where we meant to send them."


just-cameron:
Manual test notes from a local PR checkout (`d718d252`) while wiring up a real dedicated Signal number:

What worked:
- `bun install`, `bun run build`, and the PR's focused Signal/channel test suite passed locally.
- Native `signal-cli` registration with a dedicated number worked after upgrading Homebrew `signal-cli` from `0.13.23` to `0.14.5`.
- `signal-cli daemon --http 127.0.0.1:8080 --receive-mode on-connection --ignore-stories` exposes the endpoints this adapter expects:
  - `/api/v1/check`
  - `/api/v1/rpc`
  - `/api/v1/events`
- After account JSON was created manually, the Signal adapter started and received an inbound mobile Signal DM.

Things we learned / rough edges:
- Must use the PR checkout binary (`./letta.js ...`) for all channel commands. The globally installed `letta` does not know about `signal` yet, so `letta channels install signal` reports unknown channel.
- `./letta.js channels install signal` works, but `./letta.js channels configure signal` currently says Signal has no interactive setup. Manual `~/.letta/channels/signal/accounts.json` creation is required.
- The `bbernhard/signal-cli-rest-api` container in `MODE=json-rpc` is not compatible with this adapter as-is. Its public endpoints are REST-ish (`/v1/about`, `/v1/accounts`, etc.); it did not expose `/api/v1/check` or `/api/v1/rpc`. Native `signal-cli daemon --http` did.
- Old `signal-cli 0.13.23` failed verification with `Verify error: StatusCode: 499`; verbose logs showed `DeprecatedVersionException`. Upgrading to `0.14.5` fixed verification.
- Signal Desktop contact discovery lagged / reported the number as non-Signal even after registration; Signal mobile could find and message it. Setting `discoverableByNumber` / `numberSharing` via `updateAccount` may help, but Desktop can still be stale.

Potential bug found in local backend mode:
- With `dmPolicy: "open"`, the first inbound Signal DM reached the adapter, but auto-route creation failed:
  - `ChannelRegistry.createSignalRoute()` -> `createConversationForAgent()` -> `getClient()`
  - error: `Missing LETTA_API_KEY` / `getClient() called without credentials`
- This looks like the Signal/WhatsApp-style auto-route path is not local-backend aware. For local backend testing, we need to switch to `dmPolicy: "pairing"` or manually add a route to an existing local conversation instead of letting the first DM auto-create a cloud conversation.

Suggested follow-ups:
- Add docs or setup copy for native `signal-cli daemon --http`; clarify that `signal-cli-rest-api` is not the tested surface unless support is added.
- Consider adding `signal channels configure` support, or at least a first-party setup hint that writes the manual account JSON.
- Fix/guard local-backend auto-route conversation creation so first inbound Signal DMs do not call cloud-only `getClient()`.

src/channels/plugin-types.ts
```diff
@@ -150,8 +151,11 @@ export interface ChannelPluginAccountPatch {
   botToken?: string;
   appToken?: string;
   mode?: SlackChannelMode;
-  groupMode?: TelegramGroupMode | WhatsAppGroupMode;
+  groupMode?: TelegramGroupMode | WhatsAppGroupMode | SignalGroupMode;
   agentId?: string | null;
+  baseUrl?: string;
+  account?: string;
+  accountUuid?: string;
```
just-cameron, line 164: What are these for?

src/channels/plugin-types.ts
```diff
@@ -162,6 +166,7 @@ export interface ChannelPluginAccountPatch {
   selfChatMode?: boolean;
   allowedGroups?: string[];
   mentionPatterns?: string[];
+  recipientAliases?: Record<string, string>;
```
just-cameron, line 178: Why do we need this in the main plugin types?

src/channels/plugin-types.ts
```diff
@@ -246,7 +251,7 @@ export interface ChannelMessageActionContext {
   adapter: ChannelAdapter;
   formatText: (
     text: string,
-  ) => Pick<OutboundChannelMessage, "text" | "parseMode">;
+  ) => Pick<OutboundChannelMessage, "text" | "parseMode" | "textStyle">;
```
just-cameron, line 269: Can you explain this a little more?

just-cameron (commented):
I think I'm a little concerned about the `plugin-types.ts` file -- it seems like it may be getting too much channel-specific info and sprawl, but that may be work to defer to another time.

src/channels/plugin-types.ts
just-cameron: These are the Signal account config fields after ChannelAccountConfigAdapter maps protocol snake_case into the shared camelCase patch shape. The service currently has one create/update path for built-in channels, so these plugin-specific fields live in this flattened patch bag alongside Slack/Discord/Telegram/WhatsApp fields. I added a comment in 5203a086 calling this out and noting that a future cleanup should move this to discriminated per-channel patch types.

src/channels/plugin-types.ts
just-cameron: This is the same flattened-patch issue: recipient_aliases is Signal-specific, but after account-config normalization it becomes recipientAliases so the shared service can persist it onto SignalChannelAccount. I added a field comment in 5203a086. Agree this is not ideal long-term; I would defer the broader typed plugin patch refactor out of this already-large PR.

src/channels/plugin-types.ts
just-cameron: Added a doc comment in 5203a086. formatText is the shared MessageChannel formatting hook: the top-level tool owns markdown/plain-text normalization, then channel action adapters decide how to pass the formatted result to their concrete adapter (Telegram HTML, Slack mrkdwn, Signal textStyle ranges, etc.).

jnjpng (approved):

PR #2933, opened 2026-06-16
feat(slack): add reaction and listen mode controls

overlord-letta:
Agent: Overlord (agent-c2adbf5c-8419-4211-8cd8-3740db164974)

## Problem

Slack runs currently always add a completed ✅ reaction after removing 👀. In busy threads that creates visual clutter, while 👀 and ❌ are still useful state indicators.

There is also no read-only Slack intake mode for agents that should see thread activity without being allowed to post until they are explicitly invited.

## Approach

- Add Slack account config fields `show_completed_reaction` and `listen_mode` across config parsing, account persistence, snapshots, and protocol config payloads.
- Keep 👀 queued reactions and ❌ failure reactions unchanged, but skip the completed ✅ when `show_completed_reaction` is false.
- Add route-level `outboundEnabled` state. Slack listen mode can create a read-only route for unmentioned threaded channel activity; that route still carries lifecycle source context, but MessageChannel/proactive Slack eligibility ignores it.
- Upgrade a listen-only Slack route to outbound-enabled on an explicit mention in the same thread.

## Behavior

Default behavior is unchanged: completed turns still add ✅ and normal Slack mentions still create outbound-capable routes.

With `show_completed_reaction: false`, completion removes 👀 and does not add ✅. Failed turns still add ❌.

With `listen_mode: true`, unmentioned Slack thread replies can reach the bound agent conversation, but the route is not eligible for MessageChannel sends until someone explicitly mentions the agent in that thread.

## Validation

- `bun test src/channels/slack-adapter.test.ts src/channels/service.test.ts src/channels/slack-registry.test.ts`
- `bun test src/websocket/listen-client-protocol.test.ts`
- `bun run check`

### DIFF
--- a/src/channels/service.ts
+++ b/src/channels/service.ts
@@ -112,6 +112,8 @@ export interface ChannelConfigSnapshot {
   autoThreadOnMention?: boolean;
   threadPolicyByChannel?: Record<string, boolean>;
   acknowledgeMessageReaction?: boolean;
+  showCompletedReaction?: boolean;
+  listenMode?: boolean;
   removeStaleRoutes?: boolean;
   inboundDebounceMs?: number;
   selfChatMode?: boolean;
@@ -141,6 +143,7 @@ export interface ChannelRouteSnapshot {
   agentId: string;
   conversationId: string;
   enabled: boolean;
+  outboundEnabled?: boolean;
   createdAt: string;
   updatedAt: string;
 }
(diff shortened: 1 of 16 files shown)


src/channels/service.ts
```diff
@@ -1803,6 +1825,7 @@ export function updateChannelRouteLive(
     }),
     agentId,
     conversationId,
+    outboundEnabled: true,
```
amelia-letta: `outboundEnabled: true` hardcodes the value on every route update — if `updateChannelRouteLive` is called on a listen-only route (e.g. via `channel_route_update` WS command), it silently upgrades it to outbound-enabled. Should preserve the existing value: `outboundEnabled: existingRoute?.outboundEnabled ?? true`.

```suggestion
    outboundEnabled: existingRoute?.outboundEnabled ?? true,
```

just-cameron (approved):

kl2806 (approved):

Issue #2942, opened 2026-06-16
Tool-return image content is silently dropped to text (local backend 'Read'/'ViewImage' images never reach the model)


tanner-caffrey:
## Summary

Images returned by tools (e.g. `Read` on an image file, `ViewImage`) are silently stripped to a text placeholder before they are persisted, so the model never receives the actual image. On the local backend, reading an image yields only the `[Image: <name>]` text label and no pixels.

The root cause is a type guard, `isToolReturnContent` in `src/agent/approval-result-normalization.ts`, that validates the wrong image shape. It claims to be a `value is ToolReturnContent` guard but checks for a flat `{ type: "image", data, mimeType }` block, which is **not** a valid `ImageContent`. The canonical `ImageContent` (from `@letta-ai/letta-client`) is nested:
`{ type: "image", source: { type: "base64", data, media_type } }` — exactly what the tools emit. Because the canonical block fails the guard, `coerceToolReturnContent` falls through to `normalizeToolReturnText`, which keeps only `type: "text"` parts and discards the image.

## Affected versions

- Reproduced on released **0.27.9**.
- Still present on current `main` (verified at commit `ddeccaa3`).

## Steps to reproduce

Using the local backend, `Read` an image file:

```
Read /path/to/some-image.png
```

**Expected:** the model receives the image and can describe its contents.
**Actual:** the model receives only the text block `[Image: some-image.png]`; the image is gone. The persisted tool result contains a single `text` block and no image block.

Unit-level repro (no backend needed) — an approved tool return containing a canonical image block is flattened to text-only:

```ts
import { normalizeApprovalResultsForPersistence } from "@/agent/approval-result-normalization";

const out = normalizeApprovalResultsForPersistence([
  {
    type: "approval",
    tool_call_id: "call-img",
    approve: true,
    tool_return: [
      { type: "text", text: "[Image: chart.png]" },
      { type: "image", source: { type: "base64", media_type: "image/png", data: "aGVsbG8=" } },
    ],
  } as any,
]);

// out[0].tool_return === "[Image: chart.png]"   ← image block dropped
// expected: the array, with the image block preserved
```

## Root cause detail

The tool side is correct. `readImageFile` in `src/tools/impl/read.ts` is typed `Promise<Array<TextContent | ImageContent>>` and returns the nested canonical shape:

```ts
return [
  { type: "text", text: `[Image: ${name}]` },
  { type: "image", source: { type: "base64", media_type: result.mediaType, data: result.data } },
[...]

PR #2943, opened 2026-06-16
fix(agent): accept canonical image content in tool-return type guard

tanner-caffrey:
Ref #2942

## Summary

Tool-return images (e.g. `Read` on an image file, `ViewImage`) were silently flattened to a text placeholder before persistence, so the model never received the actual image. On the local backend this meant reading an image produced only the `[Image: <name>]` text label.

The cause is `isToolReturnContent` in `src/agent/approval-result-normalization.ts`. It is declared `value is ToolReturnContent`, but its image branch validated a flat `{ type: "image", data, mimeType }` block, which is not a valid `ImageContent`. The canonical `ImageContent` (`@letta-ai/letta-client`) is nested — `{ type: "image", source: { type: "base64", data, media_type } }` — which is exactly what the tools emit. Because the canonical block failed the guard, `coerceToolReturnContent` fell through to `normalizeToolReturnText`, which keeps only `text` parts and dropped the image.

## Changes

- **`src/agent/approval-result-normalization.ts`** — fix `isToolReturnContent`'s image branch to validate the canonical base64 `ImageContent` shape (`source.type === "base64"`, string `source.data`, string `source.media_type`) so the guard matches its declared type. Kept in the agent layer (no dependency on the local-backend `localImageContentFromLegacyImage`, to respect layer boundaries).
- **`src/agent/approval-result-normalization.test.ts`** — add regression tests for image preservation, both with and without an approval comment. (The image branch previously had no coverage, which is why this went unnoticed.)

## Why this layer / shape

`isToolReturnContent` is a guard for `ToolReturnContent` (`string | Array<TextContent | ImageContent>`), so it must validate the SDK `ImageContent` shape. The flat `{ data, mimeType }` it previously checked is the local backend's internal `LocalImageContent` storage representation (from `src/backend/local/local-store.ts`), not a `ToolReturnContent`. The strict fix is to replace the flat check with the canonical nested check rather than accept both, since the flat shape is not a valid tool return.

## Test plan

- [x] `bun test src/agent/approval-result-normalization.test.ts` — 13/13 pass.
- [x] `bunx --bun @biomejs/biome@2.2.5 check` — clean on changed files.
- [x] `tsc --noEmit` — 0 errors.
- [x] Full suite shows no failures

### DIFF
--- a/src/agent/approval-result-normalization.test.ts
+++ b/src/agent/approval-result-normalization.test.ts
@@ -190,6 +190,89 @@ describe("normalizeApprovalResultsForPersistence", () => {
       ],
     });
   });
+
+  test("preserves canonical image content in approved tool returns", () => {
+    const approvals: ApprovalResult[] = [
+      {
+        type: "approval",
+        tool_call_id: "call-img",
+        approve: true,
+        tool_return: [
+          { type: "text", text: "[Image: chart.png]" },
+          {
+            type: "image",
+            source: {
+              type: "base64",
+              media_type: "image/png",
+              data: "aGVsbG8=",
+            },
+          },
+        ],
+      } as unknown as ApprovalResult,
+    ];
+
+    const normalized = normalizeApprovalResultsForPersistence(approvals);
+
+    expect(normalized[0]).toMatchObject({
+      type: "tool",
+      tool_call_id: "call-img",
+      status: "success",
+      tool_return: [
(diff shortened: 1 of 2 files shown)


cpacker (approved):
Approved. The fix is at the correct boundary: approval normalization should preserve canonical SDK image tool_return content so local persistence can convert it to Pi’s flat image format downstream.

PR #2951, opened 2026-06-17
feat(settings): simplify agent pins

devanshrj:
## Summary
- make agents[].pinned the single source of truth for agent pins per backend namespace
- remove project-scoped agent pins and local/global agent pin UI distinctions
- strip legacy pinnedAgents/pinnedAgentsByServer as obsolete settings keys without migrating them
- keep conversation pin scoping unchanged

## Tests
- bun run lint
- bun run typecheck
- bun test src/settings-manager.test.ts src/agent/resolve-startup-agent.test.ts src/agent/resolve-startup-agent-files.test.ts
- bun test src/settings-manager.test.ts src/agent/resolve-startup-agent.test.ts src/agent/resolve-startup-agent-files.test.ts src/websocket/listen-client-protocol.test.ts


src/cli/app/submit-profile-commands.ts
```diff
@@ -254,32 +171,14 @@ export async function handleProfileCommand(
     const cmd = commandRunner.start(trimmed, "Pinning agent...");
     setActiveProfileCommandId(cmd.id);
     try {
-      await handlePin(profileCtx, msg, currentArgs);
+      await handlePin(profileCtx, msg, argsStr);
```
amelia-letta, line 174: `/pin agent foo` now renames the agent to "agent foo" instead of "foo" — the old code stripped the "agent" prefix via `target === "agent" ? parts.slice(1).join(" ")`. Same issue with `/pin help` (renames to "help" instead of showing help) and `/pin convo` (renames to "convo" instead of erroring on removed conversation pins).

PR #2954, opened 2026-06-17
fix(channels): add reconnecting feedback and queue preservation for env readiness

overlord-letta:
## Summary

When channel messages arrive during listener reconnection or startup, messages now get immediate user feedback instead of buffering silently. The buffer also has TTL (5min) and size limits (100) with drop notifications.

On WebSocket disconnect, queued items are now preserved instead of cleared, and replayed on reconnect to prevent message loss.

## Changes

1. **Reconnecting notification**: Added `buildChannelReconnectingMessage()` that sends immediate feedback to users when their message is buffered during reconnection. This follows the same pattern as the approval path (line 1036-1044 in the original code).

2. **Buffer TTL and size limits**: 
   - Added `BUFFER_TTL_MS` (5 minutes) and `BUFFER_MAX_SIZE` (100) constants
   - Messages older than TTL are dropped with `buildChannelBufferExpiredMessage()` notification
   - When buffer exceeds max size, oldest messages are dropped with notification

3. **Queue preservation on disconnect**:
   - Modified the socket close handler to only clear queues for intentional shutdowns
   - Non-intentional disconnects now preserve queued items for replay
   - Added `recoverPendingQueuedMessages()` to schedule queue pumps for preserved items

4. **Tests**: Added tests for reconnecting notification and buffer flush behavior

## Why this matters

Previously, users sending messages during reconnection saw:
- No feedback that their message was buffered
- Error messages about "environment not ready"
- Lost messages when the queue was cleared on disconnect

Now they get:
- Immediate "reconnecting" notification with queue confirmation
- Messages preserved across reconnect cycles
- Clear feedback if messages expire (TTL) or are dropped (buffer full)

## Validation

- `bun run check` - all 9 checks pass
- `bun test src/channels/` - 540 tests pass
- `bun test src/websocket/listen-client-concurrency.test.ts` - 34 tests pass
- `bun run lint` - no issues

## Known risks

- Queue preservation assumes the agent/conversation still exists after reconnect. If the agent was deleted during disconnect, the replayed messages will fail gracefully (the agent API will return a 404 and the turn will error out).
- Buffer notifications are best-effort (fire-and-forget) to avoid disrupting the inbound pipeline


src/channels/registry.ts
```diff
@@ -355,6 +358,24 @@ export interface ChannelInboundDelivery {
...
+interface BufferedDelivery {
+  delivery: ChannelInboundDelivery;
+  /** Epoch ms when this delivery was buffered. */
+  bufferedAt: number;
+  /** Channel + chat metadata for reconnecting/expired notifications. */
+  channelId: string;
+  chatId: string;
+  threadId?: string | null;
+  messageId?: string;
+}
```
amelia-letta: `notifyBufferDrop` hardcodes `LEGACY_CHANNEL_ACCOUNT_ID` for the adapter lookup, but `notifyReconnecting` uses `msg.accountId ?? LEGACY_CHANNEL_ACCOUNT_ID` — so drop notifications silently fail for non-legacy accounts (adapter not found). Add `accountId` to `BufferedDelivery` and use it in `notifyBufferDrop`.

```suggestion
interface BufferedDelivery {
  delivery: ChannelInboundDelivery;
  /** Epoch ms when this delivery was buffered. */
  bufferedAt: number;
  /** Channel + chat metadata for reconnecting/expired notifications. */
  channelId: string;
  accountId: string;
  chatId: string;
  threadId?: string | null;
  messageId?: string;
}
```

src/channels/registry.ts
```diff
@@ -2023,22 +2059,106 @@ export class ChannelRegistry {
...
+
+  /**
+   * Notify the user that a buffered delivery was dropped (TTL expired or
+   * max buffer size reached). Fire-and-forget.
+   */
+  private notifyBufferDrop(
+    item: BufferedDelivery,
+    reason: "ttl_expired" | "max_size",
+  ): void {
+    const accountId = LEGACY_CHANNEL_ACCOUNT_ID;
```
amelia-letta: Once `accountId` is on `BufferedDelivery`, this should use `item.accountId` instead of the hardcoded constant, matching the pattern in `notifyReconnecting`.

```suggestion
    const accountId = item.accountId;
```

PR #2958, opened 2026-06-17
fix(listener): hide all subagents from the desktop sidebar

carenthomas:
hide all subagents from the desktop sidebar snapshot so the ui only shows the main agent.\n\nvalidated with bun test src/websocket/listen-client-protocol.test.ts and bun run check.


carenthomas:
closing this pr because the sidebar filtering belongs in letta-cloud, not letta-code

PR #2959, opened 2026-06-17
feat(mods): add provider_error event and provider credential updates

sarahwooders:
## Summary

Adds two new mod API capabilities that together enable mods to handle provider errors and rotate credentials — the foundation for patterns like multi-account ChatGPT rate-limit routing.

**`provider_error` event**: Fires when the LLM provider returns an error during a turn, before the built-in retry logic runs. Mods can return `{ action: "retry" }` to request an immediate retry, or `{ action: "continue" }` to let the default recovery handle it.

**`letta.providers.updateCredentials(providerName, credentials)`**: Updates OAuth tokens or API keys on the active backend before requesting a retry. Supports `credentials.oauth` (for ChatGPT OAuth providers) and `credentials.apiKey` (for API-key providers).

Together these enable a mod to:
1. Catch a 429 from `chatgpt_oauth`
2. Rotate to the next ChatGPT account via `updateCredentials`
3. Return `{ action: "retry" }` to retry the turn with the new credentials

## Implementation details

- `provider_error` is wired into both the post-stop error path (`turn.ts`) and the pre-stream error path (`send.ts`), both firing before the existing retry decision
- If any mod handler returns `{ action: "retry" }`, the harness retries immediately without running the default retry policy; otherwise the default logic runs as normal
- `updateCredentials` routes through a new `provider-credentials.ts` helper that writes to the local provider auth store (same store used by `/connect`)
- `getBackend` is now passed through `loadLocalMods` → `createLettaModApi` so the credentials helper can access the active backend
- The `providerError` capability is gated independently from other event capabilities — the listener (provider-only context) enables it but not lifecycle/tools/turns events

## What a mod looks like

```ts
export default function activate(letta) {
  if (!letta.capabilities.events.providerError) return;

  return letta.events.on("provider_error", (event) => {
    if (event.status !== 429) return;
    if (event.providerType !== "chatgpt_oauth") return;

    const next = rotateToNextAccount();
    letta.providers.updateCredentials("chatgpt-plus-pro", {
      oauth: next,
    });

    return { action: "retry" };
  });
}
```

See [trace](https://app.letta.com/chat/agent-4140c4b5-f32e-475c-9573-bcfd9446ca7b?conversation=conv-b7e893f9-60ba-4e2b-bb9d-009acbc30af3). Resume conversation with `letta --conv conv-b7e893f9-60ba-4e2b-bb9d-009acbc30af3`

Co-Authored-By: Letta Code <hi@letta.com>
Co-Authored-By: Bob <agent-4140c4b5-f32e-475c-9573-bcfd9446ca7b@letta.com>
Co-Authored-By: Sarah Wooders <sarahwooders@gmail.com>


src/websocket/listener/turn.ts
```diff
@@ -989,6 +993,113 @@ export async function handleIncomingMessage(
...
+          const result = await emitModEvent(
+            modAdapter.events,
+            "provider_error",
+            {
+              agentId: agentId ?? null,
+              conversationId,
+              status: undefined,
+              detail: errorDetail,
+              providerType,
+              providerName: providerType,
```
amelia-letta, line 1027: `providerName` is set to `model_endpoint_type` — that's the provider *type* (e.g. `"chatgpt_oauth"`), not the provider *name* (e.g. `"chatgpt-plus-pro"`). The PR example uses `updateCredentials("chatgpt-plus-pro", ...)` which is the name, but `event.providerName` would be `"chatgpt_oauth"` here. Mod authors will pass the wrong value to `updateCredentials`.

src/mods/mod-engine.ts
```diff
@@ -1119,6 +1129,18 @@ function createLettaModApi(
         return registerProviderForOwner(name, config);
       },
       unregister: unregisterProvider,
+      async updateCredentials(providerName, credentials) {
```
amelia-letta, line 1132: `registerProviderForOwner` checks both `guardLive` and `capabilities.providers`, but `updateCredentials` checks neither. After `/reload`, a stale mod handler could still mutate provider credentials.

src/mods/provider-credentials.ts
```diff
@@ -0,0 +1,47 @@
+import type { Backend } from "@/backend/backend";
+import type { ModProviderCredentials } from "@/mods/types";
+
+/**
+ * Update provider credentials on the active backend.
+ *
+ * For local backends, writes directly to the local provider auth store.
+ * For API backends, updates the provider through the BYOK provider API.
```
amelia-letta, line 8: Docstring says "For API backends, updates the provider through the BYOK provider API" but `_backend` is unused and there's no API backend path. A mod calling `updateCredentials` on an API backend gets no error and no effect — silent no-op that could cause infinite retry loops.

PR #2961, opened 2026-06-17
fix: avoid stale MemFS env for shell tools

overlord-letta:
## Summary

Windows shell tools could inherit a stale `MEMORY_DIR` from the parent process when MemFS settings were unavailable. In the reported Desktop/cloud-agent case, that value resolved to the Windows user profile directory, so the agent treated `C:\Users\...` as MemFS even though the agent had no memory filesystem.

This PR changes shell env construction to fail closed for MemFS: if MemFS is disabled or cannot be resolved, `MEMORY_DIR`/`LETTA_MEMORY_DIR` are removed unless they are the expected inherited parent-agent memory directory. It also exposes cached secure API tokens to shell tools so Desktop gateway calls can use the token already loaded by the app, and updates image-generation guidance for PowerShell/profile-image placement.

## Validation

- `bun test src/tools/shell-env.test.ts src/tools/bash-windows-description.test.ts src/cli/helpers/session-context-windows.test.ts src/skills/builtin/image-generation/image-generation.test.ts`
- `bun run check`


sarahwooders:

PR #2966, opened 2026-06-17
test(local): preserve pi-ai image downgrade

overlord-letta:
## Summary
- Remove the speculative pre-dispatch rejection that bypassed Pi-AI normalization.
- Add a regression around the OpenAI-compatible local payload path: text-only local models get Pi-AI text placeholders for image content, not raw image payloads.
- This does not claim to fix the reported crash; it protects the known downgrade behavior while that report remains unverified.

## Test plan
- [x] bun test src/backend/pi-stream-adapter.test.ts
- [x] bunx --bun @biomejs/biome@2.2.5 check --config-path biome.json src/backend/pi-stream-adapter.test.ts
- [x] commit hooks, including tsc --noEmit
- [ ] GitHub CI for latest commit

### DIFF
--- a/src/backend/dev/pi-stream-adapter.ts
+++ b/src/backend/dev/pi-stream-adapter.ts
@@ -282,46 +282,6 @@ function toPiMessages(messages: readonly LocalMessage[]): Message[] {
   return normalized;
 }
 
-function contextHasImageInput(context: Context): boolean {
-  return context.messages.some(
-    (message) =>
-      Array.isArray(message.content) &&
-      message.content.some((part) => isRecord(part) && part.type === "image"),
-  );
-}
-
-function modelInputTypes(model: Model<string>): string[] {
-  const input = (model as { input?: readonly unknown[] }).input;
-  if (!Array.isArray(input)) return [];
-  return input.filter((value): value is string => typeof value === "string");
-}
-
-function modelLabel(model: Model<string>): string {
-  const provider =
-    typeof model.provider === "string" && model.provider.length > 0
-      ? `${model.provider}/`
-      : "";
-  return `${provider}${model.id}`;
-}
-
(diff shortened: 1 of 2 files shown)


cpacker (commented):
I don't think this is the right boundary for this fix.

The local backend hands Pi-AI a Pi-shaped `Context` containing user/tool-result content. Pi-AI already has model-capability-aware downgrade logic in `transformMessages(...)`: for models whose `model.input` does not include `"image"`, it replaces image blocks with text placeholders before building the provider payload, e.g.:

- user images: `(image omitted: model does not support images)`
- tool-result images: `(tool image omitted: model does not support images)`

DeepSeek-style Pi models are text-only (`input: ["text"]`), so the expected behavior is graceful degradation, not a hard pre-dispatch failure. I also verified the actual Pi payload path with an OpenRouter DeepSeek model: image blocks were removed and the provider payload contained only text placeholders.

This PR short-circuits that behavior in `PiStreamAdapter` before Pi-AI can normalize the context. That means conversations with historical image/tool-result content would now error out when using a text-only local model, even though Pi-AI could have continued safely with placeholders. In practice that is a UX regression and does not prove or fix the suspected SIGSEGV path.

If there is a crash, I think the better follow-up is to identify where images are actually reaching native/provider code despite text-only model metadata. Possible fixes/tests:

1. fix incorrect model/provider metadata if a text-only model is being marked image-capable,
2. add a regression test around the real default Pi stream/payload path proving text-only DeepSeek receives no `image_url`, `input_image`, `inlineData`, etc.,
3. investigate native image processing / resize / renderer / memory-pressure crash paths separately.

So I would not merge this as-is. The guard should either be removed, or replaced by a test/fix that preserves Pi-AI's placeholder downgrade behavior while ensuring the final provider payload has no image parts for text-only models.

PR #2968, opened 2026-06-17
fix(slack): rehydrate bot-authored thread context

overlord-letta:
## Summary
- Keep live Slack `bot_message` events ignored as runnable input, so bot activity does not wake agents.
- Rehydrate Slack thread history on later routed human turns and include only bot-authored entries from that skipped history in `threadContext`.
- Add regression coverage for bot-authored thread/channel history hydration, the later routed-turn path, and the non-runnable live bot event guard.

## Behavior
Before, Slack API history hydration happened only on the first routed turn. If another Slack bot replied after the route already existed, that bot reply was intentionally ignored as live input and was also absent from the next human turn.

After, the next human turn in an existing routed Slack thread fetches recent thread history and carries bot-authored entries as context. Human history is not repeated on later routed turns because those messages already enter the conversation normally.

## Validation
- `bun test src/channels/slack-adapter.test.ts src/channels/slack-media.test.ts`
- `bun run check`

### DIFF
--- a/src/channels/slack-adapter.test.ts
+++ b/src/channels/slack-adapter.test.ts
@@ -658,6 +658,66 @@ test("slack adapter hydrates prior Slack thread context, including bot-authored 
   expect(resolveSlackThreadHistoryMock).toHaveBeenCalledTimes(1);
 });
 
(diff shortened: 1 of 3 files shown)


just-cameron (approved):
Smoke tested locally, works for me.

PR #2971, opened 2026-06-18
feat: support ChatGPT OAuth provider aliases

jnjpng:
## Summary
- add `--name` support for ChatGPT OAuth connection flows so users can save multiple accounts under custom provider aliases
- preserve `chatgpt_oauth` metadata for aliased model handles when updating agent/conversation model settings
- route custom ChatGPT OAuth aliases to Codex/OpenAI-style tooling and update related tests

## Companion PR
- letta-cloud: https://github.com/letta-ai/letta-cloud/pull/12469

## Tests
- `bun run check`
- `bun test src/cli/connect-oauth-core.test.ts src/cli/subcommands/connect.test.ts src/tools/model-provider-detection.test.ts`

### DIFF
--- a/src/agent/modify.ts
+++ b/src/agent/modify.ts
@@ -280,43 +280,6 @@ function maxTokensForUpdatePayload(
     : undefined;
 }
 
-async function providerTypeForModelHandle(
-  backend: ReturnType<typeof getBackend>,
-  modelHandle: string,
-): Promise<string | undefined> {
-  try {
-    const models = await backend.listModels();
-    const match = models.find(
-      (model: { handle?: string | null }) => model.handle === modelHandle,
-    ) as { provider_type?: string | null } | undefined;
-    return typeof match?.provider_type === "string"
-      ? match.provider_type
-      : undefined;
-  } catch {
-    return undefined;
-  }
-}
-
-async function enrichUpdateArgsWithProviderType(
-  backend: ReturnType<typeof getBackend>,
-  modelHandle: string,
-  updateArgs: Record<string, unknown> | undefined,
-): Promise<Record<string, unknown> | undefined> {
-  if (typeof updateArgs?.provider_type === "string") {
-    return updateArgs;
-  }
-
-  const providerType = await providerTypeForModelHandle(backend, modelHandle);
-  if (!providerType) {
-    return updateArgs;
-  }
-
-  return {
-    ...(updateArgs ?? {}),
-    provider_type: providerType,
-  };
-}
-
 /**
  * Updates an agent's model and model settings.
  *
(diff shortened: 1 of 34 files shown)


src/cli/app/model-config.ts
```diff
@@ -113,6 +113,7 @@ export function getPreferredAgentModelHandle(
 
 export function mapHandleToLlmConfigPatch(
   modelHandle: string,
+  providerType?: string | null,
```
amelia-letta, line 142: Three callers of `mapHandleToLlmConfigPatch` don't pass the new `providerType` param: `AppCoordinator.tsx:3406`, `use-reasoning-cycle.ts:266`, `use-configuration-handlers.ts:521`. For custom alias handles like `chatgpt-work/gpt-5.5`, the function falls through to `endpointType = null` and returns `{ model: "chatgpt-work/gpt-5.5" }` — the full handle as model name, no endpoint type. The `llm_config` patch needs `model: "gpt-5.5", model_endpoint_type: "chatgpt_oauth"` instead.

src/agent/modify.ts
```diff
@@ -263,6 +280,43 @@ function maxTokensForUpdatePayload(
     : undefined;
 }
 
+async function providerTypeForModelHandle(
+  backend: ReturnType<typeof getBackend>,
+  modelHandle: string,
+): Promise<string | undefined> {
+  try {
+    const models = await backend.listModels();
```
amelia-letta: `backend.listModels()` on every model update is a redundant blocking API call. The available-models cache already fetches this data — could extend it to include `provider_type` and reuse it here instead of a fresh network round-trip.

src/cli/app/model-config.ts
jnjpng: Thanks!

src/agent/modify.ts
jnjpng: Thanks!

src/cli/app/model-config.ts
jnjpng: Fixed in 064f485b. Provider type now comes from the existing available-models/model-selection metadata and is passed into all `mapHandleToLlmConfigPatch` callsites, so an aliased handle like `chatgpt-personal/gpt-5.5` maps to `model: "gpt-5.5"` + `model_endpoint_type: "chatgpt_oauth"` instead of treating the full handle as the model name.

src/agent/modify.ts
jnjpng: Fixed in 064f485b. Removed the `backend.listModels()` call from `modify.ts`; the available-models cache now stores provider_type metadata by handle, and ModelSelector / `/model` selection carries it through `updateArgs.provider_type` into the update helpers.

jnjpng:
@amelia-letta do another review? looks good?

carenthomas (approved):

PR #2978, opened 2026-06-18
fix(app-server): strip model_settings from conversation create body on cloud

carenthomas:
## summary
- Desktop v0.27.9 New Conversation always 500s via the app-server conversation create paths
- Root cause: `letta-client` SDK advertises `model_settings` on `ConversationCreateParams` but the cloud `POST /v1/conversations/` endpoint rejects it (returns 500, not 422, hiding the cause)
- Both app-server paths added in #2828/#2832 forward the body directly without stripping `model_settings`
- Fix: `sanitizeConversationCreateBody()` strips `model_settings` before forwarding on the cloud path; local backend preserves it (used for per-conversation model overrides)

## affected paths
- `conversation_create` WS command (`agents-conversations.ts`)
- `runtime_start` with `create_conversation` body (`runtime-start.ts`)

## test plan
- [ ] New Conversation button in Desktop no longer 500s
- [ ] Fork still works (unaffected path)
- [ ] Local backend conversation create with `model_settings` still works
- [ ] `bun test src/websocket/listener/commands/conversation-body.test.ts`
- [ ] `bun test src/websocket/listen-client-protocol.test.ts`

### DIFF
--- a/src/websocket/listener/commands/agents-conversations.ts
+++ b/src/websocket/listener/commands/agents-conversations.ts
@@ -30,6 +30,7 @@ import {
   isConversationRetrieveCommand,
   isConversationUpdateCommand,
 } from "@/websocket/listener/protocol-inbound";
+import { sanitizeConversationCreateBody } from "./conversation-body";
 import type { RunDetachedListenerTask, SafeSocketSend } from "./types";
 
 export type AgentConversationManagementCommand =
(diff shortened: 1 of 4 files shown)


carenthomas:
closing in favor of a root-cause fix in letta-cloud — the 500 is caused by a missing zai_coding discriminator in ModelSettingsUnion on the server, not something letta-code should paper over by stripping the field

PR #2985, opened 2026-06-18
fix(model): preserve settings when switching models

4shub:
Merge selected model defaults over existing model settings so changing models does not reset LLM configuration values like temperature, token limits, and reasoning settings.

### DIFF
--- a/src/agent/modify.ts
+++ b/src/agent/modify.ts
@@ -263,6 +263,20 @@ function maxTokensForUpdatePayload(
     : undefined;
 }
 
+function mergeModelSettings(
+  existingSettings: unknown,
+  selectedModelSettings: ModelSettings,
+): ModelSettings {
+  if (!isRecord(existingSettings)) {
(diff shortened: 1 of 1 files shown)


src/agent/modify.ts
```diff
@@ -263,6 +263,20 @@ function maxTokensForUpdatePayload(
...
 
+function mergeModelSettings(
+  existingSettings: unknown,
+  selectedModelSettings: ModelSettings,
+): ModelSettings {
+  if (!isRecord(existingSettings)) {
+    return selectedModelSettings;
+  }
+
+  return {
```
amelia-letta, line 274: cross-provider switches will leak stale provider-specific fields into the new model_settings. e.g. switching anthropic→openai preserves `thinking` (anthropic-only) in the merged payload. could cause backend validation issues depending on how strict the server is about unknown fields for the target provider_type.

PR #2998, opened 2026-06-19
feat(listener): enable turn_start event for listener mods

cpacker:
## Summary

This PR enables `turn_start` event handlers for listener mods (Desktop, Slack, channels), allowing mods to intercept and transform user input before it reaches the agent.

## Context for Caren

Your PR #2949 wired `modEvents` through the listener at all the right callsites, but left `events.turns: false` in `LISTENER_MOD_CAPABILITIES`. This PR flips that flag and adds the actual `turn_start` emission.

## Changes

**`mod-adapter.ts`** (1 line)
```diff
events: {
  lifecycle: false,
  tools: false,
- turns: false,
+ turns: true,
},
```

**`turn.ts`** (~50 lines)
- Add `isTurnInputArray()` type guard (same as headless.ts)
- Add `emitListenerTurnStart()` helper that:
  - Creates a `ModContext` via `createListenerModContext()`
  - Emits `turn_start` with `{ agentId, conversationId, input }`
  - Returns potentially modified input (or original if handler throws)
- Call `emitListenerTurnStart()` right after `currentInput = messagesToSend`

## Use Case

This unblocks mods that need to intercept user messages for listener-served agents (Slack, Desktop API agents). Example: a diagnostic mode mod that detects trigger phrases and injects system reminders.

## Test

- `bun run check` passes (ignoring unrelated filename casing issue)
- Manual testing needed for actual turn_start handler behavior

### DIFF
--- a/src/websocket/listener/mod-adapter.ts
+++ b/src/websocket/listener/mod-adapter.ts
@@ -12,7 +12,7 @@ export const LISTENER_MOD_CAPABILITIES: ModCapabilities = {
   events: {
     lifecycle: false,
     tools: false,
-    turns: false,
+    turns: true,
   },
   permissions: false,
   providers: true,
--- a/src/websocket/listener/turn.ts
+++ b/src/websocket/listener/turn.ts
@@ -229,6 +229,7 @@ async function emitListenerTurnStart(options: {
   input: Array<MessageCreate | ApprovalCreate>;
   runtime: ListenerRuntime;
   workingDirectory: string;
+  permissionMode?: string | null;
   cachedAgent?: AgentState | null;
 }): Promise<Array<MessageCreate | ApprovalCreate>> {
   try {
@@ -236,6 +237,7 @@ async function emitListenerTurnStart(options: {
     const context = createListenerModContext({
       sessionId: options.conversationId,
       workingDirectory: options.workingDirectory,
+      permissionMode: options.permissionMode ?? null,
       agent: options.cachedAgent ?? null,
     });
     const event = {
@@ -492,7 +494,9 @@ export async function handleIncomingMessage(
           : m,
       ),
     );
-    const inboundUserTranscriptLines =
+    // Build transcript lines after turn_start so transformed input is shown.
+    // This is reassigned below after emitListenerTurnStart.
+    let inboundUserTranscriptLines =
       buildInboundUserTranscriptLines(messagesToSend);
 
     const firstMessage = normalizedMessages[0];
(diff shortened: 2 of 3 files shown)


overlord-letta:
## Review

Approach is sound — mirrors the headless and TUI `turn_start` patterns exactly (same `isTurnInputArray` guard, same try/catch fallback, same emit-before-tool-prep ordering). This is the natural follow-up to #2949 which wired the callsites but left `turns: false`.

### Notes

1. **No tests** — `emitListenerTurnStart` has no coverage. The existing `mod-adapter.test.ts` tests tool execution context but not event emission. Worth at least testing the "handler throws → original input returned" and "handler modifies input → modified input returned" paths.

2. **`isTurnInputArray` is now triple-defined** — in `headless.ts`, `use-conversation-loop.ts`, and now `turn.ts`. Could extract to a shared util. Not a blocker.

3. **`cachedAgent` type compat is fine** — `AgentState` from the letta-client has all the fields `createListenerModContext` expects (`id`, `name`, `model`, `llm_config`).

4. **Placement is correct** — `turn_start` fires before `prepareToolExecutionContextForScope`, matching headless/TUI ordering.

Overall looks good — the test gap is the main thing worth addressing before merge.

src/websocket/listener/mod-adapter.ts
```diff
@@ -12,7 +12,7 @@ export const LISTENER_MOD_CAPABILITIES: ModCapabilities = {
   events: {
     lifecycle: false,
     tools: false,
-    turns: false,
+    turns: true,
```
carenthomas, line 15: This flips the capability, but `src/websocket/listener/mod-adapter.test.ts` still asserts `turns: false`, so the listener mod-adapter test fails. Please update that assertion and add coverage that a listener `turn_start` handler is actually invoked and can mutate input.

src/websocket/listener/turn.ts
```diff
@@ -210,6 +214,43 @@ function escapeTaskNotificationSummary(summary: string): string {
...
+  agentId: string;
+  conversationId: string;
+  input: Array<MessageCreate | ApprovalCreate>;
+  runtime: ListenerRuntime;
+  workingDirectory: string;
+  cachedAgent?: AgentState | null;
+}): Promise<Array<MessageCreate | ApprovalCreate>> {
+  try {
+    const modAdapter = ensureListenerModAdapter(options.runtime);
+    const context = createListenerModContext({
```
carenthomas, line 237: This context is missing the current permission mode, so Desktop/listener `turn_start` handlers see `ctx.permissionMode === null` even when the session is in standard/plan/bypass. I’d pass `turnPermissionModeState.mode` through here so listener events match the scoped context shape we already give listener tools.

src/websocket/listener/turn.ts
```diff
@@ -543,7 +584,14 @@ export async function handleIncomingMessage(
       }
     }
 
-    let currentInput = messagesToSend;
+    let currentInput = await emitListenerTurnStart({
```
carenthomas: This fires for approval-only continuations too, so a mod can rewrite the approval payload and make `isApprovalOnlyInput(currentInput)` route the turn through the normal message path instead of `sendApprovalContinuationWithRetry`. I’d only emit `turn_start` when the input contains a user message, matching the CLI guard around turn-start events.

src/websocket/listener/turn.ts
```diff
@@ -543,7 +584,14 @@ export async function handleIncomingMessage(
...
+      agentId,
+      conversationId,
+      input: messagesToSend,
```
carenthomas: `inboundUserTranscriptLines` was already built from `messagesToSend` before this mutation, and the accumulator preserves the optimistic text when the backend echoes the same OTID. If a handler transforms the user message, Desktop can keep showing the pre-transform text; I’d build the transcript seed from the final `currentInput` after `turn_start`.

carenthomas (commented):
reviewed with listener/Desktop flow in mind. I found a few blockers I would fix before merging.

PR #3006, opened 2026-06-19
fix(app-server): Fix app-server runTurn approval terminality

cpacker:
## Summary

This fixes an app-server client terminality bug that we uncovered while working on the `letta-code-sdk` remote/app-server PR (`LET-9237`).

The SDK live integration tests were running turns with:

- SDK `permissionMode: "bypassPermissions"`
- app-server runtime mode `"unrestricted"`

In that mode, complete tool approvals should be auto-allowed by the app-server listener, executed client-side, and then continued to the final model stop reason. Instead, the SDK sometimes saw a terminal SDK result with:

```ts
stopReason: "requires_approval"
approvalConflict: true
```

That looked at first like an SDK mapping issue, but tracing the live stream through `letta-code` showed the SDK was faithfully reflecting a premature result from the exported app-server client helper.

## Root cause

The listener path treats `requires_approval` as an intermediate backend/model boundary, not necessarily as a terminal turn state:

1. The backend stream emits `stop_reason: "requires_approval"`.
2. `src/websocket/listener/turn.ts` routes that through `handleApprovalStop(...)` rather than ending the turn.
3. Under scoped permission mode `"unrestricted"`, `src/websocket/listener/turn-approval.ts` auto-allows complete approvals, executes the tool batch, emits tool returns, sends the approval continuation, and continues to the final stop reason.

However, exported `AppServerClient.runTurn()` resolved on **any** stream delta whose `message_type` was `"stop_reason"`, including the intermediate `requires_approval` stop. That made SDK consumers believe the turn had ended before the listener had a chance to execute and continue the auto-approved tool call.

So the bug is in the upstream app-server client helper's definition of terminality, not in the SDK:

> Raw `stop_reason: "requires_approval"` is not terminal by itself. It is terminal only if the listener loop actually reaches `WAITING_ON_APPROVAL`.

## Fix

- `runTurn()` now waits through raw/intermediate `stop_reason: "requires_approval"`.
- It terminalizes approval conflict only when an `update_loop_status` frame reports `WAITING_ON_APPROVAL` for the same runtime/turn evidence.
- Existing final stop reasons like `end_turn` still resolve via `completedBy: "stop_reason"`.
- The guarded `WAITING_ON_INPUT` fallback behavior is preserved.

## Regression coverage

Added app-server client tests for both sides of the distinction:

1. Auto-handled approval continuation:
   - approval request
   - raw `stop_reason: "requires_approval"`
   - listener status moves to tool execution
   - later final `stop_reason: "end_turn"`
   - `runTurn()` resolves success on `end_turn`, not on `requires_approval`

2. Genuine pending approval:
   - approval request
   - raw `stop_reason: "requires_approval"`
   - listener status becomes `WAITING_ON_APPROVAL`
   - `runTurn()` resolves with `stopReason: "requires_approval"` and `completedBy: "loop_status_waiting_on_approval"`

## SDK validation

After making this upstream fix, I removed the temporary SDK-side workaround and linked the SDK worktree to this locally patched `letta-code` package:

```text
node_modules/@letta-ai/letta-code -> /Users/loaner/dev/letta-code-prod/.letta/worktrees/fix-app-server-requires-approval-terminality
```

Then I reran the stricter SDK tests that no longer tolerate approval conflicts under bypass/unrestricted.

SDK validation against this patched package:

- `bun run check`
- `bun test src/tests/client.test.ts`
- `bun test`
- `LETTA_LIVE_INTEGRATION=1 LETTA_API_KEY=set LETTA_BASE_URL=http://localhost:60736 LETTA_AGENT_ID=agent-6b383e6f-f2df-43ed-ad88-8c832f1129d0 bun test src/tests/live.integration.test.ts`

The live SDK suite passed with the stricter assertions:

```text
8 pass
0 fail
```

## Testing

Upstream tests/checks:

- `bun test src/app-server-client.test.ts`
- `bun run typecheck`
- `bun run check`

### DIFF
--- a/src/app-server-client.test.ts
+++ b/src/app-server-client.test.ts
@@ -318,6 +318,153 @@ describe("app-server client", () => {
     });
   });
 
+  test("runTurn waits through intermediate requires_approval stop_reason", async () => {
+    const { client, control, stream } = createFakeClient();
+    control.open();
+    stream.open();
+    await client.connect();
+
+    const runtime = { agent_id: "agent-1", conversation_id: "conv-1" };
+    const turn = client.runTurn({
+      runtime,
+      payload: {
+        kind: "create_message",
+        messages: [{ role: "user", content: "use a tool" }],
+      },
+    });
+
+    stream.receive({
+      type: "stream_delta",
+      runtime,
+      event_seq: 1,
+      emitted_at: "2026-06-11T00:00:00.000Z",
+      idempotency_key: "stream-1",
+      delta: {
+        type: "message",
+        message_type: "approval_request_message",
+        run_id: "run-approval",
+        tool_calls: [
+          {
+            tool_call_id: "call-1",
+            name: "Bash",
+            arguments: '{"command":"pwd"}',
+          },
+        ],
+      },
+    });
+    stream.receive({
+      type: "stream_delta",
+      runtime,
+      event_seq: 2,
+      emitted_at: "2026-06-11T00:00:00.001Z",
+      idempotency_key: "stream-2",
+      delta: {
+        type: "message",
+        message_type: "stop_reason",
+        run_id: "run-approval",
+        stop_reason: "requires_approval",
+      },
+    });
+    stream.receive({
+      type: "update_loop_status",
+      runtime,
+      event_seq: 3,
+      emitted_at: "2026-06-11T00:00:00.002Z",
+      idempotency_key: "loop-1",
+      loop_status: {
+        status: "EXECUTING_CLIENT_SIDE_TOOL",
+        active_run_ids: ["run-approval"],
+      },
+    });
+    stream.receive({
+      type: "stream_delta",
+      runtime,
+      event_seq: 4,
+      emitted_at: "2026-06-11T00:00:00.003Z",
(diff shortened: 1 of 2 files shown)


overlord-letta:
Reviewed by Amelia — this is legit.

**Bug:** `runTurn()` was resolving on any `stop_reason` message, including intermediate `requires_approval` stops. In unrestricted permission mode, the listener auto-approves, executes the tool, and continues to `end_turn` — but the client resolved prematurely on the `requires_approval` stop, making SDK consumers see a terminal result with `stopReason: "requires_approval"` and `approvalConflict: true`.

**Fix:** Correct. `runTurn()` now waits through raw `requires_approval` stop reasons and only terminalizes when:
1. A final stop reason (e.g. `end_turn`) arrives, or
2. The listener emits `WAITING_ON_APPROVAL` loop status (genuine pending approval)

Verified that `WAITING_ON_APPROVAL` is actually emitted by `requestApprovalOverWS` in `approval.ts:300` — the signal is real, not speculative.

Tests pass (13/13). Both new tests cover the right cases — auto-handled continuation vs genuine pending approval. Diff is clean.

PR #3008, opened 2026-06-19
feat(app-server): add Codex-style websocket auth guard

cpacker:
## Summary

Adds Codex-style websocket auth for `letta app-server` so the server can safely bind to non-loopback hosts.

- Loopback listeners (`localhost`, `127.x.x.x`, `::1`) still work without auth.
- Non-loopback listeners now refuse startup unless websocket auth is configured.
- Supported auth modes now match current Codex main:
  - `--ws-auth capability-token`
    - `--ws-token-file <absolute-path>` or `--ws-token-sha256 <64-char-hex>`
  - `--ws-auth signed-bearer-token`
    - `--ws-shared-secret-file <absolute-path>`
    - optional `--ws-issuer`, `--ws-audience`, `--ws-max-clock-skew-seconds`
- Requests/upgrades with an `Origin` header are rejected.
- The in-repo app-server client can pass a capability token as `Authorization: Bearer <token>` when using a WebSocket implementation that supports headers.

## Codex parity self-review

I compared this PR against latest `openai/codex` main at `64bdeed9f7adbe60c725153b3fb74ed044a36221`, specifically:

- `codex-rs/app-server-transport/src/transport/auth.rs`
- `codex-rs/app-server-transport/src/transport/websocket.rs`
- `codex-rs/app-server/src/main.rs`
- `codex-rs/app-server/tests/suite/v2/connection_handling_websocket.rs`

The first pushed version only had `capability-token`; latest Codex also has `signed-bearer-token`, so this PR now includes that mode too.

## Easy review path for Caren

1. **Start with `src/websocket/app-server-auth.ts`**
   - New isolated auth helper module.
   - Parses the CLI-shaped auth config.
   - Capability-token mode: loads/trims token files or accepts a SHA-256 digest; validates `Authorization: Bearer <token>` using SHA-256 + constant-time compare.
   - Signed-bearer mode: loads/trims a shared secret, requires at least 32 bytes, validates HS256 JWTs, `exp`, optional `nbf`, optional issuer, optional audience, and clock skew.
   - Contains loopback vs non-loopback detection.

2. **Then review `src/websocket/app-server.ts`**
   - Removes the old unconditional “loopback only” parser restriction.
   - Adds the startup guard: non-loopback + no auth => refuse to start.
   - Adds per-upgrade auth checks.
   - Rejects requests/upgrades with an `Origin` header.

3. **Then review `src/cli/subcommands/app-server.ts`**
   - Wires the Codex auth flags through to `startAppServer`.
   - Updates help text and examples.

4. **Then review `src/app-server-client.ts`**
   - Minimal client-side support for passing a capability bearer token on both control and stream sockets.
   - No protocol frame changes.

5. **Finally review tests**
   - `src/websocket/app-server.test.ts` covers auth parsing, startup refusal, rejected missing/invalid capability tokens, valid capability token connection, signed JWT validation, valid signed bearer connection, and `Origin` rejection.
   - `src/app-server-client.test.ts` covers capability bearer header wiring.

## Behavior matrix

| Listener host | Auth configured? | Result |
| --- | --- | --- |
| `ws://127.0.0.1:4500` | No | Starts, same as before |
| `ws://localhost:4500` | No | Starts, same as before |
| `ws://0.0.0.0:4500` | No | Refuses startup |
| `ws://0.0.0.0:4500` | `capability-token` | Starts; websocket clients need `Authorization: Bearer <token>` |
| `ws://0.0.0.0:4500` | `signed-bearer-token` | Starts; websocket clients need a valid signed HS256 bearer JWT |
| Any host | Request/upgrade includes `Origin` | Rejected |

## Test plan

- `bun test src/app-server-client.test.ts src/websocket/app-server.test.ts`
- `bun run check`

## Notes

This intentionally tracks Codex’s current websocket auth posture: loopback remains frictionless, any non-loopback websocket listener must opt into explicit websocket auth, and the currently supported Codex auth modes are available.

### DIFF
--- a/src/app-server-client.test.ts
+++ b/src/app-server-client.test.ts
@@ -1,6 +1,7 @@
 import { afterEach, describe, expect, test } from "bun:test";
 import {
   type AppServerSocketLike,
+  type AppServerSocketOptions,
   createAppServerClient,
   resolveAppServerChannelUrl,
 } from "./app-server-client";
@@ -14,7 +15,10 @@ class FakeSocket implements AppServerSocketLike {
   readonly sent: string[] = [];
   private readonly listeners = new Map<string, Set<Listener>>();
 
-  constructor(readonly url: string) {
+  constructor(
+    readonly url: string,
+    readonly options?: AppServerSocketOptions,
+  ) {
     FakeSocket.instances.push(this);
   }
 
@@ -79,6 +83,29 @@ describe("app-server client", () => {
         "stream",
       ),
     ).toBe("wss://example.test/ws?channel=stream&token=abc");
+  });
+
+  test("passes capability token as websocket authorization header", () => {
+    createAppServerClient({
+      url: "http://127.0.0.1:4500",
+      authToken: " super-secret-token\n",
+      WebSocket: FakeSocket,
+    });
+    const [control, stream] = FakeSocket.instances;
+    expect(control?.options).toEqual({
+      headers: { Authorization: "Bearer super-secret-token" },
+    });
+    expect(stream?.options).toEqual({
+      headers: { Authorization: "Bearer super-secret-token" },
+    });
+
+    expect(() =>
+      createAppServerClient({
+        url: "http://127.0.0.1:4500",
+        authToken: " \n",
+        WebSocket: FakeSocket,
+      }),
+    ).toThrow(/auth token must not be empty/);
   });
 
   test("connects both sockets and resolves request_id responses", async () => {
--- a/src/app-server-client.ts
+++ b/src/app-server-client.ts
@@ -46,13 +46,20 @@ export interface AppServerSocketLike {
   once?(type: string, listener: (event: unknown) => void): void;
 }
 
+export interface AppServerSocketOptions {
+  headers?: Record<string, string>;
+}
+
 export type AppServerSocketConstructor = new (
   url: string,
+  options?: AppServerSocketOptions,
 ) => AppServerSocketLike;
 
 export interface AppServerClientOptions {
   /** Base app-server URL, e.g. ws://127.0.0.1:4500 or http://127.0.0.1:4500. */
   url: string;
+  /** Optional capability token sent as Authorization: Bearer <token>; requires a WebSocket implementation with header support. */
+  authToken?: string;
   /** Optional WebSocket constructor for Node/tests. Browsers use globalThis.WebSocket. */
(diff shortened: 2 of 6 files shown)


overlord-letta:
Reviewed by Amelia

**Code assessment: approve with a test fix needed**

The auth module (`app-server-auth.ts`) is well-designed:
- Clean separation: parsing (`parseAppServerWebsocketAuthSettings`), policy building (`policyFromSettings`), authorization (`authorizeUpgrade`)
- `timingSafeEqual` for constant-time SHA-256 comparison — prevents timing attacks
- Never stores raw tokens in memory, only digests
- `isLoopbackListenHost` correctly handles IPv4 `127.x.x.x`, IPv6 `::1`, and `localhost`
- Bearer token parsing is thorough (handles arrays, missing separators, empty tokens)
- Startup refusal for non-loopback without auth is good fail-safe design
- Origin header rejection (CSRF protection) is correct
- The `rejectUpgrade` change from `socket.write()` + `socket.destroy()` to `socket.end()` is a good improvement — lets the client read the HTTP error response before the connection closes

Client changes are minimal and clean — `authToken` option, `appServerSocketOptions` helper, token trimming/validation.

**Test failures (3/7):** All three failures are caused by the `expectWebSocketOpenFailure` test helper, not the auth code. When the server rejects an upgrade with `socket.end()`, the `ws` client emits an `"error"` event. The helper catches it via `socket.once("error", handleError)`, which calls `cleanup()` (removing the listener) and resolves — but does NOT call `terminateClient(socket)`. The `ws` library then emits additional error events when the underlying socket closes, and since the listener was removed, they become unhandled `ErrorEvent`s.

Fix: add `terminateClient(socket)` in `handleError` (line 72-74 of `app-server.test.ts`), matching what `handleOpen` and the timeout handler already do. The third failing test ("starts a runtime") is collateral — it catches unhandled errors from prior tests' dangling sockets.

The core auth logic tests all pass, including the startup refusal test and the unit tests for `authorizeUpgrade` and `parseAppServerWebsocketAuthSettings`.

cpacker:
Addressed Amelia’s test-helper review in `2547017d`: `expectWebSocketOpenFailure` now terminates the rejected `ws` client after handling the expected error event, so later socket-close errors cannot surface as unhandled test failures.

Validated locally:
- `bun test src/websocket/app-server.test.ts src/app-server-client.test.ts`
- `bun run check`

PR #3012, opened 2026-06-20
fix(app-server): ignore stale approval loop status

cpacker:
## Summary
- ignore stale pre-turn WAITING_ON_APPROVAL / waiting loop-status frames before real turn evidence
- avoid recording stale run ids from ignored terminal loop-status frames
- add a regression covering stale approval status followed by normal turn activity

## Test Plan
- bun test src/app-server-client.test.ts
- bun run check

### DIFF
--- a/src/app-server-client.test.ts
+++ b/src/app-server-client.test.ts
@@ -522,6 +522,70 @@ describe("app-server client", () => {
     await expect(turn).rejects.toThrow("Timed out waiting for app-server turn");
   });
 
+  test("runTurn ignores waiting-on-approval status before turn evidence", async () => {
+    const { client, control, stream } = createFakeClient();
+    control.open();
+    stream.open();
+    await client.connect();
+
+    const runtime = { agent_id: "agent-1", conversation_id: "conv-1" };
+    const turn = client.runTurn(
+      {
+        runtime,
+        payload: {
+          kind: "create_message",
+          messages: [{ role: "user", content: "hello" }],
+        },
+      },
+      { allowLoopStatusFallback: true },
+    );
+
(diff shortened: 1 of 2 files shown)


overlord-letta:
Reviewed by Amelia

Legit fix, not slop. Here's the bug it addresses:

When a stale `WAITING_ON_APPROVAL` loop status frame (with `active_run_ids`) from a previous turn arrives before any stream deltas for the new turn, the original code would:
1. Iterate `active_run_ids` — adding stale run IDs to `runIds` and setting `observedTurnEvidence = true`
2. Hit the terminal check `(observedTurnEvidence || observedRequiresApprovalStop) && isWaitingOnApprovalLoopStatus(message)` — now `true` because step 1 just set `observedTurnEvidence`
3. Prematurely terminate the turn with `requires_approval`, and contaminate `runIds` with stale run IDs

The fix captures `hadTurnEvidenceBeforeLoopStatus` before processing `active_run_ids`, ignores the frame entirely if no evidence yet, and uses the captured value for the terminal checks. Clean and minimal — the `hadTurnEvidenceBeforeLoopStatus` variable also reads better than the previous inline `observedTurnEvidence || observedRequiresApprovalStop` expressions that were duplicated.

Test is well-constructed: stale `WAITING_ON_APPROVAL` with `active_run_ids: ["stale-run"]` → real stream deltas with `run_id: "run-1"` → `end_turn` stop reason. Asserts `runIds: ["run-1"]` (not "stale-run") and `completedBy: "stop_reason"` (not "loop_status_waiting_on_approval").

15/15 tests pass, all 9 checks pass.

PR #3016, opened 2026-06-21
fix(recovery): retry as regular message on no-awaiting-approval desync

adrianchan94:
## Problem

A Skill/tool approval round-trip that gets interrupted (provider/model swap, reconnect, a tool-call parse glitch) leaves a stale approval continuation that the client re-sends on the next turn. The server rejects it with:

> No tool call is currently awaiting approval. Please send a regular message to interact with the agent.

On `main` this string is deliberately classified as **neither** `isInvalidToolCallIdsError` **nor** `isApprovalPendingError` (it's the *opposite* state — there are no pending approvals), and `shouldAttemptPostStopApprovalRecovery` returns `false` for it. That was a reasonable guard against the *old* recovery, where "retry" meant re-deriving historical approvals and sending **fresh denials** — which re-trips the same error in a loop. The net effect today, though, is that the conversation stays **wedged**: no self-heal, the error card is surfaced, and it survives `/clear`, restart, and `/new`.

Reproduced on a Constellation/cloud agent (CLI 0.27.14); confirmed via the messages API that the conversation held dangling `approval_request_message`s with no responses.

## Fix

Give the desync a distinct, safe **recovery disposition** instead of leaving it unrecoverable: **strip the stale approval payload and retry the turn as a regular message** — exactly what the server instructs.

- `agent/turn-recovery-policy.ts`: add `isNoAwaitingApprovalError` (distinct from `isApprovalPendingError` — the opposite state) and `selectStaleApprovalRecoveryApprovals(existing, …errorTexts)`, which returns `[]` for this class so `rebuildInputWithFreshDenials` strips rather than re-denies.
- `websocket/listener/recovery.ts`: `isApprovalToolCallDesyncError` now also matches the no-awaiting desync (drives post-stop recovery + error-card suppression). **`isApprovalPendingError` and the pre-stream / headless deny paths are intentionally left untouched.**
- `websocket/listener/turn.ts` + `cli/app/use-conversation-loop.ts`: route `existingApprovals` through the selector at the post-stop recovery sites; the TUI gate now fires for this class. **Edge guard:** if stripping a pure stale approval leaves nothing to send, finalize the turn cleanly instead of issuing an empty regular-message request.
- `agent/approval-recovery.ts`: re-export the two new helpers.

### Behavior change (intentional)

This flips the post-stop gate for the no-awaiting class from "don't retry" to "retry with the strip disposition", so the previously-encoded `listen-client-protocol` test is updated (renamed + expectation flipped) **with rationale**. The classifier-purity invariants are preserved: `isApprovalPendingError` / `isInvalidToolCallIdsError` still return `false` for this string (covered by `check-approval.test.ts` and a new purity test). Discussed with the Letta team (LET-9251).

## Tests

`agent/turn-recovery-policy.test.ts`: `isNoAwaitingApprovalError` (case-insensitive, rejects opposite/non-string), classifier purity ("keeps no-awaiting out of approval-pending / deny-oriented classifiers"), and `selectStaleApprovalRecoveryApprovals` — returns `[]` for this class, keeps approvals for the others, **produces no `ApprovalCreate` in the rebuilt input**, and documents the empty-input edge.
`websocket/listen-client-protocol.test.ts`: post-stop recovery now retries the no-awaiting desync (strip-and-retry), with a rationale comment.

Validation: `tsc --noEmit` clean · `biome check` clean (no new warnings) · `bun test` across `turn-recovery-policy`, `check-approval`, `listen-client-protocol`, `listen-client-concurrency`, `recovery`, `headless-approval-recovery`, `reflection-transcript-wiring` → **331 pass / 0 fail**.

The separate TUI `requires_approval` + empty `approvalsFromStream` → backend `pendingApprovals` fallback remains out of scope (a follow-up PR), per review.


cpacker:
Should be fixed as part of larger refactor here: #3301

PR #3019, opened 2026-06-21
fix(cron): validate conversation targets

### LINKED ISSUE #3018: cron: validate --conversation targets before scheduling
## Summary

`letta cron add --conversation <value>` should reject invalid conversation targets instead of accepting arbitrary strings.

The CLI currently accepts values like `--conversation cron-status` and stores them as `task.conversation_id`. That looks successful, but `cron-status` is not a real conversation id unless the user happened to create a conversation with exactly that id. Later the scheduler will try to route the cron into a nonexistent target.

## Why this matters

This is an easy footgun for agents and users because `--conversation` reads like it might accept a conversation name/label. For cron setup, a false success is worse than a loud failure: the task appears active but is bound to a target that may not exist.

## Current behavior

Example command:

```bash
letta cron add \
  --name "memory-maintenance" \
  --description "Memory hygiene" \
  --cron "0 */6 * * *" \
  --agent agent-... \
  --conversation cron-memory \
  --prompt '...'
```

Observed/expected current output shape stores:

```json
{
  "status": "active",
  "conversation_id": "cron-memory"
}
```

No validation confirms that `cron-memory` exists.

## Expected behavior

When `--conversation` is provided:

- allow `default`
- allow `new` if retaining the current per-fire sentinel behavior
- otherwise require a valid existing conversation id
- fail before creating the cron if the conversation cannot be retrieved or does not belong to the selected agent

The error should explain the distinction, e.g.:

> Invalid conversation target `cron-memory`. `--conversation` expects `default`, `new`, or an existing conversation id like `conv-...`. To create a stable dedicated conversation for this cron, use `--dedicated`.

## Acceptance criteria

- `letta cron add --conversation some-label ...` fails unless `some-label` is a real conversation id.
- The command does not write a cron task on validation failure.
- Existing `default` behavior continues to work.
- Existing `new` sentinel behavior continues to work, or is migrated intentionally with tests.
- If a valid conversation id belongs to another agent, the command fails.
- Tests cover valid id, invalid id, `default`, `new`, and wrong-agent cases.

overlord-letta:
## Summary
- Validate cron add conversation targets before persisting tasks.
- Preserve the existing default and new sentinels while rejecting missing or wrong-agent conversation ids.
- Add targeted CLI coverage for valid ids, invalid labels, default, new, and wrong-agent targets.

Fixes #3018

## Test plan
- [x] bun test ./src/cli/subcommands/cron.test.ts
- [x] bun test ./src/cron
- [x] bunx --bun @biomejs/biome@2.2.5 check src/cli/subcommands/cron.ts src/cli/subcommands/cron.test.ts
- [x] bun run typecheck

### DIFF
--- a/src/cli/subcommands/cron.test.ts
+++ b/src/cli/subcommands/cron.test.ts
@@ -1,0 +1,221 @@
+import {
+  afterEach,
+  beforeEach,
+  describe,
+  expect,
+  mock,
+  spyOn,
+  test,
+} from "bun:test";
+import { mkdirSync, rmSync } from "node:fs";
+import path from "node:path";
+import { __testSetBackend, type Backend } from "@/backend";
+import { runCronSubcommand } from "@/cli/subcommands/cron";
+import { listTasks } from "@/cron";
+
+const TEST_DIR = path.join(import.meta.dir, "__cron_subcommand_test_tmp__");
+
+const retrieveConversationMock = mock(async (conversationId: string) => ({
+  id: conversationId,
+  agent_id: "agent-test",
+}));
+
+const backend = {
+  capabilities: {
+    remoteMemfs: true,
+    serverSideToolManagement: true,
+    serverSecrets: true,
+    agentFileImportExport: true,
+    promptRecompile: true,
+    byokProviderRefresh: true,
+    localModelCatalog: false,
+    localMemfs: false,
+  },
+  retrieveConversation: retrieveConversationMock,
+} as unknown as Backend;
+
+type SavedEnv = {
+  LETTA_HOME?: string;
+  LETTA_AGENT_ID?: string;
+  LETTA_CONVERSATION_ID?: string;
+};
+
+let savedEnv: SavedEnv;
+
+function captureConsole() {
+  const stdout: string[] = [];
+  const stderr: string[] = [];
+  const logSpy = spyOn(console, "log").mockImplementation(
+    (...args: unknown[]) => {
+      stdout.push(args.map((arg) => String(arg)).join(" "));
+    },
+  );
+  const errorSpy = spyOn(console, "error").mockImplementation(
+    (...args: unknown[]) => {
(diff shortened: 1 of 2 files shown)


src/cli/subcommands/cron.ts
```diff
@@ -98,9 +99,44 @@ function getConversationId(fromArgs?: string): string {
...
+function invalidConversationTargetMessage(conversationId: string): string {
+  return `Invalid conversation target "${conversationId}". --conversation expects "default", "new", or an existing conversation id like "conv-..."; labels/names are not accepted.`;
+}
+
+async function validateConversationTarget(
+  agentId: string,
+  conversationId: string,
+): Promise<void> {
+  if (
+    conversationId === DEFAULT_CONVERSATION_TARGET ||
```
overlord-letta, line 116: This preserves the two existing scheduler sentinels: `default` keeps using the agent default conversation, and `new` keeps the per-fire fresh-thread behavior. Everything else now has to resolve as a real conversation id before we persist the task.

overlord-letta (commented):
Overlord (agent-c2adbf5c-8419-4211-8cd8-3740db164974): inline notes for reviewer context.

src/cli/subcommands/cron.test.ts
```diff
@@ -0,0 +1,221 @@
...
+      throw new Error("not found");
+    });
+
+    const capture = captureConsole();
+    try {
+      const code = await runCronSubcommand(cronAddArgs("cron-memory"));
+
+      expect(code).toBe(1);
+      expect(retrieveConversationMock).toHaveBeenCalledWith("cron-memory");
+      expect(listTasks()).toHaveLength(0);
```
overlord-letta, line 157: This is the regression guard for the reported footgun: labels like `cron-memory` fail before `addTask`, so invalid conversation targets do not get written to local cron storage.

sarahwooders:

PR #3036, opened 2026-06-23
feat(mods): manage installed packages

carenthomas:
## Summary
- extend `letta mods list` with installed package visibility and registry diagnostics
- add local-only package `enable`, `disable`, and `remove` commands for already-installed managed mods
- guard `remove` so it only deletes the expected npm package root, not arbitrary loose mods or broad package directories

## Tests
- `bun run check`

### DIFF
--- a/src/mods/package-registry.ts
+++ b/src/mods/package-registry.ts
@@ -1,7 +1,8 @@
-import { existsSync, readFileSync } from "node:fs";
+import { existsSync, readFileSync, rmSync, writeFileSync } from "node:fs";
 import path from "node:path";
 import {
   isSafeLettaPackageModEntryPath,
+  type LettaPackageCapability,
   readLettaPackageManifest,
 } from "@/mods/package-manifest";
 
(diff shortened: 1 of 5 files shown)


src/mods/package-registry.ts
```diff
@@ -360,69 +521,303 @@ function resolvePackage(
...
+  source: string;
+  version: string;
+}): string {
+  return `${pkg.source}@${pkg.version}`;
+}
+
+function expectedRootRelativePathForSource(source: string): string | null {
+  if (!source.startsWith("npm:")) return null;
+  const packageName = source.slice("npm:".length);
+  const normalizedPackageName = normalizeRelativePath(packageName);
```
carenthomas, line 660: This still allows `source: "npm:@scope"` + `root: "packages/npm/@scope"` to pass the removal guard, which would recursively delete the whole scope directory. Can we validate npm sources as either `name` or `@scope/name` before deriving the expected removal root?

carenthomas (commented):
found one safety issue around package removal

PR #3038, opened 2026-06-23
feat(mods): install local packages

carenthomas:
## Summary
- add a local managed mod package installer that validates package.json#letta, copies into the managed packages dir, and upserts packages.json
- expose local package install through top-level `letta install ./path/to/package` while keeping `letta mods` for list/enable/disable/remove
- derive package roots from npm sources and preserve existing installs on failed replacement rollback

## Tests
- `bun test src/mods/package-installer.test.ts`
- `bun test src/cli/subcommands/skills.test.ts`
- `bun run typecheck`
- `bun run check`

### DIFF
--- a/src/mods/package-installer.ts
+++ b/src/mods/package-installer.ts
@@ -1,0 +1,339 @@
+import {
+  copyFileSync,
+  existsSync,
+  lstatSync,
+  mkdirSync,
+  mkdtempSync,
+  readdirSync,
+  readFileSync,
+  renameSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
+import path from "node:path";
+import {
+  type LettaPackageCapability,
+  readLettaPackageManifest,
+} from "@/mods/package-manifest";
+import {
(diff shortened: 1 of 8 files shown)


carenthomas (commented):
found one rollback issue

src/mods/package-installer.ts
```diff
@@ -0,0 +1,339 @@
...
+        entries: packageInfo.entries,
+        packageDirectory: packageInfo.packageDirectory,
+        registryPath: upsertResult.registryPath,
+        root: destinationRoot,
+        rootRelativePath: packageInfo.rootRelativePath,
+        source: packageInfo.source,
+        version: packageInfo.version,
+      };
+    } catch (error) {
+      restoreDestinationIfNeeded({ backupRoot, destinationRoot });
```
carenthomas: If this is a first install, `backupRoot` is null, so a registry write failure after `renameSync(stagingRoot, destinationRoot)` leaves copied package files installed but unregistered. We should remove `destinationRoot` when there is no backup before restoring the registry.

PR #3040, opened 2026-06-23
feat(mods): install npm packages

carenthomas:
## Summary
- add `letta install npm:<package>` for managed mod package installs
- run npm with safer install defaults and reuse managed package rollback/upsert behavior
- load managed package entries from package-local cache paths so package dependencies resolve

## Test plan
- `bun test src/mods/package-installer.test.ts`
- `bun test src/cli/subcommands/skills.test.ts`
- `bun run typecheck`
- `bun run check`

### DIFF
--- a/src/mods/package-installer.ts
+++ b/src/mods/package-installer.ts
@@ -1,4 +1,9 @@
 import {
+  type ChildProcess,
+  type SpawnOptions,
+  spawn,
+} from "node:child_process";
+import {
   copyFileSync,
   existsSync,
   lstatSync,
@@ -10,6 +15,7 @@ import {
   rmSync,
   writeFileSync,
 } from "node:fs";
+import { tmpdir } from "node:os";
 import path from "node:path";
 import {
   type LettaPackageCapability,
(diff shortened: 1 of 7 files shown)


src/mods/package-installer.ts
```diff
@@ -349,3 +502,153 @@ export function installLocalManagedModPackage(params: {
...
+    throw new Error(`Invalid npm mod package specifier: ${specifier}`);
+  }
+
+  const slashIndex = trimmed.lastIndexOf("/");
+  const atIndex = trimmed.lastIndexOf("@");
+  let source = trimmed;
+  let version: string | undefined;
+  if (atIndex > "npm:".length && atIndex > slashIndex) {
+    source = trimmed.slice(0, atIndex);
+    version = trimmed.slice(atIndex + 1);
```
carenthomas, line 602: This accepts arbitrary npm spec suffixes like `npm:alias@npm:other` or `npm:pkg@file:..`, which pulls PR6 into alias/file installs even though those are out of scope. Can we validate the optional suffix as a plain version/dist-tag here, at least rejecting `:`, `/`, and `\` before building `installSpec`?

carenthomas (commented):
found one scope/safety issue around npm spec parsing

src/mods/package-installer.ts
carenthomas: fixed in f19690e1, optional npm suffixes now reject non-plain version/tag values before building the install spec.

PR #3046, opened 2026-06-23
feat(worktree): auto-provision, enter, and lock worktrees in CreateWorktree

devanshrj:
## Summary

Makes `CreateWorktree` produce worktrees that are immediately usable, lets a conversation switch into an existing one, and prevents two agents from stepping on each other in the same worktree.

### Auto-provisioning new worktrees
A fresh worktree previously needed a manual setup pass (missing `.env`, dead git hooks). Creation now provisions it best-effort:
- **Git hooks**: when `core.hooksPath` is relative (e.g. husky's `.husky/_`, whose contents are gitignored and absent in a fresh worktree), symlinks the populated hooks dir from the primary checkout so pre-commit hooks actually run.
- **Copies** `.letta/settings.local.json` and any gitignored paths listed in a repo-root `.worktreeinclude` file (e.g. `.env`).
- **`--no-track`** on `git worktree add` so the new branch doesn't adopt the base ref as its upstream (which produced misleading ahead/behind status and risked an accidental push to the base branch).
- Configurable via a `worktree` block in `.letta/settings.json` (`symlinkDirectories`, `copyLocalSettings`, `linkHooks`, `include`). Best-effort: reports what it did/skipped in the result and never aborts creation.

### Dependency sharing is opt-in (`symlink_dependencies`, default `false`)
Sharing `node_modules` via symlink means a package install in the worktree writes *through* to the primary checkout and can silently corrupt its deps. So worktrees are **isolated by default**; set `symlink_dependencies: true` to share `node_modules` from the primary checkout (only for tasks that won't change packages). The result message guides the agent accordingly — install in the isolated worktree by default, or (when shared) a warning not to install. Modern package managers (Bun/pnpm) install fast from a shared cache, so the isolated default is cheap. (See discussion thread on the default choice.)

### Enter an existing worktree (`path`)
`CreateWorktree` accepts `path` to switch the conversation into an existing worktree instead of creating one — to resume work or hand off one another conversation started. Validation-only: the target must be a registered, non-prunable linked worktree under `.letta/worktrees/`; the main working tree and unmanaged/external worktrees are rejected. Mutually exclusive with `name`/`branch_name`/`base_ref`.

### Cross-agent lock
Switching into a worktree (by creating or entering) takes an advisory lock recording the owning conversation, stored as JSON in the worktree's per-worktree git admin dir (out of the working tree; removed automatically by `git worktree remove`).
- Entering a worktree another **active** agent holds fails with a clear error; `force: true` takes over.
- Locks from a crashed/exited process (dead pid, same host) are reclaimed automatically.
- Re-entering a worktree you already hold is a no-op; the lock is released when the conversation switches to a different worktree, so moving between worktrees never strands a self-held lock.

### Internal
- The pure file-backed lock primitive lives in `src/utils/worktree-lock.ts`; the tool keeps only the session-aware orchestration.

## Test plan
- `bun test src/tools/create-worktree.test.ts` — 22 tests: provisioning (hooks wiring, `.worktreeinclude`/`.env` copy, `settings.local.json` copy); `symlink_dependencies` opt-in (shares on `true`, isolated by default); enter a managed worktree + reject external/unregistered/`path`+`name`; lock (block a second agent + `force` override, re-entrancy, release-on-switch, stale-lock reclaim).
- `bun run check` — all 9 checks pass (cycles, layer boundaries, exported-fn style, filename casing, mock isolation, biome, tsc).


cpacker:
@carenthomas @just-cameron we should double-check this works fine on windows

cpacker (changes requested):
Should make sure that the symlinking pattern is a very clear and obvious disable-able flag to the agent, in case the agent needs to re-setup eg node-modules for diff packages etc

devanshrj:
> Should make sure that the symlinking pattern is a very clear and obvious disable-able flag to the agent…

Good call — addressed, and on reflection we flipped the default rather than just adding a disable flag.

The hazard with symlink-on-by-default is that `node_modules` is a symlink into the primary checkout, so a `bun install` (or any write into `node_modules`) in the worktree writes *through* to the main checkout and can silently corrupt its deps — exactly the "messing things up" risk, and agents install reflexively. Since Bun/pnpm already install fast from a shared cache, the isolated default costs little while removing the footgun.

So as of `de5367c5`:
- **`symlink_dependencies` defaults to `false`** — worktrees are isolated; the result message tells the agent to install deps there.
- Set **`symlink_dependencies: true`** to opt into sharing (for tasks that won't touch packages); the message then warns *not* to install.

Schema + tool description updated to make the flag and its risk obvious to the agent. LMK if you'd rather keep it default-on with just the disable flag.

cpacker (approved):

PR #3049, opened 2026-06-23
fix(mods): load legacy extensions alongside mods

carenthomas:
## Summary
- load legacy `~/.letta/extensions` files before `~/.letta/mods` during the migration window
- give `~/.letta/mods` higher source priority so new mods shadow legacy extensions
- emit warning diagnostics telling users to move each loaded legacy extension into `~/.letta/mods`

## Tests
- `bun test src/mods/mod-diagnostics.test.ts src/mods/mod-engine.test.ts src/mods/paths.test.ts`
- `bun run typecheck`
- `bun run check`

### DIFF
--- a/src/mods/mod-engine.ts
+++ b/src/mods/mod-engine.ts
@@ -54,7 +54,6 @@ import {
   getGlobalModsDirectory,
   getLegacyGlobalExtensionsDirectory,
   getModCacheDirectory,
-  resolveDefaultGlobalModsDirectory,
 } from "@/mods/paths";
 import {
   getModPermissionDefinition,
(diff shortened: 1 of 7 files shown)


src/mods/mod-engine.ts
```diff
@@ -314,27 +317,43 @@ export function resolveLocalModSources(
   options: ResolveLocalModSourcesOptions = {},
 ): LocalModSource[] {
   const globalModsDirectory =
-    options.globalModsDirectory ?? resolveDefaultGlobalModsDirectory();
+    options.globalModsDirectory ?? getGlobalModsDirectory();
+  const legacyGlobalExtensionsDirectory =
+    options.legacyGlobalExtensionsDirectory ??
+    (options.globalModsDirectory
```
amelia-letta, line 323: Callers that pass `globalModsDirectory` explicitly (e.g. `listMods` in `mods.ts`) skip legacy auto-resolution — `/mods` won't show legacy extensions that are actively being loaded.

carenthomas: fixed in e5a01857, `/mods list` now passes the legacy extensions directory through and formats a Legacy extensions section so it matches runtime loading.

PR #3050, opened 2026-06-23
fix: support Windows credential auth in Node CLI

just-cameron:
## Summary
- Add a Node-compatible Windows Credential Manager backend for secure auth persistence.
- Use the same credential target format as Bun/Desktop so npm CLI and Desktop can share saved auth.
- Batch token reads, preserve graceful fallback behavior, and keep settings.json fallback for unavailable secure storage.
- Add Windows credential adapter tests, including UTF-8 and Bun.secrets compatibility coverage.

## Testing
- bun test src/utils/secrets.test.ts
- bun test src/settings-manager.test.ts
- bunx --bun @biomejs/biome@2.2.5 check src/utils/secrets.ts src/utils/windows-credential-manager.ts src/utils/secrets.test.ts
- bun run typecheck
- bun run check
- Node smoke test with LETTA_API_KEY removed and temporary HOME: exit=0, no fallback tokens written to settings.json

## Notes
- On Windows, bun run build emits letta.js but still fails at the existing Unix chmod step; the emitted bundle was used for the Node smoke test.

### DIFF
--- a/src/utils/secrets.test.ts
+++ b/src/utils/secrets.test.ts
@@ -2,6 +2,7 @@
 // Tests for secrets utility functions
 
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
+import { randomUUID } from "node:crypto";
 import {
   __resetSecretWarningStateForTests,
   __setSecretGetOverrideForTests,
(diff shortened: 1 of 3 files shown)


sarahwooders:

PR #3052, opened 2026-06-23
feat(secrets): support local agent storage

carenthomas:
## Summary
- store local agent secrets in OS secure storage via Bun.secrets-backed helpers
- keep local secrets scoped by agent id and hydrate the existing in-memory secrets cache
- route `/secret set/list/unset` and secret batch apply through local storage for local agents
- validate secret keys at the store boundary so invalid keys cannot create orphan local entries
- update help text to distinguish cloud/server and local/OS credential manager storage

## Tests
- `bun test src/cli/commands/secret.test.ts src/cli/commands/registry.test.ts src/tools/secret-substitution.test.ts src/tools/shell-secrets.test.ts src/reminders/engine.test.ts src/websocket/listener-secrets-sync.test.ts`
- `bun run typecheck`
- `bun run check`

### DIFF
--- a/src/utils/secrets-store.ts
+++ b/src/utils/secrets-store.ts
@@ -1,11 +1,18 @@
 /**
- * Server-backed secret storage for Letta Code.
- * Secrets are stored on the Letta server via the agent secrets API
- * and cached in memory for fast $SECRET_NAME substitution in shell commands.
+ * Agent-scoped secret storage for Letta Code.
+ * Cloud agent secrets are stored on the Letta server. Local agent secrets are
+ * stored in the operating system credential manager through Bun.secrets. Both
+ * paths hydrate the same in-memory cache for fast $SECRET_NAME substitution.
  */
 
+import { isLocalAgentId } from "@/agent/agent-id";
 import { getCurrentAgentId } from "@/agent/context";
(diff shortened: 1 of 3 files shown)


src/utils/secrets-store.ts
```diff
@@ -19,18 +26,41 @@ type SecretsBackend = {
...
 
+function getLocalSecretStorage(): LocalSecretStorage {
+  return (
+    testLocalSecretStorageOverride ?? {
+      delete: deleteSecretValue,
+      get: getSecretValue,
+      set: setSecretValue,
+    }
+  );
+}
```
amelia-letta, line 164: `setSecretValue` throws when `secretsAvailable` is false, which is the case when CLI runs on Node (production). Local agent secrets won't work in production without a fallback like settings-manager's pattern — check `isKeychainAvailable()` and fall back to file-based storage if unavailable.

carenthomas: fixed in 975057a5, local agent secrets now use `isKeychainAvailable()` and fall back to a file under the local backend storage dir when OS secure storage is unavailable. Added a test with `LETTA_SKIP_KEYCHAIN_CHECK=1` to cover the Node-production fallback path.

PR #3061, opened 2026-06-23
feat(artifacts): add artifact app system

4shub:


src/websocket/listener/commands/artifacts.ts
```diff
@@ -0,0 +1,574 @@
...
+    path: dataPath,
+    read: () => {
+      if (!existsSync(dataPath)) {
+        return null;
+      }
+      const content = readFileSync(dataPath, "utf8");
+      if (content.trim().length === 0) {
+        return null;
+      }
+      return content;
```
amelia-letta: `data.read()` returns the raw JSON string, but the docs and example server code expect a parsed object. `updateState` in the example would crash with `TypeError` because `current.items` is undefined when `current` is a string.

```suggestion
      return JSON.parse(content);
```

src/websocket/listener/commands/artifacts.ts
```diff
@@ -0,0 +1,574 @@
...
+      logs.push({
+        level,
+        message: args.map(formatConsoleArg).join(" "),
+        timestamp: new Date().toISOString(),
+      });
+      original(...args);
+    };
+  };
+
+  console.log = createLogger("log", originalLog);
```
amelia-letta: Replacing global `console.*` is not safe for concurrent calls. If two artifact calls overlap, the `finally` block restores the wrong reference, leaving console permanently wrapped with a stale logger.

src/tools/impl/artifact-call.ts
```diff
@@ -0,0 +1,110 @@
...
+
+export async function artifact_call(args: ArtifactCallArgs): Promise<{
+  content: string;
+}> {
+  const agentId = getCurrentAgentId();
+  if (!agentId) {
+    throw new Error("artifact_call: current agent id is unavailable");
+  }
+
+  const memoryRoot = getMemoryFilesystemRoot(agentId);
```
amelia-letta: Should use `getScopedMemoryFilesystemRoot` to match `list_memory_directory` and other memory handlers. With local backend enabled, `getMemoryFilesystemRoot` returns the wrong path.

src/websocket/listener/commands/memory.ts
```diff
@@ -993,5 +1123,164 @@ export function handleMemoryProtocolCommand(
...
+        const {
+          getMemoryFilesystemRoot,
+          ensureLocalMemfsCheckout,
+          isMemfsEnabledOnServer,
+        } = await import("@/agent/memory-filesystem");
+        const { existsSync } = await import("node:fs");
+        const { join } = await import("node:path");
+        const { callArtifactServerFunction } = await import("./artifacts");
+
+        const memoryRoot = getMemoryFilesystemRoot(parsed.agent_id);
```
amelia-letta: Same issue as the tool path. Should use `getScopedMemoryFilesystemRoot` to match `list_memory_directory` handler above.

sarahwooders:

PR #3062, opened 2026-06-23
feat(mods): support GitHub package installs

carenthomas:
## Summary
- Add `letta install https://github.com/owner/repo` and `letta install git:github.com/owner/repo` for managed mod packages
- Install GitHub packages under `packages/git/github.com/owner/repo` with normalized `git:https://github.com/owner/repo` registry sources
- Support package.json#letta manifests first, with a compatibility fallback for conventional mod entries like `src/mod.ts`
- Run safe runtime dependency installs for GitHub packages with `npm install --ignore-scripts --omit=dev --no-audit --no-fund --package-lock=false --no-save`

## Tests
- `bun test src/mods/package-registry.test.ts src/mods/package-installer.test.ts src/cli/subcommands/skills.test.ts`
- `bun run check`
- Smoke-tested temp installs for `klittle32/letta-mcp-mod` and `rl-0x000/letta-secrets-scanning-mod`

### DIFF
--- a/src/cli/subcommands/skills.test.ts
+++ b/src/cli/subcommands/skills.test.ts
@@ -443,10 +443,9 @@ describe("skills subcommand", () => {
         },
       });
 
-      const exitCode = await runInstallSubcommand(
-        ["https://github.com/caren/git-mod"],
-        { globalModsDirectory: modsRoot },
-      );
+      const exitCode = await runInstallSubcommand(["github:caren/git-mod"], {
+        globalModsDirectory: modsRoot,
+      });
 
       expect(exitCode).toBe(0);
       expect(consoleCapture.logs.join("\n")).toContain(
--- a/src/cli/subcommands/skills.test.ts
+++ b/src/cli/subcommands/skills.test.ts
@@ -115,6 +115,22 @@ function writeInstalledNpmModPackage(params: {
   );
 }
 
+function writeGitModPackage(packageRoot: string): void {
+  mkdirSync(join(packageRoot, "src"), { recursive: true });
+  writeFileSync(join(packageRoot, "src", "mod.ts"), "export {};\n");
+  writeFileSync(
+    join(packageRoot, "package.json"),
+    `${JSON.stringify(
+      {
+        name: "git-mod",
+        version: "0.1.0",
+      },
+      null,
+      2,
+    )}\n`,
+  );
+}
+
 describe("skills subcommand", () => {
   afterEach(() => {
     __testOverrideNpmManagedModPackageInstaller({});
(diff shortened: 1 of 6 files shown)


carenthomas (commented):
## Testing

Tested with `https://github.com/klittle32/letta-mcp-mod` (compatibility mode, has runtime deps):

- ✅ `letta install https://github.com/klittle32/letta-mcp-mod` — installs to `packages/git/github.com/klittle32/letta-mcp-mod`, writes compatibility letta manifest, installs deps via npm
- ✅ `letta install git:github.com/klittle32/letta-mcp-mod` — shorthand works
- ✅ `letta install https://github.com/klittle32/letta-mcp-mod@main` — ref checkout works (full clone + checkout)
- ✅ Reinstall updates in place
- ✅ Non-GitHub URL rejected (`https://gitlab.com/...`)
- ✅ Agent-scoped install rejected
- ✅ `--force` rejected
- ✅ All 50 tests pass, `bun run check` passes (9/9)

## Review

**Behavior change: local packages no longer copy node_modules**

`copyPackageInternalNodeModules` was unconditional before this PR. Now it's gated behind `includePackageNodeModules`, which is set for npm and git paths but **not** for `installLocalManagedModPackage`. Local packages with runtime dependencies will lose their `node_modules` on copy to the managed packages directory. The `copyPackageDirectory` function skips `node_modules` (via `SKIPPED_PACKAGE_COPY_NAMES`), so this was the only mechanism that copied it.

If this is intentional (local packages resolve from their original location?), it should be documented. If not, `installLocalManagedModPackage` should pass `includePackageNodeModules: true`.

**Minor: `getGitExecutable()` doesn't handle Windows**

`getNpmExecutable()` returns `"npm.cmd"` on Windows, but `getGitExecutable()` just returns `"git"`. Git for Windows puts `git.exe` on PATH so `spawn("git", ...)` should work, but worth confirming on Windows CI.

Everything else looks clean — symlink rejection in both `copyPackageDirectory` and `copyDependencyDirectoryFiltered`, path traversal prevention in `parseManagedGitPackageSource`, temp dir cleanup via try/finally, and good test coverage across all the new code paths.

carenthomas:
Addressed the review items:

- Updated the PR title to `feat(mods): support GitHub package installs` so the conventional-title check can pass.
- Checked the local `node_modules` note against `origin/main`: local package installs already skipped `node_modules` before this PR, and the existing local install test asserts that behavior. The new `includePackageNodeModules` path is only enabled for npm/git installs where we install into a temp package root and need runtime deps copied forward.
- Leaving `getGitExecutable()` as `git`: Git for Windows exposes `git.exe` on PATH and Node can resolve that directly; `npm` needs the `.cmd` special-case because it is commonly a shell shim.

PR #3068, opened 2026-06-24
fix(channels): preserve unresolved credential refs

overlord-letta:
## Summary

A Telegram support report showed an account restarting with an empty token after secure credential hydration failed. The risky path was in `hydrateChannelAccountSecrets`: when an account had a stored secret ref but the keyring read returned no value, it removed the ref, wrote an empty string into the in-memory account, and saved that empty token back to `accounts.json`.

This changes hydration to leave unresolved secret refs intact. If the secure store is temporarily unavailable, the account remains recoverable on the next successful read instead of being converted into a permanently empty-token account.

## Validation

- `env -u USER_CWD bun test src/channels/credential-store.test.ts`
- `env -u USER_CWD bun run typecheck`
- `env -u USER_CWD bun run lint`

## Notes

The new regression test covers Telegram specifically: after a keyring-backed Telegram token is persisted, a later missing keyring read preserves `__letta_secret_refs.token` and does not write `token: ""`.


just-cameron:
Superseded by #3270, which keeps the fail-closed hydration behavior and also covers blank-token config updates for Telegram/Discord/Slack plus enabled-account token validation.

just-cameron:
Closing in favor of #3270.

PR #3069, opened 2026-06-24
fix(channels): preserve missing credential refs

overlord-letta:
Linear: LET-9268

## Summary

This fixes a destructive channel-credential failure mode. When an account has saved secure-store refs but the referenced secret cannot be loaded, hydration now fails with an actionable credential-store error and leaves accounts.json unchanged instead of clearing the ref and saving an empty credential.

Telegram startup now also reports a clear missing BotFather token error before grammY constructs a bot with an empty token.

## Validation

- env -u USER_CWD bun test src/channels/credential-store.test.ts
- env -u USER_CWD bun test src/channels/telegram-adapter.test.ts
- env -u USER_CWD bun run check

## Risk

Low. This changes the broken missing-secret path from silent mutation to explicit failure. Existing plaintext migration and valid keyring hydration paths are covered by the existing credential-store tests.


just-cameron:
Superseded by #3270, which keeps the missing-secret-ref preservation and also covers blank-token config updates for Telegram/Discord/Slack plus enabled-account token validation.

just-cameron:
Closing in favor of #3270.

PR #3071, opened 2026-06-24
fix(local): match reference memory projection

overlord-letta:
## Summary

CI on main and PR #3068 is failing in `local-system-prompt-parity.test.ts` because the API now treats `reference/...` memory-block labels as external memory, while the local backend was still forcing every non-`system/` memory block into `system/`.

This updates local backend initial-memory file placement so `reference/...` labels stay under `reference/` instead of becoming `system/reference/...`. The parity test now asserts the reference block appears through `<external_projection>` rather than as a pinned system memory block.

## Validation

- `env -u USER_CWD bun test src/backend/local-system-prompt-compilation.test.ts src/integration-tests/local-system-prompt-parity.test.ts --timeout 60000`
- `env -u USER_CWD bun run typecheck`
- `env -u USER_CWD bun run lint`

## Notes

This is separate from PR #3068. It repairs the shared CI failure already visible on recent `main` runs and on #3068 platform jobs.

### DIFF
--- a/src/backend/local/local-backend.ts
+++ b/src/backend/local/local-backend.ts
@@ -87,7 +87,11 @@ function sanitizeFrontmatterValue(value: string): string {
 
 function memoryBlockPath(label: string): string {
   const normalized = label.trim().replace(/\\/g, "/").replace(/\.md$/, "");
-  if (normalized === "system" || normalized.startsWith("system/")) {
+  if (
+    normalized === "system" ||
+    normalized.startsWith("system/") ||
+    normalized.startsWith("reference/")
+  ) {
     return `${normalized}.md`;
   }
   return `system/${normalized}.md`;
--- a/src/integration-tests/local-system-prompt-parity.test.ts
+++ b/src/integration-tests/local-system-prompt-parity.test.ts
@@ -63,8 +63,79 @@ async function apiDryRunRecompileWithMemory(agentId: string): Promise<string> {
   return compiled;
 }
 
+/**
+ * Assert structural properties of compiled prompts.
+ *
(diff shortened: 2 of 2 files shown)


just-cameron (approved):

just-cameron (changes requested):
Hm this might be wrong, investigating

overlord-letta:
Reviewing the test this PR modifies (`src/integration-tests/local-system-prompt-parity.test.ts`):

The intent is solid — catching divergence between local and API prompt compilation is worth testing. But the test has structural problems that this PR inherits rather than fixes:

1. Full-string equality (`toBe`) on two independently-evolved compilers produces massive, hard-to-debug diffs when it breaks. Structural assertions (e.g., "reference/ blocks appear in external_projection, not in system memory") would be more useful and more resilient.

2. The previous `toContain` assertions (lines 114-118) were actively wrong — they pinned `reference/details` as a system-memory block with a `<projection>` tag, but the whole point of the `reference/` prefix is that it should be external/progressive memory. The test was asserting the local backend's incorrect behavior, not the correct behavior. This PR fixes the local backend to match the API, but the test still uses full-string equality as the primary assertion, so the next divergence will be equally hard to debug.

3. It depends on a live API (`LETTA_API_KEY`), so it silently skips in environments without the key, which can mask regressions.

Suggestion: replace the `toBe` equality with targeted structural assertions about where each memory block appears (system vs external_projection) and what form it takes. That would make failures actionable and avoid the test being a brittle full-prompt diff.

---

just-cameron:
Solving this issue elsewhere.

PR #3074, opened 2026-06-24
feat(mods): add result effect to tool_start event

carenthomas:
## Summary
- Allow mods to return `{ result: { status, output } }` from `tool_start` handlers to short-circuit tool execution
- First handler wins — subsequent handlers cannot override an existing result

## Test plan
- [x] Unit test: first handler wins when multiple return result
- [x] Unit test: no result when condition not met
- [x] All existing tool_start tests pass
- [x] Type check clean

### DIFF
--- a/src/mods/mod-engine.test.ts
+++ b/src/mods/mod-engine.test.ts
@@ -1104,6 +1104,52 @@ describe("mod engine", () => {
     }
   });
 
+  test("tool_start result: error status populates error field", async () => {
+    const root = createTempDir();
+    try {
+      const modDir = path.join(root, "global-mods");
+      mkdirSync(modDir, { recursive: true });
+      writeFileSync(
+        path.join(modDir, "result-error.ts"),
+        `export default function(letta) {
+          letta.events.on("tool_start", (event) => {
+            if (event.toolName === "Bash" && event.args.command === "dangerous") {
+              return { result: { status: "error", output: "blocked: dangerous command" } };
+            }
+          });
+        }`,
+      );
+
+      const engine = createEngine(root);
+      await engine.reload();
+      const event = {
+        agentId: "agent-1",
+        conversationId: "conversation-1",
+        toolCallId: "toolu-1",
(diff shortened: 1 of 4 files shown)


src/tools/manager.ts
```diff
@@ -2623,14 +2632,17 @@ export async function executeTool(
...
-    const { args: eventArgs } = await emitToolStartEvent({
+    const { args: eventArgs, result } = await emitToolStartEvent({
       args,
       events: modEvents,
       executionScope,
       modContext,
       toolCallId: options?.toolCallId,
       toolName: name,
     });
+    if (result) {
```
amelia-letta, line 2643: Error case returns `{ toolReturn, status: "error" }` but normal errors return `{ error, status: "error" }`. Inconsistent field names — callers might expect `error` when status is "error".

src/tools/manager.ts
```diff
@@ -2657,14 +2669,17 @@ export async function executeTool(
...
```
amelia-letta, line 2683: Same inconsistency here — error case should return `{ error: result.output, status: "error" }`.

src/tools/manager.ts
```diff
@@ -2713,14 +2728,17 @@ export async function executeTool(
...
-  const { args: eventArgs } = await emitToolStartEvent({
+  const { args: eventArgs, result } = await emitToolStartEvent({
     args,
     events: modEvents,
     executionScope,
     modContext,
     toolCallId: options?.toolCallId,
     toolName: internalName,
   });
+  if (result) {
```
amelia-letta, line 2745: Same inconsistency here — error case should return `{ error: result.output, status: "error" }`.

src/mods/mod-engine.test.ts
```diff
@@ -1012,6 +1012,98 @@ describe("mod engine", () => {
     }
   });
```
amelia-letta, line 1014: Missing test for `status: "error"` case — should verify error return format matches normal tool errors.

PR #3085, opened 2026-06-24
feat(headless): emit local tool calls and returns in stream-json

devanshrj:
Makes the headless `--output-format stream-json` output surface local tool I/O and match the shape of other coding agents (Claude Code / Codex). Two commits (the second, #3086, merged into this branch):

1. **Emit local tool calls + returns** (`0d3f3dcf`)
2. **Strip approval events** (`4948fde3`, #3086)

## Problem

Locally-executed tools (Bash, Read, Edit, Write, …) run client-side via `executeApprovalBatch`, and their results are fed back to the server as the next request's input — so they never appear on the server stream. As a result, `stream-json` consumers saw the tool *call* (only via an `auto_approval` event) but never the tool's **output**, while the stream was also cluttered with approval-flow events no other agent emits.

## What this does

Local tools now get the same `tool_call_message` → `tool_return_message` pairing (keyed by `tool_call_id`) that server-side tools already get — i.e. the `tool_use` → `tool_result` shape of Claude Code / Codex — and the approval-flow noise is removed.

### 1. Emit local tool calls + returns
- New `src/headless-tool-events.ts` with `emitLocalToolCalls` / `emitLocalToolReturns` (multimodal returns stringified via `getDisplayableToolReturn`; approval-only results skipped).
- Wired into both headless paths (one-shot `-p` and bidirectional `--input-format stream-json`): emit calls before `executeApprovalBatch`, returns after.
- Unit tests in `src/headless-tool-events.test.ts` (call/return shape, multimodal stringification, empty-args default, approval-only skip).

### 2. Strip approval events
- `headless.ts` (both paths): suppress `approval_request_message` / `approval_response_message` chunks; remove all three `auto_approval` emit sites. The post-drain approval *logic* (classify → execute) is untouched — only what's printed changes. Approvals remain available out-of-band via the control protocol (`can_use_tool`).
- `manager.ts` (the only in-repo consumer of these events): reconstruct subagent tool calls from `tool_call_message` instead of `auto_approval` + partial `approval_request_message` deltas — dropping `pendingToolCalls`, the two approval handlers, and the result-time flush, since it now gets complete args in one event.
- Bidirectional integration test asserts `tool_call_message` instead of `auto_approval`, plus a regression guard that approval events are absent.

## Result

Same forced-`Bash` prompt, before → after: 12 events → 9. The stream is now `system/init` → `reasoning` → `tool_call_message` → `tool_return_message` → `assistant_message` → `result`, with no `auto_approval` or `approval_request_message`.

## Verification

Live runs against real Letta cloud + Sonnet:
- **Single `Bash` call** — `tool_call_message` (complete args) → `tool_return_message` (status + output), paired by `tool_call_id`.
- **Parallel reads (3 at once)** — 3 calls then 3 returns, all paired by id, no errors.
- **Subagent run** — exercises the migrated `manager.ts`; parent records the child's tool calls from `tool_call_message`, completes cleanly.

`bun run check` passes (all 9 checks).

## Note

`AutoApprovalMessage` is now unemitted but left defined in `protocol.ts` (the SDK imports from there); dead-type cleanup can be a separate pass.

### DIFF
--- a/src/agent/subagents/manager.ts
+++ b/src/agent/subagents/manager.ts
@@ -108,7 +108,6 @@ interface ExecutionState {
     stepCount?: number;
   } | null;
   displayedToolCalls: Set<string>;
-  pendingToolCalls: Map<string, { name: string; args: string }>;
 }
 
 // ============================================================================
@@ -384,11 +383,22 @@ function handleInitEvent(
 }
 
 /**
- * Handle an approval request message event
+ * Handle a tool_call_message event. The subagent runs headless with
+ * --output-format stream-json and emits a tool_call_message (with complete
+ * arguments) for every tool it calls — both server-side tools and locally
+ * executed ones. Record each so the subagent's tool-call list stays in sync.
  */
-function handleApprovalRequestEvent(
-  event: { tool_calls?: unknown[]; tool_call?: unknown },
+function handleToolCallEvent(
+  event: {
+    tool_call?: { tool_call_id?: string; name?: string; arguments?: string };
+    tool_calls?: Array<{
+      tool_call_id?: string;
+      name?: string;
+      arguments?: string;
+    }>;
+  },
   state: ExecutionState,
+  subagentId: string,
 ): void {
   const toolCalls = Array.isArray(event.tool_calls)
     ? event.tool_calls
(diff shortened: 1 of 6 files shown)


sarahwooders:

PR #3093, opened 2026-06-25
feat: remove memory permission mode

devanshrj:
## Summary
- Introduce an explicit `SubagentLaunchProfile` (`default` | `memory-subagent`) for a subagent's filesystem/env launch behavior, replacing the overloaded `permissionMode="memory"` signal. The reflection/init/memory/history-analyzer builtins declare `memory-subagent`, forward `MEMORY_DIR` to the parent's repo, run from the inherited root, and rely on the kernel filesystem sandbox for write scoping.
- Launch subagent child processes with `--permission-mode unrestricted`; their writes are constrained by the OS filesystem sandbox, while tool allowlists remain the noninteractive control surface.
- Remove `memory` from `PermissionMode` everywhere (CLI, device/protocol, websocket listener, runtime context, input-hint UI) and delete the static memory-mode fallback/denial logic — `memory-denial-reason.ts` and the `memory` case in `checkModeOverride`.
- Rename the sandbox-policy builder `buildMemoryModeSandboxPolicy` → `buildMemorySubagentSandboxPolicy` and update docs/comments from memory-mode / parent-profile wording to "memory-subagent profile" / "write-scoped profile".

## Test plan
- [x] `bun test $(find src/permissions -name '*.test.ts')` — 493 pass
- [x] `bun test src/agent/subagent-builtins.test.ts src/agent/subagent-env-composition.test.ts src/agent/subagent-model-resolution.test.ts`
- [x] `bun test src/agent/subagents/sandbox.test.ts src/permissions/sandbox-policy.test.ts src/permissions/cross-agent-guard.test.ts`
- [x] `bun test src/cli/approval-classification.test.ts src/websocket/listener-permission-mode.test.ts src/websocket/listener/protocol-inbound.test.ts`
- [x] `bun test src/agent/subagent-builtins.test.ts src/agent/subagent-env-composition.test.ts src/agent/subagent-model-resolution.test.ts src/agent/subagents/sandbox.test.ts src/permissions/sandbox-policy.test.ts`
- [x] `bun run typecheck`
- [x] `bun run check` (cycles / boundaries / exported-functions / biome / typescript)
- [ ] Post-rename GitHub CI — running


src/agent/subagents/manager.ts
```diff
@@ -1147,20 +1138,20 @@ async function executeSubagent(
...
       inheritedApiKey,
       inheritedBaseUrl,
       transcriptPath,
       inheritedChannelContext,
     });
 
-    // Optionally confine memory-mode subagents to an OS filesystem sandbox.
+    // Optionally confine parent-memory subagents to an OS filesystem sandbox.
     // Returns null (spawn unchanged) when disabled, not applicable, or no
     // backend is available on this host.
```
amelia-letta: When the sandbox backend is unavailable (e.g. Linux without bwrap), parent-memory subagents fall back to `standard` permission mode with no filesystem write restrictions beyond the cross-agent guard. The old `memory` mode's static path checks (write target validation, shell scoping) served as a defense-in-depth fallback for this case.

devanshrj: Should be resolved with bundled bwrap

carenthomas (approved):

PR #3106, opened 2026-06-25
feat(tools): guard MessageChannel against no-op calls

overlord-letta:
## Problem

Agents can call `MessageChannel` with no message text, no media, and no reaction — silently sending an empty message to the channel. This was causing Kian's Telegram channel to receive blank messages.

## Change

Added `validateMessageChannelNoOp()` which rejects calls that would produce no output:

- `send` with no message text and no media → error
- `react` with no emoji → error
- `upload-file` with no media path → error

The check runs after input normalization but before registry/scope resolution, so the agent gets an immediate, actionable error string without needing the channel system to be initialized.

Input normalization already trims whitespace via `firstNonEmptyString`, so a message of `"   "` is treated as absent.

## Validation

- 5 new test cases covering each no-op scenario plus a positive case (media-only send is allowed)
- All 12 tests in `message-channel.test.ts` pass
- Biome lint clean on changed files
- Pre-commit hooks (layer boundaries, filename casing, type-check) passed

## Co-author

Cameron requested this from the bug-bash Slack thread after Kian reported empty Telegram messages.

### DIFF
--- a/src/channels/message-channel.test.ts
+++ b/src/channels/message-channel.test.ts
@@ -1189,7 +1189,10 @@ describe("MessageChannel", () => {
       },
     });
 
-    expect(result).toBe("Error: Slack send requires message or media.");
+    // The no-op guard catches this before the Slack plugin's own validation,
+    // since the legacy `text` alias is ignored and `message` is absent.
+    expect(result).toContain("Error");
+    expect(result).toContain("requires a non-empty message or media");
     expect(sendMessage).not.toHaveBeenCalled();
   });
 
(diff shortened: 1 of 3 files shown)


devanshrj:
Thanks for putting this together. I checked current `main`, and the underlying Telegram no-op cases are already handled in the channel action layer — specifically by #1780, which added Telegram validation for sends without message/media, uploads without media, and reactions without required inputs.

Since we do not want to add a second generic validation layer just as defense-in-depth, I am going to close this as superseded. Also, as-is this would subtly change existing reaction-removal behavior for some channels (`remove: true` without an emoji), so it is not purely redundant.

PR #3108, opened 2026-06-25
feat(mods): expose mod slash commands in the desktop listener

carenthomas:
## Summary
Mod-registered slash commands worked in the TUI but were silently dropped in the desktop/app-server (listener) path. This wires them through end to end:

- **Enable**: flip the mod `commands` capability on for the listener so `letta.commands.register(...)` actually registers (it was a no-op).
- **Advertise**: `DeviceStatus.supported_commands` now includes registered mod command ids (built-in list is returned unchanged when there are no mod commands, preserving the per-status allocation optimization).
- **Execute**: `handleExecuteCommand` routes unknown built-ins to the mod registry and runs them, mirroring the TUI mapping: `output` shows as command output, `handled` closes silently, `prompt` injects a user turn via `handleIncomingMessage`.
- **Reload**: re-push device status after `/reload` so newly (un)registered commands appear in the palette.

Command runtime context is built the same way as the TUI (`createModConversationHandle` + the shared `sendMessageStreamWithBackend`), so `fork`/`send`/`updateLlmConfig` work across local and Constellation backends.

## Not included (follow-up)
- Command **descriptions/args** are not surfaced yet — `supported_commands` is names-only. Rich palette rendering needs a client-side (letta-cloud) change to read a new metadata field.
- The desktop client must send a `command_id` it sees in `supported_commands`. If letta-cloud has a hardcoded command allowlist beyond `supported_commands`, it will need a matching change for these to appear/execute.

## Test plan
- [x] `bun test src/websocket/listener/` (96 pass)
- [x] New `mod-commands.test.ts`: advertise/lookup accessors + run output & prompt results
- [x] Updated `mod-adapter.test.ts`: listener now registers mod commands
- [x] typecheck + lint clean
- [ ] Validate in Desktop: register a mod command, confirm it shows in the palette, executes, and refreshes after `/reload`

### DIFF
--- a/src/websocket/listener/commands.ts
+++ b/src/websocket/listener/commands.ts
@@ -26,6 +26,7 @@ import {
   gatherInitGitContext,
 } from "@/cli/helpers/init-command";
 import { getReflectionSettings } from "@/cli/helpers/memory-reminder";
+import { buildModCommandPrompt } from "@/cli/mods/command-runtime";
 import {
   DEFAULT_SUMMARIZATION_MODEL,
   SYSTEM_REMINDER_CLOSE,
@@ -33,6 +34,7 @@ import {
 } from "@/constants";
 import { goalLoopMode } from "@/goal-loop-mode";
 import { runPreCompactHooks } from "@/hooks";
+import type { ModCommand } from "@/mods/types";
 import { settingsManager } from "@/settings-manager";
 import { trackBoundaryError } from "@/telemetry/error-reporting";
 import type {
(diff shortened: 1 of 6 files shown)


carenthomas:
closing in favor of a reworked approach.

the core problem with this PR: it advertises mod commands by merging their ids into `supported_commands`. that conflates two sets the client needs to keep apart:

- **A** = built-in remote commands the desktop client intentionally does *not* surface (`/remember`, `/goal`, `/toolset`, `/channels`, ...). these are in `supported_commands` already and stay hidden because desktop has no hardcoded entry for them.
- **Z** = mod commands, which the client *should* auto-render (it cannot pre-know them).

once mod ids are merged into `supported_commands` as bare strings, A and Z become indistinguishable, so a client cannot apply the one policy that matters: "allowlist built-ins, auto-show mods."

design conclusion from the discussion: the listener should stay **client-agnostic** and advertise *facts*, while **curation stays per-client** (it is a per-client UX decision and does not scale to push into the listener). that means mod commands need their own advertised facts with metadata (`{ id, description, args }`), separate from `supported_commands`, so any client can render them by its own policy. invocation reuses the existing `execute_command` path (already routed to the mod registry here).

the enable + execute + reload-repush work here is still correct and will carry over; only the advertising channel changes. reopening as a fresh PR with the dedicated `mod_commands` advertisement.

PR #3111, opened 2026-06-26
feat(channels): add /letta raw injection escape hatch

overlord-letta:
## Summary

Adds a `/letta` prefix that lets channel users (Slack, Telegram, Discord) inject text directly into the listener as if typed into the TUI command input.

- `/letta /reload` — dispatches `/reload` through the listener's `execute_command` path (same as the web UI)
- `/letta /compact all` — forwards `/compact all` with args
- `/letta hello` — strips the prefix and sends `hello` as a regular user message to the agent

## Why

While the team pushes Mods updates through channels, there's no way to trigger listener-level commands (like `/reload` for refreshing settings/mods) from a channel without switching to the TUI. The `/letta` escape hatch bridges that gap — it's a raw injection prefix that forwards whatever follows it to the listener, bypassing channel command interpretation.

## How it works

1. `parseLettaEscapeHatch()` in `src/channels/commands.ts` detects the `/letta` prefix (case-insensitive, bot-suffix-aware) and returns the stripped text
2. `ChannelRegistry.handleInboundMessage()` intercepts `/letta` before `tryHandleChannelSlashCommand`:
   - If the stripped text starts with `/`, it resolves the route and calls the `executeCommandHandler` (wired from `lifecycle.ts`), which parses the command and calls `handleExecuteCommand`
   - If the stripped text does not start with `/`, it mutates `msg.text` to the stripped value and falls through to the normal delivery path (route lookup, formatting, agent turn)
3. `lifecycle.ts` wires `setExecuteCommandHandler` with a closure that calls `handleExecuteCommand` and returns the output text as a direct channel reply
4. `handleExecuteCommand` return type changed from `Promise<void>` to `Promise<string>` so the channel handler can capture the output text and send it back

## Before/after

**Before:** `/reload` from a channel returns "not supported in channels yet." There is no way to trigger listener-level commands from a channel.

**After:** `/letta /reload` returns "Reloaded settings, local mods, and agent secrets" as a direct channel reply. `/letta hello` sends `hello` to the agent as a normal message.

## Limits / tradeoffs

- Commands that trigger agent turns via `handleIncomingMessage` (like `/doctor`, `/init`, `/remember`) will stream their output to the web UI but not to the channel — only the summary text is returned as a direct reply. This is a known limitation of the raw injection approach; the primary use case (`/reload`) works fully.
- `/letta` with no text returns a usage hint.
- Unknown commands return "Unknown command: X" from `handleExecuteCommand`.

## Validation

- `tsc --noEmit` passes
- `madge --circular` — no new circular dependencies
- `bun test src/channels/commands.test.ts` — 14/14 pass (10 existing + 4 new `parseLettaEscapeHatch` tests)
- Pre-commit hooks (biome, tsc, layer boundaries, exported functions, filename casing) all pass


src/channels/commands.ts
```diff
@@ -26,6 +26,14 @@ export type ChannelSlashCommandHandlerResult = {
   text?: string;
 };
 
+export type LettaEscapeHatchHandler = (params: {
+  text: string;
+  runtime: {
+    agent_id: string;
+    conversation_id: string;
+  };
+}) => Promise<ChannelSlashCommandHandlerResult>;
```
amelia-letta: `LettaEscapeHatchHandler` is exported but never used — the actual handler type is `ChannelExecuteCommandHandler` in registry.ts with a different return shape. Dead code.

overlord-letta: Good catch — removed. The actual handler type is `ChannelExecuteCommandHandler` in `registry.ts`, which `lifecycle.ts` wires via `setExecuteCommandHandler`. Pushed in e09da0a7.

overlord-letta:
Superseded by a narrower current-main channel command path.

PR #3130, opened 2026-06-27
feat(tools): support exec_command descriptions

just-cameron:
## Summary

- Add an optional `description` field to Codex `exec_command`, matching the existing optional Bash and Gemini shell description shape.
- Prefer shell descriptions in tool row display while preserving raw commands in approval previews for transparency.
- Carry exec command descriptions into later `write_stdin` rows so background-terminal follow-ups stay readable.

## Test plan

- `bun test src/tools/codex-unified-exec-toolset.test.ts src/cli/format-args-display.test.ts src/cli/helpers/unified-exec-accumulator.test.ts`
- `bun run typecheck`
- `bun run check`

### DIFF
--- a/src/tools/descriptions/ExecCommand.md
+++ b/src/tools/descriptions/ExecCommand.md
@@ -2,7 +2,7 @@ Runs a command in a PTY, returning output or a session ID for ongoing interactio
 
 For ordinary one-shot commands, omit `yield_time_ms` and let the default wait for completion; set `yield_time_ms` only when intentionally returning early from a long-running or interactive command.
 
-Use the optional `description` field for a clear, concise user-facing description of what the command does. Describe the purpose, not the shell syntax. This may be shown directly in chat as part of a status row like `Running command: <description>` or `Ran command: <description>`. For simple commands, keep it brief (5-10 words). For commands that are harder to parse at a glance, add enough context to clarify what it does.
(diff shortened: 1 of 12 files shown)


kl2806 (approved):

src/tools/descriptions/ExecCommand.md
```diff
@@ -2,6 +2,8 @@ Runs a command in a PTY, returning output or a session ID for ongoing interactio
 
 For ordinary one-shot commands, omit `yield_time_ms` and let the default wait for completion; set `yield_time_ms` only when intentionally returning early from a long-running or interactive command.
 
+Use the optional `description` field for a clear, concise user-facing description of what the command does. Describe the purpose, not the shell syntax. This may be shown directly in chat as part of a status row like `Running command: <description>` or `Ran command: <description>`. For simple commands, keep it brief (5-10 words). For commands that are harder to parse at a glance, add enough context to clarify what it does.
```
cpacker: @just-cameron do we want to drop this part if we're gonna actually just raw expose it?

Or should we align it with how it's acutally shown if we're not changing it? eg `Ran ___` not `Ran command: `

cpacker (changes requested):

src/tools/descriptions/ExecCommand.md
just-cameron: I think this is actually good guidance because it implies tense-invariance, but perhaps there's something more direct to be done

just-cameron: Resolved

PR #3137, opened 2026-06-29
fix(models): split GLM 5.2 context presets

sarahwooders:
Makes the default `zai/glm-5.2` preset use the observed 200K-class limit and adds an explicit `zai/glm-5.2[1m]` preset for opt-in long context. This avoids suppressing compaction for the default direct GLM path while keeping the 1M variant available.

Validated with:
- `bun run check`

### DIFF
--- a/src/models.json
+++ b/src/models.json
@@ -1664,7 +1664,7 @@
     },
     {
       "id": "glm-5.2-1m",
-      "handle": "zai/glm-5.2[1m]",
+      "handle": "zai/glm-5.2",
       "label": "GLM-5.2 1M",
       "description": "zAI's latest reasoning and coding model with 1M context",
       "isFeatured": false,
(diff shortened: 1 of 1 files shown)


src/models.json
```diff
@@ -1653,9 +1653,22 @@
...
       "free": true,
+      "updateArgs": {
+        "context_window": 200000,
+        "max_output_tokens": 131072,
+        "parallel_tool_calls": true
+      }
+    },
+    {
+      "id": "glm-5.2-1m",
+      "handle": "zai/glm-5.2[1m]",
```
amelia-letta: The handle `zai/glm-5.2[1m]` is not a valid Letta API model handle — the API only has `zai/glm-5.2` (with `max_context_window: 1000000`). This variant will be filtered out of the model selector when connected to the API and fail at agent update time with "Handle not found". Both presets should use the same `zai/glm-5.2` handle with different `context_window` values in `updateArgs` — the existing `getModelInfoForLlmConfig` context_window narrowing already handles disambiguation.

sarahwooders: Fixed in 6d6d4c5d — the 1M preset now reuses `zai/glm-5.2` and only differs by `id` / `context_window`.

PR #3145, opened 2026-06-29
fix(tools): clarify system memory path guidance

kl2806:
## Summary
- reword Edit and Write tool descriptions to avoid suggesting literal `$MEMORY_DIR/system` paths
- clarify that system memory files already shown in the prompt do not need a Read call
- direct agents to use relative memory-repo paths like `system/human.md` or expanded absolute paths

## Testing
- pre-commit checks ran during `git commit` (including TypeScript `tsc --noEmit`)

### DIFF
--- a/src/tools/descriptions/Edit.md
+++ b/src/tools/descriptions/Edit.md
@@ -3,7 +3,7 @@
 Performs exact string replacements in files. 
 
 Usage:
-- You must use your `Read` tool at least once in the conversation before editing. This tool will error if you attempt an edit without reading the file. Exception: files under `$MEMORY_DIR/system/` do not require a Read call first, since their contents are already provided in your system prompt.
+- You must use your `Read` tool at least once in the conversation before editing. This tool will error if you attempt an edit without reading the file. Exception: system memory files whose contents are already shown in your prompt do not require a Read call first. When passing a `file_path`, use a relative path such as `system/human.md` (from the memory repo) or an expanded absolute path; environment variables are not expanded.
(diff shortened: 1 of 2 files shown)


src/tools/descriptions/Edit.md
```diff
@@ -3,7 +3,7 @@
 Performs exact string replacements in files. 
 
 Usage:
-- You must use your `Read` tool at least once in the conversation before editing. This tool will error if you attempt an edit without reading the file. Exception: files under `$MEMORY_DIR/system/` do not require a Read call first, since their contents are already provided in your system prompt.
+- You must use your `Read` tool at least once in the conversation before editing. This tool will error if you attempt an edit without reading the file. Exception: system memory files whose contents are already shown in your prompt do not require a Read call first. For memory files, pass an expanded absolute path (use Bash to resolve `$MEMORY_DIR` first if needed).
```
devanshrj, line 6: > For memory files, pass an expanded absolute path (use Bash to resolve `$MEMORY_DIR` first if needed).
oh are we still updating the tools to expand vars then?

devanshrj (approved):

src/tools/descriptions/Edit.md
kl2806: Yes, #3146 should allow the env vars for execution. I was thinking originally that the prompt could be less permissive than the actual tools similar to how schema specifies absolute path in description but relative paths are still accepted
```  "properties": {
      "file_path": {
        "description": "The absolute path to the file to modify",
        "type": "string"
      },
```

PR #3146, opened 2026-06-29
feat(tools): expand ~ and $VAR in file_path for Read/Write/Edit

kl2806:
## Summary
- Add `src/utils/file-path.ts` with `expandFilePath()` that expands `~` (home dir) and `$VAR`/`${VAR}` env vars before resolving paths
- Wire it into `Read`, `Write`, and `Edit` tool implementations, replacing the inline `path.isAbsolute ? ... : path.resolve(cwd, ...)` pattern
- Unresolved variables are left as-is so error messages stay readable

Allows agents to pass paths like `$MEMORY_DIR/system/human.md` directly to file tools instead of needing to pre-expand via Bash.

## Test plan
- `bun run check` passes (all 9 checks)

### DIFF
--- a/src/utils/file-path.ts
+++ b/src/utils/file-path.ts
@@ -1,0 +1,29 @@
+import * as os from "node:os";
+import * as path from "node:path";
+
+/**
+ * Expand a file_path argument before resolving it:
+ * 1. Expand leading `~` to the home directory.
+ * 2. Expand `$VAR` and `${VAR}` references using process.env (fallback: leave
+ *    the token as-is so the downstream error message is still readable).
+ * 3. Resolve relative paths against `userCwd`.
+ */
+export function expandFilePath(filePath: string, userCwd: string): string {
+  // 1. Tilde expansion
+  let expanded = filePath;
+  if (expanded === "~" || expanded.startsWith("~/")) {
+    expanded = path.join(os.homedir(), expanded.slice(1));
+  }
+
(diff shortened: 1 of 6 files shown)


src/utils/file-path.ts
```diff
@@ -0,0 +1,29 @@
...
+import * as path from "node:path";
+
+/**
+ * Expand a file_path argument before resolving it:
+ * 1. Expand leading `~` to the home directory.
+ * 2. Expand `$VAR` and `${VAR}` references using process.env (fallback: leave
+ *    the token as-is so the downstream error message is still readable).
+ * 3. Resolve relative paths against `userCwd`.
+ */
+export function expandFilePath(filePath: string, userCwd: string): string {
```
amelia-letta, line 11: `multi-edit.ts` (line 32) still uses the old `path.isAbsolute ? ... : path.resolve(userCwd, file_path)` pattern — an agent that learns `$VAR` works in Edit will try it in MultiEdit and get a confusing "file not found". Worth wiring in here too since it is the direct sibling tool.

kl2806:
Addressed Amelia’s MultiEdit comment in 7cebe322: MultiEdit now uses the shared `expandFilePath()` helper too, and I added a regression test for `$VAR` expansion in MultiEdit.

Validation: `bun test src/tools/multiedit.test.ts` and `bun run check` pass.

PR #3148, opened 2026-06-29
feat: sync attached repositories

AriWebb:
Keeps hosted repositories attached to an agent cloned next to the memory repo, includes them in the system prompt projection, and syncs clean pending repository commits after turns.

Removes the agent-facing create_repository tool; agents only see attached repository locations.

Checks:
- bun run typecheck


src/agent/memory-git.ts
```diff
@@ -1562,6 +1581,125 @@ export async function initializeLocalMemoryRepo(
...
+    mkdirSync(args.directory, { recursive: true });
+    await runGitWithRetry(
+      args.directory,
+      ["clone", args.remoteUrl, "."],
+      args.token,
+      {
+        operation: `clone repository ${args.repositoryName}`,
+        timeoutMs: GIT_CLONE_TIMEOUT_MS,
+      },
+    );
```
amelia-letta: If the clone fails after `mkdirSync` creates the directory, the empty dir blocks all retries — the next call hits the "mount path already exists and is not a git repository" throw. Should clean up on failure.

```suggestion
    mkdirSync(args.directory, { recursive: true });
    try {
      await runGitWithRetry(
        args.directory,
        ["clone", args.remoteUrl, "."],
        args.token,
        {
          operation: `clone repository ${args.repositoryName}`,
          timeoutMs: GIT_CLONE_TIMEOUT_MS,
        },
      );
    } catch (err) {
      rmSync(args.directory, { recursive: true, force: true });
      throw err;
    }
```

src/agent/memory-git.ts
```diff
@@ -1898,6 +2190,288 @@ async function getMemoryAheadBehind(
...
+export async function syncPendingRepositoryCommitsAfterTurn(
+  agentId: string,
+): Promise<RepositoriesPostTurnSyncResult> {
+  const { getBackend } = await import("@/backend");
+  const backend = getBackend();
+  const localOnly =
+    backend.capabilities.localMemfs && !backend.capabilities.remoteMemfs;
+  const repositories = await listAttachedAgentRepositories(agentId);
+  const token = await getAuthToken();
+  const results = await Promise.all(
```
amelia-letta: `Promise.all` here means one repo throwing (e.g. corrupt repo, git status failure) rejects the entire batch and suppresses all other repos results. `syncAttachedAgentRepositories` uses `Promise.allSettled` for the same pattern — this should too.

src/reminders/memory-git-sync.ts
```diff
@@ -80,12 +134,17 @@ export async function runPostTurnMemorySync(
 
   try {
     const syncResult = await syncPendingMemoryCommitsAfterTurn(params.agentId);
-    const syncReminder = formatMemoryPostTurnSyncReminder(syncResult);
-    if (!syncReminder) {
-      return;
+    const repositorySyncResult = await syncPendingRepositoryCommitsAfterTurn(
```
amelia-letta: If `syncPendingRepositoryCommitsAfterTurn` throws (e.g. API call to list repos fails), the already-computed memory sync reminder is never enqueued — the entire try block bails to catch. Memory and repo syncs should be independent so a repo sync failure does not suppress the memory reminder.

PR #3158, opened 2026-06-30
feat(mods): add turn_start cancellation

carenthomas:
## Summary
- add an explicit `turn_start` `{ cancel: { reason } }` effect for trusted mods
- surface cancellation before backend/model send in TUI, headless, and listener/Desktop paths
- normalize/truncate cancel reasons and preserve first valid cancel reason
- document the new event effect and add engine/listener coverage

## Tests
- `bun run check`
- `bun test src/websocket/listen-client-concurrency.test.ts`

### DIFF
--- a/src/cli/app/use-conversation-loop.ts
+++ b/src/cli/app/use-conversation-loop.ts
@@ -699,6 +699,7 @@ export function useConversationLoop(ctx: ConversationLoopContext) {
           });
           buffersRef.current.order.push(statusId);
           refreshDerived();
+          userCancelledRef.current = false;
           return;
         }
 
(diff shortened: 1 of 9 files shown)


src/cli/app/use-conversation-loop.ts
```diff
@@ -685,6 +690,18 @@ export function useConversationLoop(ctx: ConversationLoopContext) {
...
+        if (turnStartCancelReason) {
+          const statusId = uid("status");
+          buffersRef.current.byId.set(statusId, {
+            kind: "status",
+            id: statusId,
+            lines: [turnStartCancelReason],
+          });
+          buffersRef.current.order.push(statusId);
+          refreshDerived();
+          return;
```
amelia-letta, line 703: early return skips the `userCancelledRef.current = false` reset below — if the user pressed ESC during the `turn_start` await, the stale flag cancels the next turn too.

```suggestion
        userCancelledRef.current = false;
        return;
```

carenthomas: fixed in b004e258, the TUI turn_start cancel branch now clears userCancelledRef before returning.

PR #3171, opened 2026-07-01
fix: reuse an existing agent when default agent creation hits the account limit

### LINKED ISSUE #3167: CLI tries to create a default agent instead of reusing existing agent when stale local session hits agent quota
Letta Code can get stuck on startup if `~/.letta/settings.json` has a stale `lastSession`/`sessionsByServer` agent ID and the account is already at its agent limit. There's a perfectly usable existing agent in the account, but startup decides it needs to create a new default one anyway, and creation fails with a 402 because the plan limit is already hit.

Version: Letta Code 0.27.14 (global npm install), Windows/PowerShell, Node v25.8.1, Letta Cloud backend.

What I saw:
```
Failed to create default agent: Failed to create default agents: 402 {"error":"You have reached your limit for agents, please upgrade your plan or delete some agents","limit":3}

Letta Code v0.27.14
Constellation
Loading pinned agents...
```
(There was also an unrelated Node deprecation warning printed just before this.)

My account had exactly 3 remote agents (`Architectural Advisor`, `Dependency Auditor`, `Letta Code`). Locally, `lastAgent` pointed at the valid `Letta Code` agent, but `lastSession.agentId` and `sessionsByServer["api.letta.com"].agentId` both pointed at a stale ID that no longer existed, there were no pinned agents, and `agents[]` had a few other IDs that were also gone remotely. `letta --info` showed "Will resume: (will create new agent)" / "Pinned agents: (none)" — which is exactly the path that hits the 402.

Feels like startup shouldn't immediately jump to creating a new default agent if `lastSession` is stale but there's still a valid `lastAgent` or existing agent around. A few options that would all fix it: fall back to `lastAgent`, show the profile/agent selector, or at least print a message that identifies the stale session and points at `letta --agent <id>` / `/pin` / `/profile save`. At minimum, a 402 here shouldn't be presented as "failed to create default agents" when a valid agent already exists.

Workaround that got me unstuck: back up `~/.letta/settings.json`, manually pin the existing agent in `agents[]` (`{"agentId": "agent-2a50...", "memfs": true, "systemPromptPreset": "custom", "pinned": true}`), and clear out `lastSession`/`sessionsByServer`. After that `letta --info` showed the selector instead of trying to create a new agent.

This is especially painful on low agent-limit plans — stale local metadata alone can make the CLI unusable even though there's a perfectly good agent already there, and right now recovering means hand-editing settings.json.

Coldaine:
Fixes #3167.

When an account is already at its agent limit, startup would still try to create a new default agent, fail with a 402, and exit — even though the account already has a usable Letta Code agent sitting right there. This mostly shows up when local settings point at a stale/missing `lastSession` agent ID, so the CLI decides it needs to create a new one instead of noticing an existing agent is available.

Now, when default-agent creation fails specifically because of an agent-limit/quota error, the CLI lists the account's existing agents, picks the first one, prints which agent it selected, and waits for a keypress before continuing — so the fallback is visible rather than silent. Any other create failure (network errors, auth issues, etc.) still exits immediately like before, and if listing agents also fails or comes back empty, it still hard-fails rather than pretending everything's fine. The acknowledgement prompt only shows up in an interactive TTY session; non-interactive runs skip straight through.

**Testing**
- `bun test src/startup-flow.test.ts` — fallback wiring is in place, and quota detection stays narrow (doesn't accidentally swallow unrelated errors like generic 429s)
- `bun test src/startup-setup-pty.test.ts` — end-to-end PTY coverage for the fallback path, the "no agents available" case, the "listing failed" case, and confirming non-quota failures still hard-fail (Windows skips the existing PTY cases)
- `cmd /c bun run check`

---
Authored by Claude, Codex, and Opencode. Reviewed by a real human too.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -538,6 +538,53 @@ function paginatedItems<T>(page: unknown): T[] {
   return [];
 }
 
+function errorMessage(error: unknown): string {
+  return error instanceof Error ? error.message : String(error);
+}
+
+function isDefaultAgentLimitCreateFailure(error: unknown): boolean {
+  const message = errorMessage(error).toLowerCase();
+  const status = (error as { status?: unknown } | null)?.status;
+  const statusCode = typeof status === "number" ? status : undefined;
+
+  if (message.includes("agents-limit-exceeded")) return true;
+
+  return (
+    message.includes("agent") &&
+    message.includes("limit") &&
+    (statusCode === 402 ||
+      statusCode === 429 ||
+      message.includes("402") ||
+      message.includes("429") ||
+      message.includes("you have reached your limit"))
+  );
+}
+
+function formatStartupFallbackAgent(agent: AgentState): string {
+  const name = agent.name?.trim();
+  return name ? `"${name}" (${agent.id})` : agent.id;
+}
+
+async function waitForStartupFallbackAcknowledgement(): Promise<void> {
+  if (!process.stdin.isTTY || !process.stdout.isTTY) return;
+
+  process.stdout.write("Press any key to continue...");
+  await new Promise<void>((resolve) => {
+    const wasRaw = process.stdin.isRaw;
+    const cleanup = () => {
+      process.stdin.off("data", onData);
+      process.stdin.setRawMode(wasRaw);
+      process.stdout.write("\n");
+      resolve();
+    };
+    const onData = () => cleanup();
+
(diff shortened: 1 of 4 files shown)


Coldaine:
image

Screenshot of typical presentation of issue.

PR #3029, opened 2026-06-22
feat(slack): add rich live progress cards

overlord-letta:
## Problem

Slack channel turns need immediate, useful progress without cluttering the thread. The progress card also needs to stay useful at completion: the header should show the final assistant reply, and detail rows should show the actual action that happened instead of generic labels like Completed or Running.

## What changed

This PR adds a richer Slack progress pipeline and supporting channel/tool metadata:

- Builds a shared sanitized channel-progress model for visible tool, command, approval, Skill, subagent, memory, web search, and fetch activity.
- Renders Slack native assistant stream/task cards for real progress, with stable per-tool rows and one active progress surface per thread run.
- Uses the final assistant reply as the completed Slack progress header instead of a generic Completed label.
- Adds structured tool row titles so detail-level rows can show actions such as Updated map.rs +2 -3, Failed to write lib.rs, Skill names, subagent prompts, web queries, and command descriptions.
- Avoids duplicating file paths/details once a richer row title is available.
- Requires model-facing shell command descriptions for Bash, exec_command, shell_command, and run_shell_command so Slack can display the user-facing intent rather than raw shell syntax.
- Keeps Slack task previews stable once rendered, and waits for streamed shell descriptions instead of showing provisional command text from incomplete JSON.
- Suppresses noisy or misleading progress rows for internal delivery tools, unknown tool fallbacks, transient reasoning/thinking deltas, and duplicate lifecycle reactions.
- Handles Slack stream edge cases: close failed streams before fallback, treat already-closed stopStream responses as benign, and avoid fallback-card duplication when native streams are active.
- Preserves explicit Slack DM thread routing for replies, slash commands, control responses, and immediate status updates.
- Adds a "Open in Letta Chat" footnote link to the bottom of completed progress cards, linking to the specific agent+conversation at app.letta.com. Local-backend agents (agent-local-*) are skipped since they don't exist in the web app.
- Changes bare Slack @agent /model to print the current routed model, while keeping the model selector available at @agent /model list and preserving @agent /model <handle-or-id> for switching.

The assistant reply path is unchanged. The main product behavior changed here is Slack live progress/status rendering plus the small supporting metadata needed to make those rows readable.

## Validation

Local validation:

- bun test src/channels/progress.test.ts src/channels/slack-adapter.test.ts
- bun test src/tools/codex-unified-exec-toolset.test.ts src/tools/shell-secrets.test.ts src/tools/exec-command.test.ts src/tools/run-shell-command.test.ts
- bun run lint
[...]

PR #3177, opened 2026-07-01
fix(slack): preserve progress cards across continuations

overlord-letta:
## Summary

Fixes a rich Slack progress-card lifecycle bug in the live feedback branch. Completed channel lifecycle events can happen at continuation or compaction boundaries; treating them as terminal stopped the native Slack stream and cleared the active card key. The next tool progress then started a second visible card in the same thread.

This changes completed lifecycle handling to flush pending progress and clear assistant thread status without closing the card. Final non-reaction Slack sends still close the active card, and error or cancelled lifecycle outcomes still stop the stream.

## Validation

- bun run check
- bun test src/channels/slack-adapter.test.ts


overlord-letta:
Folded into #3029.

PR #3178, opened 2026-07-01
feat(channels): support Slack mention bang commands

overlord-letta:
## Summary
- Handle Slack app-mention bang commands through the shared channel command router: `!help`, `!detach`, `!model`, `!new`, and `!reload`.
- Add a Slack-only detached route state so normal unmentioned thread replies stay silent until a later app mention reattaches the thread.
- Wire `!new` to replace the Slack thread with a fresh conversation route, and wire `!reload` to the existing listener reload path.

## Why
Slack app mentions already normalize away the bot mention before channel routing. Without a bang-command path, operational messages like `!detach` were delivered as model input instead of being handled by the listener.

## Notes
- Bang parsing is only enabled for Slack messages marked as app mentions.
- Unmentioned bang-looking text remains normal routed input unless the thread is already detached.
- `/detach`, `/new`, and `/reload` are not exposed as general channel slash commands.

## Validation
- `bun test src/channels/commands.test.ts src/channels/slack-registry.test.ts`
- `bun run typecheck`
- `bun run check`

### DIFF
--- a/src/channels/commands.test.ts
+++ b/src/channels/commands.test.ts
@@ -7,18 +7,21 @@ import {
   buildChannelCancelUnavailableMessage,
   buildChannelChatLinkMessage,
   buildChannelChatUnavailableMessage,
+  buildChannelDetachedMessage,
   buildChannelHelpMessage,
   buildChannelModelListMessage,
   buildChannelModelListUnavailableMessage,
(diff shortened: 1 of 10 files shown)


overlord-letta:
Moved into #3029 per reviewer direction; closing this duplicate delivery PR.

PR #3180, opened 2026-07-01
fix(providers): always emit is_oauth for chatgpt_oauth in connect provider response

carenthomas:
buildConnectProviderEntries() only set is_oauth=true when provider.isOAuth was truthy. A chatgpt_oauth provider can be constructed via byokProviderFromPiSpec() rather than localOAuthProviderConfigs(), and that path does not set isOAuth. When is_oauth is absent from the response, the letta-cloud desktop UI falls back to connectorKind=apiKey and shows a coming-soon banner instead of the real OAuth flow.

Fix: also emit is_oauth=true when provider_type is chatgpt_oauth, so the protocol field is always correct regardless of which code path constructed the ByokProvider.

Companion source fix to letta-cloud PR #12699.

### DIFF
--- a/src/providers/connect-provider-service.ts
+++ b/src/providers/connect-provider-service.ts
@@ -260,7 +260,9 @@ export function buildConnectProviderEntries(
       provider_type: provider.providerType,
       provider_name: provider.providerName,
       provider_names: uniqueProviderNames(provider),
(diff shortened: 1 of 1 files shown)


carenthomas:
closing in favour of a root cause fix in pi-provider-registry.ts

PR #3194, opened 2026-07-02
fix(channels): add reload slash command

just-cameron:
## Summary
- add `/reload` to supported channel slash commands
- wire channel `/reload` to the existing listener reload path for settings, local mods, and agent secrets
- add channel command coverage for the new supported command/help text

## Tests
- `bun test src/channels/commands.test.ts src/channels/registry.test.ts`
- `bun run check`


overlord-letta:
Superseded by a clean current-main replacement.

PR #3202, opened 2026-07-03
refactor(memfs): make memfs mandatory — remove all non-memfs opt-outs

cpacker:
## Summary
- **Commit 1** (`9b865893`): stamps the `git-memory-enabled` tag at creation on raw protocol paths (listener/app-server agent creation) and self-heals untagged agents on Letta Cloud during listener memfs sync, so no creation path produces an untagged agent.
- **Commit 2** (`12fec3ca`): removes every user-facing way to create or run a non-memfs agent:
  - Deletes the `--no-memfs` CLI flag (interactive + headless) and the `LETTA_LOCAL_BACKEND_NO_MEMFS` env var (replaced by an internal process-level switch used only for stateless subagent processes)
  - Deletes the `/memfs disable` command and `reattachMemoryTool()`
  - Deletes `enableMemfs` from `CreateAgentOptions` and the import options; memfs is derived from backend capabilities + subagent role
  - Deletes the dead `incognito` default agent config
  - `applyMemfsFlags()` loses its disable param (action is now `"enabled" | "unchanged"`); `updateAgentSystemPromptMemfs()` only upgrades to the memfs prompt variant
- Subagents remain the sole non-memfs carve-out, derived from `LETTA_CODE_AGENT_ROLE=subagent` + `--new-agent` (not a CLI flag), so fork/recall subagents that deploy an existing memfs-tagged agent keep their memory.
- CI/test scaffolding that used `--no-memfs` for startup speed now uses `--memfs-startup skip` instead (agents stay memfs-enabled, just skip the pull).

## Test plan
- [x] `tsc --noEmit` clean
- [x] `bun run lint` clean
- [x] Unit tests: create/defaults/import/memory-filesystem/subagent-model-resolution/shell-env/args all pass (138 focused + broader `src/agent`, `src/cli`, `src/tools`, `src/websocket`, `src/mods` runs)
- [x] Remaining failures (`startup-flow` EPERM sandbox issues, `app-server` websocket, `pullMemory` recovery) verified pre-existing on base via stash-testing
- [ ] Manual: fresh agent creation on Cloud gets memfs tag + repo; `/memfs` shows enable/sync/reset only; subagent spawn skips memfs


src/websocket/listener/memfs-sync.ts
```diff
@@ -18,18 +18,36 @@ async function syncMemfsForAgent(agentId: string): Promise<void> {
   const agent = await getBackend().retrieveAgent(agentId);
```
amelia-letta: `retrieveAgent` without `include` strips `tags` from the API response — every agent on Letta Cloud will hit the repair path on first sync, triggering a full memfs enable (prompt update + recompile + tool detach) even when memfs is already set up. Every other callsite that needs tags passes `{ include: ["agent.tags"] }`.

```suggestion
  const agent = await getBackend().retrieveAgent(agentId, {
    include: ["agent.tags"],
  });
```

cpacker: Fixed in f2094b3e — `syncMemfsForAgent` now retrieves with `include: ["agent.tags"]`, and `addGitMemoryTag` also requests tags when it has to fetch so it can never write back an incomplete tag list. (This was also the root cause of the Prod × DeepSeek blank-slate incident.)

PR #3209, opened 2026-07-03
fix(sandbox): make the cross-agent shell sandbox opt-in, keep memory-subagent sandbox default-on

cpacker:
## Summary
- The default-on kernel shell sandbox (#2853) walled off `~/.letta/agents` for the primary agent's own shell commands, so agents hit kernel `Operation not permitted` errors accessing memfs/agent dirs even in **unrestricted** mode ([report](https://letta-ai.slack.com/archives/C07AFH896V7/p1782933882270059)). A kernel deny has no approve/deny escape hatch, and making it mode-toggleable would require unmounting/remounting the process sandbox — per the thread, the agreed fix is off-by-default with documented opt-in.
- Splits the single `LETTA_FS_SANDBOX` gate into two checks in `sandbox/availability.ts`:
  - **`isFsSandboxEnabled` (default ON)** — memory-subagent whole-process sandbox (reflection/memory/init/history-analyzer). These run non-interactively with a scoped write surface, so they keep kernel confinement: sandboxes for memory subagents only.
  - **`isShellSandboxEnabled` (default OFF)** — per-shell-command cross-agent confinement of agent shells (`sandbox-gate.ts` → `applyShellSandbox`). Opt in with `LETTA_FS_SANDBOX=1` for multi-tenant deployments (app server, experiment runners) where co-hosted agents must not read each other's memory.
- `LETTA_FS_SANDBOX` semantics: **unset** → memory subagents only · **`1`/`true`** → both · **`0`/`false`** → nothing.
- The in-process cross-agent guard (Read/Write/etc.) is unchanged; its header comment is updated to reflect that spawned shells are only kernel-confined when opted in.

Front-running @devanshrj's planned PR for this since it blocks the next Letta Desktop release — Devansh, please sanity-check the split.

## Test plan
- [x] New gate tests: shell sandbox not applied with flag unset / `0`, still applied with `1`
- [x] New flag tests: `isShellSandboxEnabled` opt-in semantics alongside existing `isFsSandboxEnabled` opt-out semantics
- [x] All sandbox/permissions suites green (149 tests across 10 files); `bun run check` 9/9

### DIFF
--- a/src/agent/subagents/sandbox.test.ts
+++ b/src/agent/subagents/sandbox.test.ts
@@ -11,6 +11,7 @@ import {
 } from "@/permissions/sandbox-policy";
 import {
   isFsSandboxEnabled,
+  isShellSandboxEnabled,
   type SandboxAvailability,
 } from "@/sandbox/availability";
 import { SANDBOX_ENV_VAR } from "@/sandbox/policy";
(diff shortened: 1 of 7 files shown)


devanshrj:
LGTM! I'll do a larger refactor to clean up and only scope sandboxes to headless mode / subagents as well

PR #3211, opened 2026-07-03
fix(channels): support Telegram private bot topics

overlord-letta:
## Summary

Telegram private bot topics arrive as private-chat messages with a positive `chat_id` plus `message_thread_id`. The old path treated the topic id as a separate direct-route key, so topic traffic missed the existing root DM route. Outbound Telegram code also discarded thread ids for positive chat ids, so replies escaped the topic.

## What changed

- Added Telegram-only inbound route fallback: an exact private-topic route wins; a disabled exact private-topic route blocks fallback; otherwise a private-topic message reuses the existing root direct route.
- Preserved the inbound private-topic id on the turn source so `MessageChannel` can reply into the active private topic without reusing stale route-level thread ids for ordinary root DMs.
- Allowed Telegram outbound helpers, direct replies, lifecycle errors, and permission prompts to carry explicit `message_thread_id` values for private bot topics.
- Split reply anchoring from topic routing: `replyToMessageId` targets the user message, while `threadId` carries the Telegram topic or forum route.

## Changed behavior matrix

- Telegram private topic inbound plus only a root DM route: dispatches to the root route and records the topic id in `turnSources`.
- Telegram private topic inbound plus exact enabled topic route: uses the exact topic route.
- Telegram private topic inbound plus exact disabled topic route: drops instead of silently falling back to root.
- Telegram private-topic outbound from the active turn source: sends `message_thread_id`.
- Telegram root private DM outbound through an old direct route with a stored thread id: sends without `message_thread_id`.
- Telegram groups and forum topics: still preserve route or request thread ids.
- Non-Telegram channels: no route fallback behavior change; the fallback predicate is restricted to Telegram direct messages with a non-empty thread id.

## Limits

- This does not create one route per Telegram private bot topic. Private topics reuse the root DM route unless an exact topic route already exists.
- No live Telegram API smoke test was run; bad or stale topic ids still surface through the existing send failure path.

## Validation

- `bun run check` — 12/12 passed.
- `bun test ./src/channels/message-channel-telegram.test.ts ./src/tools/impl/message-channel.test.ts` — 19 passed.
- `bun test ./src/channels ./src/tools/impl/message-channel.test.ts ./src/websocket/listen-recovery.test.ts` — 806 passed, 1 skipped.
- Baseline-negative check: applying the final regression-test diff to `origin/main` produced 11 expected failures across the Telegram route, outbound, adapter, presentation, and control-response assertions.


overlord-letta:
@carenthomas quick nudge when you have a chance: this is green and waiting on review. The change is scoped to Telegram channel routing/reply behavior for private bot topics, with regression tests for root DMs and private-topic `message_thread_id` handling.

amelia-letta (commented):
`buildDirectReplyOptions` (used here and in registry-routes.ts) sets `replyToMessageId` to `msg.threadId ?? msg.messageId`, which means for private bot topics it uses the thread ID as the reply target. This creates a reply to the thread creation message instead of the user's message.

For "not on allowed list", pairing instructions, and unbound route instructions, you probably want to reply to the user's actual message (use `messageId`), not anchor to the thread root.

The function in `registry-presentation.ts` should likely be:
```typescript
replyToMessageId: msg.messageId ?? undefined,
```

Or if you want no reply anchor for these system messages, just omit `replyToMessageId` entirely and only pass `threadId`.

src/channels/registry-inbound.ts
```diff
@@ -304,6 +290,7 @@ export function createChannelInboundRouter(deps: {
         await adapter.sendDirectReply(
           msg.chatId,
           "You are not on the allowed users list for this bot.",
+          buildDirectReplyOptions(msg),
```
amelia-letta: `buildDirectReplyOptions` sets `replyToMessageId` to `threadId ?? messageId`, so for private bot topics this replies to the thread creation message instead of the user's message. For "not on allowed list", pairing instructions, and unbound route messages, you probably want to reply to the actual user message (use `messageId`), not anchor to the thread root.

just-cameron: Fixed in ad6ce322: `buildDirectReplyOptions` now anchors `replyToMessageId` to `msg.messageId` (the user's message) and keeps `threadId` only as the topic route, so system replies in private bot topics reply to the inbound message instead of the topic-creation message. The same correction applies to the pre-existing group call sites, where anchoring to the user's command message is also the better anchor; `message_thread_id`/`thread_ts` routing is unchanged. Added direct coverage in `registry-presentation.test.ts` and updated the control-request expectation in `registry-controls.test.ts`.

src/channels/telegram/utils.ts
```diff
@@ -132,17 +132,9 @@ export function resolveTelegramInputFileConstructor(
 }
 
 export function resolveTelegramOutboundThreadId(
-  msg: Pick<OutboundChannelMessage, "chatId" | "threadId">,
+  msg: Pick<OutboundChannelMessage, "threadId">,
 ): string | null {
```
amelia-letta: Removing the chatId sign check means `sendLifecycleErrorReply` and `handleControlRequestEvent` in adapter.ts now use `threadId ?? source.messageId` for `replyToMessageId`, which replies to the topic creation message instead of the user's message — same pattern fixed in `buildDirectReplyOptions`. Those two functions need the same fix: `source.messageId ?? undefined` for the reply anchor.

just-cameron: Fixed in b1fd1c4b: `sendLifecycleErrorReply` and `handleControlRequestEvent` now anchor `replyToMessageId` to `source.messageId ?? undefined`; `message_thread_id` still routes the reply into the topic. I also swept for the same pattern elsewhere: the registry control-request `sendDirectReply` fallback now uses `buildDirectReplyOptions(event.source)` so non-Telegram adapters get the same corrected anchor, and the remaining `threadId ?? messageId` sites are thread-identity/dedup keys (route creation, lifecycle reply dedup, typing targets), not reply anchors. Updated expectations in the lifecycle, typing, runtime, and listen-recovery tests.

just-cameron (approved):

sarahwooders (approved):

src/channels/registry-inbound.ts
```diff
@@ -36,6 +40,39 @@ import {
...
+    msg.chatId,
+    accountId,
+    msg.threadId,
+  );
+  if (exactRoute) return exactRoute;
+
+  const exactRawRoute = getRouteRaw(
+    msg.channel,
+    msg.chatId,
+    accountId,
```
amelia-letta: This fallback is only wired into the main inbound route lookup. `getStatusRoute` (line 137, same file) and the slash command routing in `registry-commands.ts` (`findRawRouteForMessage`, `getCancelRoute`) still use exact thread-id lookups without the fallback, so `/status`, `/pause`, `/resume`, and `/cancel` from a private bot topic will fail route lookup. The PR description says these are wired through the helper but the diff doesn't include those changes.

just-cameron: Fixed in `37d87165`. The rewritten branch now routes `/status` through `loadRouteForInboundMessage` in `registry-inbound.ts`, and `/pause`, `/resume`, and `/cancel` through the same shared helper in `registry-commands.ts`. `registry-command-routing.test.ts` covers private-topic pause/resume/status behavior. The branch was rebuilt after your comment.

PR #3212, opened 2026-07-03
feat(cli): add `letta dream` subcommand for non-interactive reflection

cpacker:
## Summary
- Adds a `letta dream` subcommand that runs a memory reflection pass for an agent outside of an interactive session and waits for it to finish. It is a thin CLI wrapper over the same launcher the `/reflect` slash command uses.
- Flags: `--agent <id>` (defaults to `$LETTA_AGENT_ID`, then the last-used agent), `--conv <id>` (defaults to `default`), `-i/--instruction <text>`, and `--json` for machine-readable output.
- Exit codes: 0 on success or when there is no new transcript content to process; non-zero on failure (memfs disabled, reflection already active, or the reflection run failing). Failure details are included in `--json` output.
- Purely additive: new subcommand + router wiring + tests. No changes to the reflection launcher or transcript machinery.

Verified end to end against an isolated local-backend storage dir: fresh agent → `no_payload` no-op exit 0, seeded transcript → reflection subagent runs, commits to the agent's memory repo, completion message and exit 0, and a re-run is an incremental no-op (cursor advanced).


src/cli/subcommands/dream.ts
```diff
@@ -0,0 +1,186 @@
...
+        return 1;
+      }
+    }
+  }
+
+  if (!asJson) {
+    console.log(`Processing transcript: ${result.payloadPath}`);
+  }
+
+  const outcome = await completion;
```
amelia-letta: If the reflection launcher's `onComplete` callback throws before reaching `onCompletionMessage` (e.g. telemetry or `finalizeAutoReflectionPayload` fails), `spawnBackgroundSubagentTask` swallows the error but `onCompletionMessage` is never called — so this `await completion` hangs forever. For a non-interactive CLI command that could be used in CI/scripts, consider adding a timeout or a fallback rejection path.

PR #3214, opened 2026-07-03
fix(cli): ignore legacy no-memfs launcher args

overlord-letta:
## Summary
- Treat stale `--no-memfs` launcher metadata as a hidden no-op during CLI arg preprocessing.
- Strip the same legacy flag from `LETTA_CODE_BIN_ARGS_JSON` before spawning child CLIs, so reflection subagents do not inherit a flag current CLIs reject.
- Cover top-level CLI parsing, shell launcher metadata, and subagent launcher inheritance.

## Test plan
- `bun test src/cli/args.test.ts src/tools/shell-env.test.ts src/agent/subagent-model-resolution.test.ts`
- `bun run check`

### DIFF
--- a/src/agent/subagent-model-resolution.test.ts
+++ b/src/agent/subagent-model-resolution.test.ts
@@ -48,6 +48,27 @@ describe("resolveSubagentLauncher", () => {
     });
   });
 
+  test("explicit launcher drops legacy no-memfs args", () => {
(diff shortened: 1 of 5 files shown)


overlord-letta:
Closing because LET-9436 is already being handled elsewhere.

PR #3215, opened 2026-07-03
fix(local): strip images for text-only model dispatch

overlord-letta:
## Summary
- Enforce model input capabilities before pi-ai builds local provider payloads, replacing unsupported user/tool image parts with text placeholders for text-only models.
- Add OpenRouter and ZAI GLM regression coverage for canonical images, legacy base64 image parts, image_url parts, and tool-result images.

Linear: LET-9437

## Test plan
- bun test src/backend/pi-stream-adapter.test.ts
- LETTA_FS_SANDBOX=0 bun run check

### DIFF
--- a/src/backend/dev/pi-stream-adapter.ts
+++ b/src/backend/dev/pi-stream-adapter.ts
@@ -180,6 +180,10 @@ function toPiTools(clientTools: unknown[]): Tool[] | undefined {
 }
 
 const EMPTY_TOOL_RESULT_PLACEHOLDER = "No result provided";
+const NON_VISION_USER_CONTENT_PLACEHOLDER =
+  "(image omitted: model does not support images)";
+const NON_VISION_TOOL_CONTENT_PLACEHOLDER =
+  "(tool image omitted: model does not support images)";
 
 type LocalUserContent = Extract<LocalMessage, { role: "user" }>["content"];
 type LocalAssistantContent = Extract<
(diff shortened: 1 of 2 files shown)


cpacker:
Closing this after an in-depth investigation. Summary: the invariant this PR targets is already owned by pi-ai and is not demonstrably broken in LC's dispatch path; the sanitizer added here duplicates pi-ai's logic at the wrong layer, its tests pin message shapes no LC producer can emit, and it would no-op in the one dispatch-path failure mode we *can* actually construct. Root-cause analysis of the underlying reports (LET-9437) points at the `image-understanding` mod, not core dispatch.

## 1. pi-ai already enforces this invariant

`transformMessages()` → `downgradeUnsupportedImages()` in pi-ai replaces every `type: "image"` part in user/toolResult messages with the exact `(image omitted: model does not support images)` / `(tool image omitted: ...)` placeholders whenever `model.input` lacks `"image"`. It runs on **all** relevant API paths (`openai-completions`, `anthropic`, `google-shared`, `amazon-bedrock`, `mistral`, `openai-responses-shared`).

Verified against both pi-ai **0.79.6** (what current releases ship, per #2957) and **0.80.3** (latest): all GLM 5.x text handles are cataloged `input: ["text"]` in both the `zai` and `openrouter` catalogs, and LC's resolution path (`resolvePiModelForAgent` → `getCatalogModel`) uses those entries. The pre-existing test this PR renames ("downgrades images through Pi-AI payload conversion...") already proved this works end-to-end.

This PR re-implements that downgrade in the adapter — same placeholder strings, same placeholder-dedup logic, gated on the same `model.input` check. That's a second, divergent copy of provider-runtime logic that pi-ai exists to own.

## 2. No LC producer emits the shapes the new tests pin

The added tests inject `image_url` parts and Anthropic-style `source` parts into `uiMessages` with `as never` (they do not typecheck otherwise). Tracing every real write path into the local transcript:

- `localContentFromInputContent` (user messages) and `toolResultContentFromUnknown` (tool results) in `local-store.ts` sanitize all inbound content to canonical shapes — text + canonical `{type:"image", mimeType, data}` — and **drop everything else**. This has been true since the local backend was introduced (#2065, refined in #2160/#2270).
- Channel attachments (`formatChannelNotification`), CLI/Desktop paste (`paste-registry` → Letta-style `source` parts), and tool-result images all flow through those sanitizers.
- Mod `turn_start` mutations feed back into `messages.create` and hit the same ingestion.
- `toPiMessages` reads only from the persisted store.

There is no code path that puts an `image_url` or `source`-shaped part into the pi `Context`. The tests pin fabricated shapes, which freezes speculative behavior rather than a real contract.

## 3. The PR no-ops on the one constructible dispatch failure

The only dispatch-path failure we could actually construct is **misdeclared capability metadata**: e.g. a provider mod registering a text-only model with `input: ["text", "image"]` (validation in `pi-provider-mod-validation.ts` requires the array but cannot verify truthfulness). In that case pi-ai trusts the declared caps, converts canonical images to `image_url`, and the provider 400s — and this PR's sanitizer, gated on the same `model.input`, does nothing.

## 4. The actual reported failures trace to the image-understanding mod

For the LET-9437 reports (text-only GLM 5.x via Z.AI / OpenRouter):

- **Pre-#38 mod behavior** (letta-ai/mods#38, now merged): `autoCaptionTurn` used `appendTextContent`, leaving original image parts in the turn, and `imageRefFromPart` didn't recognize base64 pastes at all. mods#38/#39 document and fix exactly this ("the original raw image parts were left in the message content ... causing downstream models without vision support to choke").
- **The mod's own vision request bypasses pi-ai entirely**: `askOpenAiCompatible` is a raw `fetch` to the user-configured `baseUrl` sending `{type:"image_url"}` parts. A GLM-only user who points the mod's vision backend at their text-only GLM endpoint gets verbatim `INVALID_ARGUMENT: content.type is invalid, allowed values: ['text']` — surfaced in-conversation, indistinguishable at a glance from a core dispatch failure.
- **Listener caveat**: `LISTENER_MOD_CAPABILITIES` has `events.turns: false`, so mod-side strip/caption can never run for listener-routed (e.g. Discord) agents; that surface needs a listener/core-level answer, but gated on real evidence of a dispatch-path leak, which we don't have yet.

## 5. If LC-side hardening is ever warranted

The right seam is `toPiMessage` (the boundary where LC hands messages to pi-ai), enforcing the canonical `LocalMessage` content type **independent of model capabilities** — drop/flatten anything non-canonical, then let pi-ai's capability logic do its job. That covers both non-vision models (downgrade) and vision models (where a non-canonical part currently becomes `data:undefined;base64,undefined` via pi-ai's accept-anything else-branch — arguably an upstream pi-ai bug worth filing). A capability-gated sanitizer in the adapter covers neither correctly.

Follow-ups tracked instead of this PR:
- mod-side: agent scoping/config hardening in letta-ai/mods (PR incoming)
- pending: failing transcript from the original reporters to confirm which mechanism produced their 400s

PR #3222, opened 2026-07-04
fix: curate the featured model list to current-generation models

cpacker:
Part of LET-9411 (model list curation, from Sarah's feedback that the hosted list shows too many models). Pairs with a letta-cloud PR that makes the desktop selector's "Letta Hosted" tab show only `letta/*` + featured entries.

## Summary
- Re-curates `isFeatured` in `models.json` to one current-generation entry per family: **added** Auto Fast, Letta GLM, DeepSeek V4 Pro; **removed** Opus 4.8 1M, Sonnet 4.6 1M, GPT-5.4, GPT-5.4 Mini, MiniMax 2.7
- Featured set now: Auto, Auto Fast, Auto Chat, Letta GLM | Fable 5, Opus 4.8, Sonnet 5 | GPT-5.5 | DeepSeek V4 Pro | GLM-5.2 | MiniMax M3 | Kimi K2.7 Code | Gemini 3.1 Pro, Gemini 3.5 Flash
- This drives the CLI `/model` short list today and the LCD "Letta Hosted" tab once the paired letta-cloud change lands (the flag already flows over the WS model list)
- Everything unfeatured stays selectable via the full list / All tab

Note: GPT-5.5 Fast was requested for the curated set but `openai/gpt-5.5-fast` is not currently served by the hosted models API (only the ChatGPT Plus/Pro BYOK variant exists), so it cannot be featured as Letta-hosted yet.

## Test plan
- [x] `bun run check`
- [x] Flag-sensitive tests: model-tier-selection, create-agent-request, listener-provider-fallback, listen-model-update, channels/commands
- [ ] CLI `/model` short list shows the curated set

### DIFF
--- a/src/agent/model-tier-selection.test.ts
+++ b/src/agent/model-tier-selection.test.ts
@@ -25,7 +25,20 @@ describe("getModelInfo", () => {
     expect(info?.handle).toBe("anthropic/claude-fable-5");
     expect(info?.label).toBe("Fable 5");
     expect(info?.updateArgs).toMatchObject({
-      context_window: 1000000,
+      context_window: 200000,
+      max_output_tokens: 128000,
+      enable_reasoner: true,
+      reasoning_effort: "high",
+      parallel_tool_calls: true,
+    });
+  });
+
+  test("resolves Fable 5 1M registry metadata", () => {
+    const info = getModelInfo("fable-1m");
+    expect(info?.handle).toBe("anthropic/claude-fable-5");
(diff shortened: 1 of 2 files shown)


cpacker:
Added in `efcb5165`: Fable 5 is now a split listing like Opus 4.8 — "Fable 5" (featured) at the standard 200k window, "Fable 5 1M" (unfeatured, 950k like the other 1M rows) in the long list. Previously all Fable rows carried the native 1M window, so it was the only family without the standard/long-context split.

PR #3228, opened 2026-07-05
Fix Desktop skill status for agent MemFS skills

adrianchan94:
## Summary
- surface the resolved client-side skill snapshot as `current_available_skills` in listener device status
- preload that snapshot on `runtime_start` so Desktop has agent MemFS skill paths before the first turn
- refresh the snapshot on turns when client skills are rebuilt

## Why
Desktop can show an agent-assigned MemFS skill but fail to render its modal when the matching `~/.letta/skills/<skill>` local catalog entry is missing. The runtime already resolves the correct skill set from project/global/agent/MemFS sources; this PR exposes that resolved snapshot through the existing device status channel instead of mutating the user local skills catalog.

## Notes
- This does not add a new skill-detail/read RPC; it only exposes availability/path metadata already resolved for the active runtime.
- Absolute skill paths are already local listener state and are scoped to the existing Desktop/listener protocol.
- Skill installs/deletes between turns will refresh on runtime_start or the next turn; a separate detail/read API can be added later if Desktop needs on-demand full skill content.

## Tests
- `bun test src/agent/client-skills.test.ts src/websocket/listener/protocol-outbound.test.ts src/websocket/listen-client-protocol.test.ts --test-name-pattern "(buildClientSkillsPayload|buildDeviceStatus|runtime_start)"`
- `bun run lint`
- `bun run typecheck`
- `bun run check`

### DIFF
--- a/src/agent/client-skills.test.ts
+++ b/src/agent/client-skills.test.ts
@@ -426,6 +426,13 @@ describe("buildClientSkillsPayload", () => {
       expect(result.skillPathById).toEqual({
         "shared-skill": join(memorySkillDir, "SKILL.md"),
       });
+      expect(result.availableSkills).toContainEqual({
+        id: "shared-skill",
+        name: "shared-skill",
+        description: "from memfs",
+        path: join(memorySkillDir, "SKILL.md"),
+        source: "agent",
+      });
     } finally {
       if (originalMemoryDir === undefined) {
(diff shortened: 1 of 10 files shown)


adrianchan94:
Closing this for now — opened prematurely while we decide the right upstream lane.

PR #3231, opened 2026-07-06
fix: background bash task completion notification

### LINKED ISSUE #3226: Background Bash task completion does not deliver task-notification to the agent
The Bash tool docs state that for `run_in_background` tasks the agent "will be notified when it finishes" via a `<task-notification>` message, and that polling is unnecessary. In an app-server/WS listener session (Letta Code 0.27.20, local backend), this notification never arrives:

**Repro observed:**
1. Agent launches `Bash` with `run_in_background: true` (long-running `gh run watch ...`)
2. Task completes while the agent is idle between turns (verified: output file written, exit 0)
3. No `<task-notification>` is delivered on completion — the agent is not woken
4. The notification is also not prepended/batched into the next user-triggered turn (user messaged 3 minutes after task completion; no notification content appeared)
5. `TaskOutput` on the task id returns `{status: "completed"}` fine when the agent explicitly checks

So the task registry knows the task finished, but the completion event never reaches the agent as a message. The agent (me) promised the user "the watcher will ping you when the deploy is done" based on the tool description, and the ping never happened — the user noticed before I did.

Possibly scoped to WS/listener sessions or the local backend; may work in the plain CLI. If notifications are intentionally not supported in some session types, the Bash tool description should say so, since agents plan around the promised wake-up.

jitendravjh:
Fixes #3226. Delivers task-notification on completion of background bash tasks.

### DIFF
--- a/src/tools/impl/bash.ts
+++ b/src/tools/impl/bash.ts
@@ -22,6 +22,8 @@ import { spawnWithLauncher } from "./shell-runner.js";
 import { applyShellSandbox } from "./shell-sandbox.js";
 import { LIMITS, truncateByChars } from "./truncation.js";
 import { validateRequiredParams } from "./validation.js";
+import { addToMessageQueue } from "@/utils/message-queue-bridge.js";
+import { formatTaskNotification } from "@/utils/task-notifications.js";
 
 // Cache the working shell launcher after first successful spawn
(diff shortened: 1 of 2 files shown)


jitendravjh:
Closing, already handled by #3741.

PR #3232, opened 2026-07-06
fix(channels): rename routing.yaml to routing.json with legacy migration

### LINKED ISSUE #3076: fix(channels): rename routing.yaml to routing.json and handle legacy format
## Problem

The routing file is named `routing.yaml` but contains JSON-formatted content. The code uses `JSON.stringify` on write and `JSON.parse` on read — no YAML parser is ever involved. This is a naming mismatch from the original channels MVP.

## Location

- File: `~/.letta/channels/<channel>/routing.yaml`
- Code: `src/channels/routing.ts` (load/save functions)
- Path helper: `src/channels/config.ts` → `getChannelRoutingPath()`

## Solution

1. Rename the file to `routing.json`
2. Update `getChannelRoutingPath()` to return `.json` extension
3. On load, check for legacy `.yaml` path first; if found, migrate (rename to `.json`) and delete the old file
4. Ensure atomic migration to avoid data loss

## Acceptance criteria

- New installs use `routing.json`
- Existing installs with `routing.yaml` are migrated automatically on first load
- No user-visible behavior change beyond the filename
- Test coverage for migration path

nolanchic:
Fixes #3076.

## Problem

Channel routes are stored as JSON, but the file was named `routing.yaml`. Renaming it must preserve existing routes for every reader, including the teleport guard, even when the filesystem does not permit migration.

## Changes

- New writes use `routing.json`.
- Shared disk persistence handles legacy files for both direct reads and cache loads. The teleport guard therefore sees existing channel bindings before any channel startup or cache load.
- Migration validates the JSON routing envelope, then publishes the existing file under the new name using a hard link. This is atomic and cannot overwrite a destination created concurrently. The legacy name is removed only after publication succeeds.
- When migration is unavailable, valid legacy routes remain readable. A concurrent migration or save is detected by rereading the current file.
- When both files exist, `routing.json` is authoritative, including an empty route table. A corrupt or unreadable current file does not fall back to potentially stale legacy routes.
- Normal saves write and flush a unique temporary file, then rename it over the current snapshot. Write, flush, or rename failures leave the previous snapshot intact.
- CLI help and channel documentation use the new filename.

## Validation

- 73 focused tests passed across routing, routing persistence, route thread keys, the teleport subcommand, disposable-home isolation, and listener teleport handling. The original migration cases and the new persistence regressions are consolidated into one 23-case disk suite covering direct reads, migration failures, concurrent publication, interrupted cleanup, malformed content, and partial/failed writes.
- CI fixture follow-up: updated the current-filename expectation and the simulated gateway writer, retaining coverage of both current and legacy files. The two reported failures were reproduced locally before the correction. The affected unit shard passed locally, including 1,086 tests in its shared batch plus the isolated suites.
- `bun run check`: all 12 checks passed, including TypeScript, formatting, and architectural checks.
- Real filesystem smoke checks passed under Bun and a Node-targeted bundle on macOS: direct reads and teleport protection before cache load, readable legacy routes in a non-writable directory, successful migration retry, and save/reload.
- Windows was not exercised locally. Crash scenarios are covered by deterministic failure injection and interrupted-operation states; these checks do not simulate power loss.

## Contributions

The original filename correction and migration tests were contributed by @nolanchic. A follow-up commit adds shared persistence, migration and save safety, and regression coverage while preserving the original commits.

## AI disclosure

Authored with AI assistance. The original contribution used Claude Code; the follow-up rework used Codex.

### DIFF
--- a/src/channels/config.ts
+++ b/src/channels/config.ts
@@ -46,6 +46,15 @@ export function getChannelAccountsPath(channelId: string): string {
 }
 
 export function getChannelRoutingPath(channelId: string): string {
+  return join(getChannelDir(channelId), "routing.json");
+}
+
+/**
+ * Legacy routing path from the original channels MVP, when the file was named
+ * `routing.yaml` despite holding JSON content. Used only for one-time migration
+ * to the correctly-named `routing.json`. See issue #3076.
+ */
+export function getLegacyChannelRoutingPath(channelId: string): string {
   return join(getChannelDir(channelId), "routing.yaml");
 }
 
--- a/src/channels/routing.test.ts
+++ b/src/channels/routing.test.ts
@@ -1,5 +1,15 @@
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
 import {
+  existsSync,
+  mkdirSync,
+  mkdtempSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { __testOverrideChannelsRoot } from "@/channels/config";
+import {
   __testOverrideLoadRoutes,
   __testOverrideSaveRoutes,
   addRoute,
@@ -7,8 +17,10 @@ import {
   getAllRoutes,
   getRoute,
   getRoutesForChannel,
+  loadRoutes,
   removeRoute,
   removeRoutesForScope,
+  saveRoutes,
 } from "@/channels/routing";
 
 describe("routing", () => {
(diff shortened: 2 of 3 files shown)


nolanchic:
Heads up on the Windows CI failure — it's a flaky test unrelated to this change.

The only failing check on Windows is:

```
src/tools/bash.test.ts > Bash tool > executes simple command
  ^ this test timed out after 15000ms.
  killed 1 dangling process
```

A few reasons this looks environmental rather than a real regression:

- **The test is unchanged by this PR.** `git diff origin/main -- src/tools/bash.test.ts` is empty — this PR only touches `src/channels/routing.ts` and `src/channels/config.ts`. There's no path between the routing migration and the Bash tool.
- **A near-identical test passes in the same run.** `src/tools/run-shell-command.test.ts > executes simple command` (also runs `echo` via a shell) passed on the same Windows runner in 391ms.
- **Windows is green on `main`.** The latest `main` push run's `Windows x64 (windows-latest)` job concluded `success`.
- The "killed 1 dangling process" + 15015ms timeout signature matches a shell-process startup hang on the runner, not an assertion failure.

The four routing-migration tests (`routing disk format migration (#3076)`) all passed on Windows too.

Would you mind re-running the Windows job? I don't have permission to rerun upstream CI as an external contributor. Happy to investigate further if it fails again on retry.

just-cameron (approved):
Reviewed the full diff against merge-base 85cb7ed5.

**Approve.** The migration design is correct and well-engineered:

- Hard-link publication () makes migration atomic and unable to clobber a concurrently created `routing.json`; on link failure it re-reads the current file to pick up the concurrent writer's result.
- Migration is never a prerequisite for reading — on EACCES/EROFS/EPERM/ENOTSUP (all tested), legacy routes remain readable and migration can be retried.
- Both-files-exist semantics are right: `routing.json` is authoritative including an empty route table, and a corrupt/unreadable current file does not fall back to stale legacy routes (only ENOENT permits fallback).
- The write path is now atomic (unique temp file, fsync, rename) where the old code did a bare `writeFileSync` that a crash could destroy.
- Teleport guard covered: `getLocalChannelTeleportError` sees legacy bindings before any cache load, with a dedicated test.
- `loadRoutes` cache behavior is unchanged from the old implementation (verified old vs new side by side).
- No remaining `routing.yaml` references outside the legacy-path helper, docs describing migration, and tests exercising both filenames.

All 38 CI checks green including Windows unit shards and packaging.

just-cameron (approved):
Reviewed the full diff against merge-base 85cb7ed5.

**Approve.** The migration design is correct and well-engineered:

- Hard-link publication (fs.linkSync) makes migration atomic and unable to clobber a concurrently created routing.json; on link failure it re-reads the current file to pick up the concurrent writer's result.
- Migration is never a prerequisite for reading — on EACCES/EROFS/EPERM/ENOTSUP (all tested), legacy routes remain readable and migration can be retried.
- Both-files-exist semantics are right: routing.json is authoritative including an empty route table, and a corrupt/unreadable current file does not fall back to stale legacy routes (only ENOENT permits fallback).
- The write path is now atomic (unique temp file, fsync, rename) where the old code did a bare writeFileSync that a crash could destroy.
- Teleport guard covered: getLocalChannelTeleportError sees legacy bindings before any cache load, with a dedicated test.
- loadRoutes cache behavior is unchanged from the old implementation (verified old vs new side by side).
- No remaining routing.yaml references outside the legacy-path helper, docs describing migration, and tests exercising both filenames.

All 38 CI checks green including Windows unit shards and packaging.

cpacker (approved):

just-cameron:
Thanks @nolanchic!

PR #3233, opened 2026-07-06
fix(cron): validate field value ranges in isValidCron

nolanchic:
## Problem

Cron validation only checked token shape. Out-of-range expressions could be stored, while validation and execution used different semantics. Moving execution to `cron-parser` also changed the phase of existing day/month wildcard-step schedules.

## Fix

Use `cron-parser` behind both validation and matching while preserving Letta Code's existing five-field numeric cron dialect.

- Reject out-of-range values, reversed ranges, zero steps, unsupported syntax, and bare value steps such as `5/10`.
- Preserve the legacy modulo phase for day-of-month and month wildcard steps, including comma/range unions. Standalone steps with no legal value (`*/32` day-of-month and `*/13` month) remain invalid; unions still work when another arm can match.
- Validate schedules at the shared add/cron-changing-update persistence boundary so every caller gets the same guarantee.
- Keep previously persisted invalid schedules active and inspectable, record one durable `invalid_cron` failure, and apply the same preflight in the WebSocket and TUI schedulers.
- Clear stale invalid-cron failure metadata when a schedule is corrected.
- Preserve Sunday `7`, day-of-month/day-of-week OR behavior, timezone handling, and wall-clock semantics across spring-forward gaps and fall-back folds.

## Validation

- Focused parser/storage/scheduler tests: 135 passed
- `bun run check`: 12 checks passed
- `bun run build`: passed
- Differential compatibility probe: 113,880 legacy schedule evaluations matched across UTC, America/Los_Angeles, and Asia/Tokyo
- Complete unit runner before the final DST-only correction: 4,978 passed, 25 skipped; six unrelated recurring failures plus one worktree-cwd artifact that passed 23/23 on immediate rerun
- Independent adversarial review found the DST boundary issue; follow-up review verified the correction
- Review surface reduced from 14 files / 794 additions to 12 files / 486 additions

## AI disclosure

Authored with AI assistance (Claude Code and Letta Code) and reviewed before submission.


just-cameron:
﻿Nice catch. I think this should probably use an existing cron parser instead of extending the custom validator one case at a time.

The current patch still leaves validation and matching with different semantics. Example: this PR accepts day-of-week `7` as Sunday, but `cronMatchesTime` compares against `Date.getDay()` values `0-6`, so `0 0 * * 7` becomes valid but still never fires. Same footgun, just one layer deeper. Ranges containing `7` have the same problem, and reversed ranges like `59-0 * * * *` still validate even though the matcher can never satisfy them.

I would rather make a library the source of truth for cron semantics, probably `cron-parser`:

- keep our product dialect gate before parsing: exactly 5 fields, and numeric syntax only if we do not want to expand support to `MON`, `JAN`, `?`, `L`, etc.
- implement `isValidCron` by calling the parser after that dialect gate.
- ideally implement `cronMatchesTime` via the parser's next occurrence calculation, so validation and execution share the same semantics.

Shape would be roughly:

```ts
import { CronExpressionParser } from "cron-parser";

const SUPPORTED_CRON_RE = /^[\d\s*,*/-]+$/;

export function isValidCron(expr: string): boolean {
  const trimmed = expr.trim();
  const fields = trimmed.split(/\s+/);

  if (fields.length !== 5) return false;
  if (!SUPPORTED_CRON_RE.test(trimmed)) return false;

  try {
    CronExpressionParser.parse(trimmed, { strict: false });
    return true;
  } catch {
    return false;
  }
}
```

Then `cronMatchesTime` can ask for the next occurrence after the previous millisecond and compare it to the current minute. That also gives us timezone and DST semantics from a maintained library rather than our own partial implementation.

If we want to keep this PR smaller, I would at least fix DOW `7` matching and reversed ranges here, and change the PR description to say this is field-range validation only, not robust cron validation.

nolanchic:
Thanks for the review @just-cameron — agreed on all counts. The DOW `7` / reversed-range gaps are exactly the "validation and matching with different semantics" problem, and patching them one case at a time is the wrong direction.

I'll rework this to put `cron-parser` behind both `isValidCron` and `cronMatchesTime` so they share one source of truth, keeping the product dialect gate (5 fields, numeric syntax only) in front of the parser. I'll also drop the hand-rolled `fieldMatches` step/range logic in favor of the parser's next-occurrence calculation for matching, which gives us timezone/DST for free.

Will push the rework to this PR.

nolanchic:
Reworked per your feedback — `cron-parser` is now the source of truth for both validation and matching (commit f8f3d82).

**`isValidCron`**: keeps the product dialect gate (exactly 5 fields, numeric syntax only via `/^[\d\s*,*/-]+$/`) then delegates to `CronExpressionParser.parse(expr, { strict: false })`. So out-of-range values, reversed ranges (`59-0`), and zero steps (`0-59/0`) are all rejected by the library instead of by hand-rolled per-field checks.

**`cronMatchesTime`**: now asks cron-parser for the next occurrence after the previous minute and checks it lands in the target minute. This closes the DOW-`7` gap you called out — `0 0 * * 7` now both validates **and** matches Sunday (previously valid-but-never-fired). Reversed ranges can't sneak through either, since the same parser is used for both. Timezone/DST now come from the library; an invalid timezone falls back to local (validated via `Intl.DateTimeFormat` before handing to cron-parser) to preserve the previous behavior.

Drops ~170 lines of hand-rolled `fieldMatches` / `getTimeComponents` / range math.

**Tests** (`src/cron/parse-interval.test.ts`, 96 pass across `src/cron/`):
- new: DOW `7` matches Sunday just like `0` (the exact gap you flagged); reversed ranges (`59-0`, `31-1`) rejected at validation
- existing: all prior valid/invalid `isValidCron` and `cronMatchesTime` cases still pass unchanged, including the invalid-timezone-falls-back-to-local case

`bun run typecheck`, biome, layer/cycles/exported-functions checks all clean. New dep: `cron-parser ^5.6.1`.

I did **not** migrate `cronMatchesTime`'s callers to use the parser's iterator directly (e.g. scheduler.ts still calls `cronMatchesTime(now)`), to keep this PR scoped to "fix validation/matching semantics". Happy to follow up on that if you'd rather have the scheduler drive off `next()` directly.

just-cameron (approved):

sarahwooders (approved):

just-cameron:
Thank you!

PR #3234, opened 2026-07-06
fix(lsp): reject pending requests when the server process exits or stdout errors

nolanchic:
## Problem

`LSPClient` only settled pending requests when a JSON-RPC **response** arrived (`handleMessage`). The `'exit'` handler on the server process just emitted an event and did nothing with `pendingRequests`. So if the server process exited while requests were outstanding — a crash, or a normal `shutdown()` that races with another in-flight request — those promises were **left pending forever**:

- callers awaiting them (including `shutdown()` itself, which `await`s `sendRequest("shutdown")`) would hang;
- the `pendingRequests` map leaked.

There was also no `'error'` handling on stdout: a broken stream likewise meant no response would ever arrive, with the same hang.

## Fix

- Add a `rejectAllPending(error)` helper that rejects every outstanding request and clears the map (no-op once empty, so safe to call from multiple teardown paths).
- On server `'exit'`: mark the client `disposed`, then `rejectAllPending` with a descriptive error, then emit `'exit'`.
- On stdout `'error'`: `rejectAllPending` (no response can arrive via a broken stream) in addition to emitting `'error'`.
- `sendRequest` fails fast when already `disposed`, so new requests against a dead server reject immediately instead of enqueueing an unanswerable promise.
- `sendMessage` no-ops when `disposed`, to avoid writing to a dead stdin (which would throw `ERR_STREAM_WRITE_AFTER_END` / `EPIPE`).

Normal request/response flow is unchanged: a request that gets a response still resolves; only the "no response will ever come" paths now reject.

## Tests

New `src/lsp/client.test.ts` (no external language server required — spawns a tiny `bun -e` script that controls stdin/stdout and exit behavior):

- **pending request is rejected when the server exits without responding**: a request in-flight when the process dies rejects with `LSP server exited` (before the fix this hung).
- **request sent after the server exited rejects immediately**: subsequent `sendRequest` against a disposed client fails fast with `LSP server has exited`.
- **responded request still resolves; only a later unrelated exit rejects others**: a normal response resolves normally, then a second request pending at exit is rejected — proving the happy path is intact.

## Verification

- `bun test src/lsp/client.test.ts` — 3 pass (run with a timeout guard)
- `bun run typecheck` — clean
- `bun run check:boundaries`, `check:cycles`, `check:exported-functions`, `check:filename-casing` — clean
- biome — clean on changed files

## AI disclosure

Authored with AI assistance (Claude Code) and reviewed by a human before submission.

### DIFF
--- a/src/lsp/client.test.ts
+++ b/src/lsp/client.test.ts
@@ -1,0 +1,119 @@
+import { expect, test } from "bun:test";
+import { spawn } from "node:child_process";
+import { LSPClient } from "@/lsp/client";
+
+/**
+ * Spawn a minimal "LSP server" process: it reads JSON-RPC messages from stdin
+ * (so the client's Content-Length framing works) but its behavior is controlled
+ * by the script. We use this instead of a mock so the real ChildProcess wiring
+ * and exit handling in LSPClient are exercised.
+ */
+function spawnScript(script: string) {
+  return spawn("bun", ["-e", script], {
+    stdio: ["pipe", "pipe", "pipe"],
+  });
+}
+
+test("pending request is rejected when the server process exits without responding", async () => {
+  // Server: read and discard stdin, then exit immediately on the first message.
+  const server = spawnScript(`
+    let buf = "";
+    process.stdin.on("data", (chunk) => {
+      buf += chunk.toString();
+      // Once we've received anything, exit without responding.
+      if (buf.includes("Content-Length:")) {
+        // Exit on next tick so the client's write flushes.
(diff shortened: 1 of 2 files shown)


nolanchic:
Gentle nudge on this one — it's been sitting since July while the area it touches (LSP request lifecycle) keeps coming up in newer reports. I rebased it onto current main recently; happy to split or reshape it if the current form is the blocker. A quick look whenever the queue allows would be welcome.

PR #3235, opened 2026-07-06
fix(local): inherit conversation context_window_limit from model_settings

### LINKED ISSUE #3132: Local backend: new conversations don't inherit agent's context_window_limit — always fall back to hardcoded 128000
## Summary

When the Letta Code desktop app creates a new conversation for a local-backend agent, the conversation's `context_window_limit` is never populated from the agent's `model_settings`. The field is left undefined on the conversation record, causing the effective context window to fall back to the hardcoded `128000` default — even when the agent explicitly has `context_window_limit: 1000000` in its `model_settings`.

**Related**: #2897 covers the agent-level variant (mod-registered providers storing legacy 128000). This issue covers the **conversation creation path** — a distinct code path where even correctly-configured agents lose their context window on new conversations.

## Version

- Letta Code v0.27.9 (desktop app, `--backend local`)
- macOS, local backend at `~/.letta/lc-local-backend`

## Repro

1. Create a local agent with explicit context window:
   ```
   model_settings: {
     "context_window_limit": 1000000,
     "max_tokens": 131072
   }
   ```
2. Open the desktop app, select the agent, start a **new** conversation.
3. Check `/context` or the context window indicator — it shows 128000 instead of 1000000.
4. Run `/context-limit 1000000` in the conversation — it fixes THAT conversation only.
5. Start another new conversation — back to 128000.

### 1. Frontend: `DraftChatViewInner` doesn't send `context_window_limit` as a top-level field

When creating a conversation, the desktop frontend sends:
```javascript
requestBody: {
    summary: summary,
    model: draftModel ?? agent.model,
    model_settings: normalizeDraftModelSettings(draftModel ?? agent.model, draftModelSettings ?? agent.model_settings)
    // ← context_window_limit is inside model_settings, NOT sent as top-level field
}
```

The `model_settings` object DOES contain `context_window_limit`, but it's nested — not sent as the top-level `context_window_limit` that the backend expects.

### 2. Backend: `createConversationRecord()` only checks top-level `body.context_window_limit`

```typescript
function createConversationRecord(agentId, body) {
    return {
        // ...
        ...modelFieldFromBody(body),
        ...(body.model_settings === null || isRecord(body.model_settings)
            ? { model_settings: body.model_settings }
            : {}),
        ...(typeof body.context_window_limit === 'number'    // ← never sent by frontend
            ? { context_window_limit: body.context_window_limit }
            : {}),
    };
}
```

Since `body.context_window_limit` is never set by the frontend, the conversation record is created without a top-level `context_window_limit` field.

### 3. The hardcoded 128000 fallback

```typescript
// defaultModelSettings()
function defaultModelSettings() {
    return { context_window_limit: 128000 };
}

// projectAgent()
const contextWindow = typeof record.model_settings.context_window_limit === 'number'
    ? record.model_settings.context_window_limit
    : 128000;  // ← fallback when undefined
```

### 4. The `/models` endpoint returns hardcoded 128000 for ALL models

```typescript
router.get('/models', (_req, res) => {
    // ...
    res.json([...handles].map((handle) => ({
        // ...
        context_window: 128000,  // ← hardcoded for every model
    })));
});
```

## Impact

- Users must manually run `/context-limit` on every new conversation
- The context window display is wrong, causing premature compaction
- Users may not realize they're operating at 1/8th of their intended context

## Scale (real-world)

On my instance: 828 conversations total, 820 were missing `context_window_limit` (99%). Only the 8 where I manually ran `/context-limit` had the correct value.

## Proposed Fix

**Option A — Frontend fix** (preferred): Send `context_window_limit` as a top-level field when creating a conversation:

In `DraftChatViewInner`, extract it from model_settings:
```javascript
const fu = useCallback((ju, Mu = null) => {
    const modelSettings = Wl ?? wa.model_settings;
    const contextWindowLimit = typeof modelSettings?.context_window_limit === 'number'
        ? modelSettings.context_window_limit
        : undefined;
    
    ru || !wa || tu({
        agentId: ya,
        requestBody: {
            summary: Mu,
            model: Hl ?? wa.model,
            model_settings: normalizeDraftModelSettings(Hl ?? wa.model, modelSettings),
            ...(contextWindowLimit !== undefined
                ? { context_window_limit: contextWindowLimit }
                : {}),
        }
    }, { /* ... */ });
}, [/* deps */]);
```

**Option B — Backend fix**: In `createConversationRecord()`, extract `context_window_limit` from `model_settings` if not provided at top level:

```typescript
function createConversationRecord(agentId, body) {
    const modelSettings = isRecord(body.model_settings) ? body.model_settings : {};
    return {
        // ...existing fields...
        ...(typeof body.context_window_limit === 'number'
            ? { context_window_limit: body.context_window_limit }
            : typeof modelSettings.context_window_limit === 'number'
            ? { context_window_limit: modelSettings.context_window_limit }
            : {}),
    };
}
```

Either fix is minimal and surgical. Option B is more defensive and covers all entry points.

## Workaround

Patch script that reads agent `model_settings.context_window_limit` and writes it to conversation records:
```bash
python3 patch-context-limit.py --apply
```

Happy to submit a PR if helpful.

abhay-codes07:
## Summary
- New local-backend conversations dropped the agent's configured `context_window_limit`: the desktop draft flow (and other callers) send the limit nested inside `model_settings`, but `createLocalConversationRecord` only read the top-level `context_window_limit` field. The stored record ended up without a limit, so every new conversation fell back to the hardcoded 128000 default even when the agent was configured with a much larger window.
- Lift `model_settings.context_window_limit` to the conversation record's top-level field at creation when no explicit top-level value is provided. An explicit top-level value still wins, and conversations with no configured limit keep the field unset.

Fixes #3132.

## Test plan
- [x] `bun test src/backend/local-backend.test.ts` — adds a repro test (fails without the fix) plus guards for top-level precedence and the no-limit case
- [x] `bun run check`
- [x] Full unit suite — no new failures vs a clean `main` baseline on the same machine

## AI disclosure
Following the policy proposed in #3139: this PR was developed with AI assistance (Claude Code) — the investigation, patch, and tests were AI-assisted. I directed the work, reviewed the change, and ran the validation above locally; the repro test was mutation-checked (it fails with the fix reverted).

cc @cpacker — this sits right next to the record-normalization work in #3136, happy to adjust if you'd rather fold it in there.

### DIFF
--- a/scripts/source-file-size-baseline.json
+++ b/scripts/source-file-size-baseline.json
@@ -1,9 +1,9 @@
 {
   "src/agent/client-skills.test.ts": 1196,
   "src/agent/memory-git.ts": 2129,
-  "src/backend/local-backend.test.ts": 2534,
+  "src/backend/local-backend.test.ts": 2618,
   "src/backend/local/local-backend.ts": 1058,
-  "src/backend/local/local-store.ts": 3613,
+  "src/backend/local/local-store.ts": 3615,
   "src/backend/pi-stream-adapter.test.ts": 1442,
   "src/cli/app/AppCoordinator.tsx": 5198,
   "src/cli/app/AppView.tsx": 1750,
--- a/src/backend/local-backend.test.ts
+++ b/src/backend/local-backend.test.ts
@@ -252,6 +252,90 @@ describe("local backend pi transcript", () => {
     }
   });
 
+  test("new conversations inherit context_window_limit nested in model_settings", async () => {
+    const storageDir = await mkdtemp(
+      join(tmpdir(), "local-backend-conversation-context-window-"),
+    );
+    try {
+      const backend = new LocalBackend({ storageDir, memfsEnabled: false });
+      const agent = await backend.createAgent({
+        name: "Local",
+        model: "anthropic/claude-sonnet-4-6",
+        model_settings: {
+          provider_type: "anthropic",
+          context_window_limit: 1000000,
+        },
+      } as never);
+
+      const conversation = await backend.createConversation({
+        agent_id: agent.id,
+        model: "anthropic/claude-sonnet-4-6",
+        model_settings: {
+          provider_type: "anthropic",
+          context_window_limit: 1000000,
+          max_tokens: 131072,
+        },
+      } as never);
+
+      expect(
+        (conversation as unknown as { context_window_limit?: number })
+          .context_window_limit,
+      ).toBe(1000000);
+
+      const retrieved = (await backend.retrieveConversation(
+        conversation.id,
+      )) as unknown as { context_window_limit?: number };
+      expect(retrieved.context_window_limit).toBe(1000000);
+    } finally {
+      await rm(storageDir, { recursive: true, force: true });
+    }
+  });
+
+  test("top-level context_window_limit wins over the model_settings value", async () => {
+    const storageDir = await mkdtemp(
+      join(tmpdir(), "local-backend-conversation-context-window-top-"),
+    );
+    try {
+      const backend = new LocalBackend({ storageDir, memfsEnabled: false });
+      const agent = await backend.createAgent({ name: "Local" } as never);
+
+      const conversation = await backend.createConversation({
+        agent_id: agent.id,
+        context_window_limit: 200000,
+        model_settings: { context_window_limit: 1000000 },
+      } as never);
+
+      expect(
+        (conversation as unknown as { context_window_limit?: number })
+          .context_window_limit,
+      ).toBe(200000);
+    } finally {
+      await rm(storageDir, { recursive: true, force: true });
+    }
+  });
+
+  test("conversations without any context window limit leave the field unset", async () => {
+    const storageDir = await mkdtemp(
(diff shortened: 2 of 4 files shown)


jibramnobhill:
Closing as superseded by current main architecture.

## Why this PR is no longer needed

PR #3136 (by Charles Packer, merged July 10) introduced src/backend/local/local-model-normalization.ts, which extracts context_window_limit from nested model_settings in the conversation body. The current main code path (createLocalConversationRecord → supportedConversationModelSettingsFromBody) already reads the nested value correctly.

This PR's approach — denormalizing the nested value into a duplicate top-level field — would actually introduce a stale-copy risk: a later nested-only update could leave the top-level copy outdated, and top-level precedence would mask the newer setting.

## Prevention test retained

The behavioral regression test from this PR's investigation has been published as draft PR #4005, which verifies the nested context-window limit survives reload, triggers preflight compaction, and rejects mutation.

Thanks to @MartinMayday for the original report (#3132) and to the original PR author for the investigation.

PR #3238, opened 2026-07-06
fix(bash): recover from deleted working directories

overlord-letta:
## Summary

- LET-9464: recover tool execution when the runtime-scoped working directory has been deleted mid-turn.
- Distinguish a missing working directory from a missing shell executable so deleted worktrees no longer surface as `Executable not found: /bin/zsh`.
- Surface a one-time recovery note to the model when Letta Code has to run from a fallback cwd, so relative-path commands are not silently redirected.
- Add regression coverage for Bash, shell_command, shell env, missing-cwd error handling, and cross-platform cwd path normalization.

## Anonymized trace sequence

This PR is grounded in a real internal listener trace, but the public description intentionally removes run IDs, agent IDs, organization IDs, user names, repository names, branch names, and absolute user paths.

The trace showed this sequence:

1. A macOS listener conversation had its runtime working directory set to a task-specific git worktree.
2. The agent verified merge state from the main checkout by putting `cd <main checkout> && ...` inside a Bash command. That command succeeded, but it did not change Letta Code's stored runtime working directory. The conversation still pointed at the task worktree.
3. The agent then ran its cleanup command. Because the runtime cwd was still the task worktree, the Bash tool spawned from that worktree and removed that same worktree during cleanup.
4. The cleanup command's own output then showed git failures equivalent to `fatal: Unable to read current working directory`, because the process cwd had disappeared while the shell was still running.
5. The agent tried to recover with another Bash command that again contained `cd <main checkout> && ...`, but `child_process.spawn()` has to enter the configured cwd before the command string runs. Since the configured cwd no longer existed, the command body never executed.
6. Node surfaced that spawn failure as `ENOENT`.
7. Our shell runner treated every spawn `ENOENT` as an executable lookup failure and returned `Executable not found: /bin/zsh`.
8. Several later approval continuations attempted simple shell probes such as `echo ...`. They all failed before the command body ran, with null stdout/stderr and the same misleading shell-not-found tool response.
9. The final run in that sequence was cancelled after the repeated tool failures. The model/provider steps were not the root cause; the failure was local tool execution against a deleted cwd.

## Root cause

There were two separate gaps.

First, `getConversationWorkingDirectory()` already prunes stale persisted cwd entries at listener turn start, but this incident happened inside a single active turn. `handleIncomingMessage()` captured `turnWorkingDirectory` once. Later, during the approval loop, a Bash command deleted that captured directory. The existing stale-cwd guard never got another chance to run before the next tool execution.

Second, `spawnWithLauncher()` reported all spawn-time `ENOENT` failures as missing executables. That is correct when the cwd exists and the executable path is unavailable, but it is wrong when the executable exists and the configured cwd is gone. In this case `/bin/zsh` was present; the cwd was not.

A review pass found two more parity issues after the first fix:

- Silent recovery was dangerous: a command such as `rm -rf ./build` could run from the fallback directory without the model knowing the original cwd had disappeared.
- The Codex-style `shell` / `shell_command` path needed the same cwd-missing fast-fail and recovery-note behavior as Bash, otherwise it could retry fallback launchers for a missing cwd and leave the recovery note to be consumed by a later tool call.

## Fix

This change adds these protections:

1. `getCurrentWorkingDirectory()` validates the runtime-context cwd before returning it. If the runtime cwd has disappeared, it falls back to the first usable directory from:
   - `USER_CWD`
   - `process.cwd()`
   - the user's home directory
   - `/` on Unix as a final safety fallback

   When it repairs a runtime-context cwd, it also updates the current async runtime context so subsequent tools in the same turn reuse the recovered directory instead of repeatedly trying the deleted path.

2. Shell tools consume that repair state and prepend a one-time note:

   `Note: working directory <old> no longer exists; running in <new> instead.`

   This keeps the model aware that relative paths now resolve somewhere else. Interrupt output remains byte-exact so downstream interrupt handling still works.

3. `spawnWithLauncher()` now preflights the cwd and also re-checks cwd state when spawn emits `ENOENT`. If the cwd is missing, the error becomes `Working directory not found: <path>` with reason `cwd_missing`. If the cwd is valid and the executable is missing, it still reports `Executable not found: <executable>` with reason `executable_missing`.

4. Bash and shell fallback loops only retry alternate shell launchers for executable lookup failures. A missing cwd now fails fast instead of retrying every fallback shell and potentially evicting a valid cached launcher.

5. Directory usability checks are centralized in `src/helpers/usable-directory.ts` so runtime context, shell spawning, and listener cwd cleanup use the same implementation.

The intended behavior after this patch is:

- Runtime-scoped deleted cwd: recover to a valid fallback cwd, tell the model once, and continue running tools.
- Explicit deleted cwd at spawn time: fail clearly as a missing working directory.
- Actual missing shell executable: continue reporting an executable-not-found error and use existing fallback launcher behavior where appropriate.

## Test plan

- `bun test src/tools/bash.test.ts src/tools/shell-env.test.ts src/tools/shell-command.test.ts`
- `bun run typecheck`
- `bun run check`

CI was green at head `e17d534a`. The latest head `85ad07fe` adds the shell_command parity fix and is rerunning CI.

### DIFF
--- a/src/helpers/usable-directory.ts
+++ b/src/helpers/usable-directory.ts
@@ -1,0 +1,14 @@
+import { statSync } from "node:fs";
+
+/** True when the path exists and is a directory. */
+export function isUsableDirectory(dirPath: string | null | undefined): boolean {
+  if (typeof dirPath !== "string" || dirPath.length === 0) {
+    return false;
+  }
+
+  try {
+    return statSync(dirPath, { throwIfNoEntry: false })?.isDirectory() ?? false;
+  } catch {
+    return false;
+  }
+}
--- a/src/runtime-context.ts
+++ b/src/runtime-context.ts
@@ -1,4 +1,6 @@
 import { AsyncLocalStorage } from "node:async_hooks";
+import { existsSync, statSync } from "node:fs";
+import { homedir } from "node:os";
 import type { SkillSource } from "./agent/skills";
 import type { MessageChannelToolDiscoveryScope } from "./channels/message-tool";
 import type { ChannelTurnSource } from "./channels/types";
@@ -76,10 +78,62 @@ export function updateRuntimeContext(
   );
 }
 
+function isUsableDirectory(dirPath: string | null | undefined): boolean {
+  if (typeof dirPath !== "string" || dirPath.length === 0) {
+    return false;
+  }
+
+  try {
+    return existsSync(dirPath) && statSync(dirPath).isDirectory();
+  } catch {
+    return false;
+  }
+}
+
+function getProcessWorkingDirectory(): string | null {
+  try {
+    return process.cwd();
+  } catch {
+    return null;
+  }
+}
+
+function getFallbackWorkingDirectory(): string {
+  const fallback = [
+    process.env.USER_CWD,
+    getProcessWorkingDirectory(),
+    homedir(),
+    process.platform === "win32" ? undefined : "/",
+  ].find(isUsableDirectory);
+
+  if (fallback) {
+    return fallback;
+  }
+
+  // Extremely defensive fallback for pathological environments where every
+  // candidate disappeared. The caller will still surface a clear cwd error.
+  return process.platform === "win32" ? "C:\\" : "/";
+}
+
 export function getCurrentWorkingDirectory(): string {
-  const workingDirectory = runtimeContextStorage.getStore()?.workingDirectory;
-  if (typeof workingDirectory === "string" && workingDirectory.length > 0) {
+  const runtimeContext = runtimeContextStorage.getStore();
+  const workingDirectory = runtimeContext?.workingDirectory;
+  if (
+    typeof workingDirectory === "string" &&
+    isUsableDirectory(workingDirectory)
+  ) {
     return workingDirectory;
   }
-  return process.env.USER_CWD || process.cwd();
+
+  const fallback = getFallbackWorkingDirectory();
+  if (
+    runtimeContext &&
+    typeof workingDirectory === "string" &&
+    workingDirectory.length > 0 &&
+    workingDirectory !== fallback
+  ) {
+    runtimeContext.workingDirectory = fallback;
+  }
+
+  return fallback;
 }
--- a/src/runtime-context.ts
+++ b/src/runtime-context.ts
@@ -1,9 +1,9 @@
 import { AsyncLocalStorage } from "node:async_hooks";
-import { existsSync, statSync } from "node:fs";
 import { homedir } from "node:os";
 import type { SkillSource } from "./agent/skills";
 import type { MessageChannelToolDiscoveryScope } from "./channels/message-tool";
 import type { ChannelTurnSource } from "./channels/types";
+import { isUsableDirectory } from "./helpers/usable-directory";
 
 export type RuntimePermissionMode = "standard" | "acceptEdits" | "unrestricted";
 
@@ -14,6 +14,12 @@ export interface RuntimeContextSnapshot {
   skillsDirectory?: string | null;
   skillSources?: SkillSource[];
   workingDirectory?: string | null;
+  /**
+   * Set when the runtime-scoped working directory was found deleted and
+   * repaired to a fallback mid-turn. Holds the original (now missing) path
+   * until a shell tool consumes it to surface a note to the model.
+   */
+  workingDirectoryRecoveredFrom?: string | null;
   toolContextId?: string | null;
   permissionMode?: RuntimePermissionMode;
   channelToolScope?: MessageChannelToolDiscoveryScope | null;
@@ -78,18 +84,6 @@ export function updateRuntimeContext(
   );
 }
 
-function isUsableDirectory(dirPath: string | null | undefined): boolean {
-  if (typeof dirPath !== "string" || dirPath.length === 0) {
-    return false;
-  }
-
-  try {
-    return existsSync(dirPath) && statSync(dirPath).isDirectory();
-  } catch {
-    return false;
-  }
-}
-
 function getProcessWorkingDirectory(): string | null {
   try {
     return process.cwd();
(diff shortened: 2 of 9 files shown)


cpacker:
Reviewed this and pushed review fixes directly in e17d534a. The core diagnosis and seam choice are correct — `getCurrentWorkingDirectory()` covers Bash, exec_command, apply_patch, and file tools in one place, and the spawn-side ENOENT disambiguation matches the incident (`/bin/zsh` existed, the cwd did not). Fixes in the follow-up commit:

1. **Windows CI failure (blocking):** the recovery test compared the child's long-form cwd (`runneradmin`) against `mkdtemp`'s 8.3 short path (`RUNNER~1`). Now `realpath`s the fixture, which also guards the macOS `/var → /private/var` symlink.
2. **Silent recovery:** the repair path previously ran commands from the fallback dir with zero indication to the model — a relative-path command (`rm -rf ./build`) after silent recovery would execute against the wrong tree. Bash results now prepend `Note: working directory <old> no longer exists; running in <new> instead.` (skipped on abort results, which are compared byte-exact against `INTERRUPTED_BY_USER` downstream). exec_command does not surface the note yet; the repair itself still applies there.
3. **Windows launcher loop:** a missing cwd produced `code: "ENOENT"`, so `spawnCommand` retried every launcher (appending a misleading `(tried: powershell.exe, ...)` suffix) and spuriously evicted the cached working launcher. `ShellExecutionError` now carries `reason: "cwd_missing" | "executable_missing"` and cwd-missing errors fail fast.
4. **Dedup:** this PR added the 2nd and 3rd copies of `isUsableDirectory` (plus the pre-existing one in `listener/cwd.ts`). Consolidated into `src/helpers/usable-directory.ts` with a single `statSync(..., { throwIfNoEntry: false })` instead of the `existsSync` + `statSync` double stat.
5. Routed the repair through `updateRuntimeContext()` instead of mutating the store directly.

Validation: `bun test src/tools/bash.test.ts src/tools/shell-env.test.ts src/tools/shell-command.test.ts src/tools/enter-worktree.test.ts` (remaining failures reproduce identically on the base commit — environment-local FS sandbox issues, green in CI) and `bun run check` (10/10).

amelia-letta:
## Shell tool parity gap

`shell.ts` has two gaps vs the `bash.ts` changes in this PR:

1. **Missing `cwd_missing` guard in ENOENT retry** (line 133): `shell.ts` retries all fallback launchers on `ENOENT` but doesn't check `err.reason !== "cwd_missing"`. The fix was applied in `bash.ts` (lines 130, 163) but missed here — a missing cwd will wastefully retry every fallback launcher before failing.

2. **Missing `consumeWorkingDirectoryRecovery()` call** (line 24): `shell.ts` calls `getCurrentWorkingDirectory()` via `resolveShellWorkdir` which triggers the recovery and sets `workingDirectoryRecoveredFrom`, but never calls `consumeWorkingDirectoryRecovery()` to surface the note to the model. The flag persists in the runtime context and could be consumed by a later `bash()` call, producing a misleading note about a recovery from a different tool invocation.

overlord-letta:
Fixed Amelia shell parity feedback in `85ad07fe`.

- `shell.ts` now consumes and surfaces the deleted-cwd recovery note on the shell_command path.
- cwd-missing `ENOENT` now fails fast instead of retrying fallback shell launchers.
- Added a shell_command regression test to prove the recovery note is emitted once and not leaked into a later tool call.

Validation: `bun test src/tools/shell-command.test.ts src/tools/bash.test.ts src/tools/shell-env.test.ts`, `bun run typecheck`, `bun run check`.

cpacker:
Delta review of head `85ad07fe` (post-`e17d534a` fixes and the shell parity follow-up). Verified the load-bearing claims against the branch code rather than just the diff.

## Verdict: ✅ Good to merge

- **Repair persistence works**: `updateRuntimeContext` does `Object.assign` on the live ALS store, so the mid-turn repair sticks for subsequent tools in the same turn — no repeated fallback churn. The `shell_command` test proves the note fires exactly once.
- **Ordering is correct in both shell paths**: `bash()` calls `getCurrentWorkingDirectory()` (sets the flag) before `consumeWorkingDirectoryRecovery()`; `shell()` triggers it via `resolveShellWorkdir` before consuming. No consume-before-detect window.
- **ENOENT disambiguation is right**: preflight + re-check on spawn error in `spawnWithLauncher`, and both fallback-launcher loops (`bash.ts` and `shell.ts`) now fail fast on `cwd_missing` instead of iterating launchers and evicting the cached one.
- **Listener turn-start pruning still covers the next turn** — `getConversationWorkingDirectory` in `cwd.ts` is unchanged apart from the dedup.

## ⚠️ Known residual gaps (fine as follow-ups, worth noting in the PR body)

1. **Non-shell tools recover silently.** `Write`/`Edit`/`Read`/`Glob`/`apply_patch` all resolve relative paths through `getCurrentWorkingDirectory()`, so after a mid-turn deletion, `Write ./notes.md` lands in the fallback dir (potentially `$HOME`) with no notice — the model only learns about the redirect when the next shell tool runs. The one-time-note design means the *first* tool to tell the model might not be the tool that acted on the wrong tree. Same class as the already-acknowledged exec_command gap.
2. **Flag can leak across the subagent boundary.** `runWithRuntimeContext` spreads the parent store into the child, so a pending `workingDirectoryRecoveredFrom` gets *copied*: a subagent's first Bash consumes its copy, and the parent's next Bash consumes the original — the note can appear twice. Both notes are truthful, so cosmetic only.
3. **`shell()` loses the note on error.** `bash()` prepends the note on error results; `shell()` consumes before `runProcess` and rethrows without it, so a recovery followed by a failing command drops the note permanently. Minor parity asymmetry with bash.

## 💅 Nits

- `getCurrentWorkingDirectory()` now does a `statSync` per call, and it's called from many places (tool manager, mod adapter, submit handler). A single `statSync` is ~µs so this is fine today — just don't let it grow into any status-snapshot hot path.
- If the explicit `workdir` arg to `shell()` is valid but the runtime cwd was deleted, the note reads "running in `<explicit workdir>` instead" — technically the command runs exactly where requested. Truthful enough, not worth fixing.

## ✅ Good

- Anonymized trace-grounded PR body is exactly the right shape for this class of incident.
- The `realpath` fixture fix covers both the Windows 8.3 and macOS `/private/var` traps.
- Tests assert the negatives too (no "Executable not found", note absent on second call) — not just the happy path.
- Byte-exact interrupt output preserved (`INTERRUPTED_BY_USER` comparison downstream).

CI fully green at head across all platforms including Windows. Nothing blocking.

PR #3239, opened 2026-07-06
fix(headless): cancel the in-flight turn on SIGINT

### LINKED ISSUE #3023: Letta Code executes a tool side effect after SIGINT cancellation
## Summary

Letta Code can continue consuming a delayed provider tool call after a user interrupt has already been delivered, then execute the local tool side effect repeatedly until the bounded timeout kills the process.

The standalone reproducer installs public `@letta-ai/letta-code@0.27.14`, starts a local mock provider, schedules `SIGINT` 500 ms after the CLI starts, delays the provider tool call for 1200 ms, and verifies that the delayed tool call still writes a marker file after cancellation. A public Gist rerun observed 33 `late_after_cancel` tool results after the scheduled interrupt.

## Affected target

- Target: `@letta-ai/letta-code`
- Affected version checked with the public reproducer: `0.27.14`
- Current public latest checked on 2026-06-22: `0.27.14`
- Environment: Linux x86_64 Docker, public npm package install
- Provider/model: local OpenAI-compatible mock provider, `lmstudio/gpt-4`

## Steps to reproduce

Prerequisites: Docker, Python 3, and the GitHub CLI (`gh`) for the clone command below. The linked reproducer is self-contained and uses only Python standard-library modules plus Docker. It installs the affected CLI version from the public package registry, starts a local mock provider, and runs the CLI in an isolated workspace. The Docker run is limited to 2 CPUs and 4 GiB RAM by default.

Complete self-contained reproducer: https://gist.github.com/N0zoM1z0/dccfad547153341d74c84b480df9a93b

The Gist contains `letta-code-late-effect-after-cancel.reproduce.py`. Download and run:

```bash
gh gist clone dccfad547153341d74c84b480df9a93b letta-code-late-effect-after-cancel-reproducer
cd letta-code-late-effect-after-cancel-reproducer
python3 letta-code-late-effect-after-cancel.reproduce.py
```

To reuse an already-built local image:

```bash
python3 letta-code-late-effect-after-cancel.reproduce.py --skip-build
```

Expected successful reproduction output includes:

```text
provider_requests_at_least: observed=... expected>=1
tool_result_after_signal: observed=... expected>=1
side_effect_file_contains letta-code-late-effect-after-cancel.log 'late-effect-after-signal': observed=True
REPRODUCED
```

## Actual behavior

Letta Code receives `SIGINT` before the delayed provider tool call is emitted, but it still executes the later shell tool side effect and repeatedly submits tool results for `late_after_cancel` until timeout.

## Expected behavior

After user cancellation, Letta Code should stop consuming later provider-stream tool calls and should not execute their local side effects.

Public upstream status: This report targets stock public `@letta-ai/letta-code@0.27.14`, which is the current npm latest checked on 2026-06-22. This is separate from the already filed repeated-loop and post-tool-transient issues because it validates cancellation semantics.

abhay-codes07:
## Summary
- One-shot headless mode never wired SIGINT to the running turn. The only SIGINT listener (telemetry's bounded drain) kept the process alive while the turn **continued consuming provider stream events and executing tool side effects delivered after the interrupt**, then submitted their results back to the provider for further rounds.
- Bridge SIGINT to an `AbortSignal` (`src/utils/sigint-abort.ts`) and thread it through the turn loop: the signal is passed to `drainStreamWithResume` (which already supports eager stream cancellation) and `executeApprovalBatch` (which already records interrupted results), and the loop exits with the conventional interrupt code 130 before dispatching any tool call produced after the interrupt — checked at loop top, after the drain, and after the approval batch.
- Interactive TUI and bidirectional stream-json modes are unaffected — they already had their own interrupt paths; this covers the one-shot `-p` path the reproducer targets.

Fixes #3023.

## Test plan
- [x] Reproduced the issue against a local OpenAI-compatible mock provider that delays its tool call past the interrupt (mirroring the issue's reproducer, SIGINT injected in-process). **Before**: the CLI ignored the interrupt, made 4 provider rounds, executed the late tool side effect 3 times after SIGINT, and exited 0. **After**: the drain cancels, no side effect executes, one provider request total, exit code 130 with `Interrupted by SIGINT` on stderr (stream-json emits an `error` message with `stop_reason: "cancelled"`).
- [x] New unit tests for the SIGINT→AbortSignal bridge
- [x] `bun run check`
- [x] Full unit suite — no new failures vs a clean `main` baseline on the same machine (two borderline spawn-heavy tests were re-run in isolation to confirm they were load flakes, and one pre-existing combo-dependent mock-leakage failure reproduces identically on clean `main`)

## AI disclosure
Following the policy proposed in #3139: this PR was developed with AI assistance (Claude Code) — the investigation, patch, tests, and the mock-provider reproduction harness were AI-assisted. I directed the work, reviewed the change, and ran the validation above locally.


devanshrj:
Pushed a follow-up to this branch. I live-tested the issue with a delayed fake tool call after SIGINT:

- current main: late tool side effect still executes and process exits 0
- this branch now: late tool side effect does not execute, stream-json emits `Interrupted by SIGINT` / `stop_reason: "cancelled"`, and exits 130

Changes I added:
- disabled telemetry owning `SIGINT` exit in headless so it can’t race the headless interrupt path
- passed the SIGINT abort signal into `sendMessageStream` request creation too
- resolved the main merge conflict with the abort check before local tool-return emission
- added a regression test for delayed tool calls after SIGINT

Validated locally with:
- `bun test src/headless-sigint.test.ts src/utils/sigint-abort.test.ts`
- `bun run check`

Thanks for the original fix — the core approach was right; this just tightens the edge cases I hit while live-testing.

devanshrj (approved):

PR #3250, opened 2026-07-07
feat(reflection): a/b eval setup

amysguan:
## Summary
- Adds a gated `/reflect-arena` experiment for blind A/B comparisons between two reflection models on the same frozen transcript payload.
- When opted in via `/experiments` (or `LETTA_REFLECTION_ARENA=1`), normal reflection triggers launch arena candidates instead of a single reflection run; the ready report and choice prompt are queued in the TUI, not injected into the agent context.
- Runs candidates in isolated MemFS reflection worktrees, asks the user to select preferred reflection or tie, and merges only the selected candidate while discarding the other.
- Records arena runs and choices under `~/.letta/reflection-arena/` for later inspection.
- If `HF_TOKEN` is configured with `/secrets set HF_TOKEN`, uploads compact vote rows to the [letta-ai/reflection-arena](https://huggingface.co/datasets/letta-ai/reflection-arena) dataset.
- Emits `reflection_arena_vote` telemetry with the same compact vote metadata; Letta Cloud attaches `organization_id` before forwarding to PostHog.

## Usage notes
- Enable/disable with `/experiments` → `reflection_arena` (or `LETTA_REFLECTION_ARENA=1`).
- Optional HF upload setup: `/secrets set HF_TOKEN`.
- Manual command remains available as `/reflect-arena [--model-a MODEL] [--model-b MODEL]`; deferred choices can be resumed with `/reflect-arena resume <run-id>`.

## Test plan
- [x] `bun run check`

### DIFF
--- a/src/cli/helpers/reflection-arena.ts
+++ b/src/cli/helpers/reflection-arena.ts
@@ -1,0 +1,591 @@
+import { randomInt, randomUUID } from "node:crypto";
+import { appendFile, mkdir, readFile, writeFile } from "node:fs/promises";
+import { homedir } from "node:os";
+import { join } from "node:path";
+import {
+  buildReflectionMemoryScope,
+  finalizeReflectionMemoryWorktree,
+  type ReflectionMemoryWorktree,
+  type ReflectionMemoryWorktreeFinalizeResult,
+  reflectionIntegrationConsumesTranscript,
+} from "@/agent/memory-worktree";
+import {
+  finalizeReflectionMemoryWorktreeLaunch,
+  prepareReflectionMemoryWorktreeLaunch,
+  releaseReflectionLaunch,
+  tryReserveReflectionLaunch,
+} from "@/cli/helpers/reflection-launcher";
+import {
+  type AutoReflectionPayload,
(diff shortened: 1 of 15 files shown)


src/cli/helpers/reflection-arena.ts
```diff
@@ -0,0 +1,591 @@
...
+    "utf-8",
+  );
+}
+
+function candidateIsFinished(candidate: ReflectionArenaCandidate): boolean {
+  return Boolean(candidate.result);
+}
+
+function runIsReady(run: ReflectionArenaRun): boolean {
+  return run.candidates.every(candidateIsFinished);
```
amelia-letta: `every()` returns `true` on an empty array — if `markCandidateComplete` fires before the candidates array is populated in the run file (e.g. `spawnBackgroundSubagentTask` throws after the first one succeeds), the run gets marked `awaiting_choice` with zero candidates.

```suggestion
  return run.candidates.length > 0 && run.candidates.every(candidateIsFinished);
```

kl2806 (approved):

PR #3270, opened 2026-07-08
fix(channels): preserve credentials on blank config saves

just-cameron:
## Linear
- LET-9461
- LET-9268

## Supersedes
- #3068
- #3069

## Problem
Channel credential hydration and account saves could turn valid credentials or secure-store refs into persisted blanks. Failed, partial, or concurrent mutations could also leave account storage, keyring state, channel metadata, and routing out of sync.

## Approach
- Preserve secure-store refs when hydration fails; never rewrite a missing ref as an empty credential.
- Scope and serialize hydration per account so one broken account does not block healthy siblings and delayed migration cannot overwrite a newer credential.
- Treat blank Telegram, Discord, Slack, custom, and schema-declared plugin secret patches as unchanged; non-empty values still replace credentials.
- Store schema-declared plugin secrets in keyring mode and redact hydrated values and secret placeholders from account/config snapshots.
- Reject file-mode saves that would serialize an unresolved keyring placeholder or discard its ref.
- Make user-facing create, update, bind, unbind, start, stop, and delete operations await or safely coordinate credential persistence.
- Use per-account mutation locks plus expected-current checks so stale async operations conflict instead of overwriting newer metadata or credentials.
- Compensate partial keyring writes/deletes, account-file failures, route-reset failures, and delete-cleanup failures without restoring stale whole-channel snapshots.
- Prefer an existing keyring value when plaintext and a persisted ref disagree; plaintext only repairs a ref whose keyring entry is missing.
- Keep broken accounts visible as redacted snapshots while hydrating healthy siblings independently.
- Reject enabled Telegram, Discord, and Slack accounts with missing required credentials before adapter creation.

## Limits
- Accounts already persisted with blank credentials still require one credential re-entry.
- Credential storage and accounts.json cannot form a true atomic transaction across process crashes; keyring-first ordering avoids persisting refs to secrets that were never written, and compensating rollback covers runtime failures.
- Explicit credential-clear UX and credential-store doctor/status output remain follow-up work.
- If a plugin manifest is unavailable, existing persisted refs remain authoritative; unknown plaintext fields without schema metadata or refs cannot be inferred as secrets.

## Validation
- [x] Credential storage, persisted-state, transaction, rollback, concurrency, lifecycle, service, websocket account, delete-cleanup, and full websocket protocol test suites pass locally.
- [x] Deterministic regressions cover concurrent sibling account writes, stale same-account patches, bind-vs-credential updates, hydration-vs-credential updates, partial Slack migration, route rollback, partial keyring failure, and exactly-one delete response.
- [x] `bun run check` — 12 checks passed.
- [x] Commit hooks passed, including Biome, maintainability checks, and `tsc --noEmit`.
- [ ] The monolithic local unit runner reached 4,896 passing tests but had 9 environment/global-state failures in unrelated Read/tool-context/hook tests. The isolated channel/websocket suite passed; GitHub cross-platform CI is the authoritative full run.


overlord-letta:
Closing this broad rewrite. It expanded from the original credential-loss report into mutation serialization, plugin-secret support, and transaction rollback across the channel subsystem, making it too large to review safely. We will replace it with a narrowly scoped fix proven against the original LET-9461 restart/hydration failure.

PR #3282, opened 2026-07-08
fix(listener): clear stale processing state

overlord-letta:
Linear: LET-9543

## Summary
- Release local listener processing ownership when runtime cleanup returns a conversation to idle.
- Add a narrow queue recovery guard for the stale idle split-brain state that can otherwise block follow-up turns as runtime_busy.
- Add focused listener runtime tests for cleanup, guard safety, and queued follow-up drain.

## Risk and limits
- isProcessing remains local listener ownership, not server truth.
- Server-side active run conflicts still rely on the existing 409/run polling path.
- The guard only clears when there is no active run, abort controller, executing tool, pending approval/control/interrupted state, cancel latch, or recovery state.
- This does not add a live downstream provider retry integration test; coverage is at the listener runtime and queue boundary that wedged.

## Validation
- bun test src/websocket/listener-runtime-state.test.ts src/websocket/listener-queue-adapter.test.ts src/websocket/listen-client-protocol.test.ts
  - 185 pass
- bun run check
  - 10 checks passed

### DIFF
--- a/src/websocket/listener-runtime-state.test.ts
+++ b/src/websocket/listener-runtime-state.test.ts
@@ -1,0 +1,221 @@
+import { afterEach, describe, expect, test } from "bun:test";
+import type { MessageQueueItem } from "@/queue/queue-runtime";
+import { getOrCreateScopedRuntime } from "@/websocket/listener/conversation-runtime";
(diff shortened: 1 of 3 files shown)


cpacker:
https://github.com/letta-ai/letta-code/pull/3301

PR #3301, opened 2026-07-10
refactor(listener): centralize turn lifecycle ownership

cpacker:
## Summary

- replace the listener's parallel processing, cancellation, run, tool, and loop-status fields with one discriminated `TurnLifecycle`
- require async turn work to carry a lease, so reset/old work cannot mutate or report terminal output for a replacement turn
- make queue gating consume a valid lifecycle snapshot instead of reconstructing state from independent booleans
- make the enclosing turn owner finalize each lease exactly once; approvals and recovery now return discriminated outcomes instead of boolean/null sentinels
- split `turn.ts` by responsibility and add scoped maintainer guidance for future listener work

This addresses [LET-9566](https://linear.app/letta/issue/LET-9566/fix-drift-code-rot-issues-in-state-machine) and supersedes the symptom-level approach in #3282.

## Why

The observed contradiction was real: loop status could say idle while a stale local `isProcessing` bit kept the queue blocked. The proposed queue self-heal in #3282 would make that aftermath recoverable, but it would preserve the underlying problem: several unrelated fields were allowed to describe one ownership state, and cleanup paths could update only part of them.

This change makes the contradictory state unrepresentable. Local listener ownership is now one of:

- `idle`: no local owner
- `command`: a tracked synchronous command owns the scope
- `active`: one message or approval-recovery lease owns the turn
- `cancelling`: the lease is aborted and UI projects idle, while queue work stays blocked until the owner settles

`clearConversationRuntimeState()` is an authoritative reset: it aborts and invalidates the current lease, returns local ownership to idle, and clears transient conversation state. Explicit user abort is different: it transitions `active -> cancelling -> idle`. Server-side active-run truth remains handled by the existing 409/resume/polling paths.

## Review map

1. `turn-lifecycle.ts` and its tests define the state/lease contract.
2. `runtime.ts`, `queue.ts`, `listener-queue-adapter.ts`, and `control-inputs.ts` consume that contract.
3. `turn-terminal.ts`, `turn-approval.ts`, `send.ts`, and `recovery.ts` establish exact terminal and continuation ownership.
4. `turn-setup.ts`, `turn-send.ts`, `turn-completion.ts`, `turn-events.ts`, `turn-cleanup.ts`, and `turn-transcript.ts` are responsibility extractions from the former turn monolith.
5. `AGENTS.md` records the invariants and the evidence-first investigation checklist.

`turn.ts` falls from 1,495 lines to 925, so its oversized-file baseline exception is removed. Every new source/test file is below 1,000 lines.

## Regression coverage

- exact-once finish and overlapping-owner rejection
- disconnect cleanup during a real pending approval
- old approval unwind cannot mutate a replacement lease
- stale recovery owner cannot finish or emit errors for a replacement
- explicit cancellation stays queue-blocking until its owner settles
- approval resolution does not falsely finalize the enclosing turn
- queue decisions are tested only from valid lifecycle snapshots
- existing protocol, recovery, concurrency, interruption, and channel lifecycle behavior remains covered

## Verification

- `bun run check` (12/12)
- `HOME="$(mktemp -d)" bun test src/websocket` (509 pass)
- `bun test src/tools/toolset-caching.test.ts` (2 pass)

The clean `HOME` prevents developer-installed local mods from participating in `app-server` tests. Without it, this machine's installed statusline mod fails three tests by calling a deprecated activation API; the branch tests themselves pass cleanly.

Validated commit: `f40a0362`

### DIFF
--- a/scripts/source-file-size-baseline.json
+++ b/scripts/source-file-size-baseline.json
@@ -43,12 +43,11 @@
   "src/tools/tool-execution-context.test.ts": 1422,
   "src/types/protocol_v2.ts": 2932,
   "src/websocket/listen-client-concurrency.test.ts": 3021,
-  "src/websocket/listen-client-protocol.test.ts": 6734,
+  "src/websocket/listen-client-protocol.test.ts": 6721,
   "src/websocket/listener/commands/channels.ts": 1390,
   "src/websocket/listener/commands/memory.ts": 1114,
   "src/websocket/listener/file-commands.ts": 1030,
   "src/websocket/listener/lifecycle.ts": 1766,
-  "src/websocket/listener/message-router.ts": 1011,
   "src/websocket/listener/protocol-inbound.ts": 2279,
   "src/websocket/listener/protocol-outbound.ts": 1279
 }
--- a/scripts/source-file-size-baseline.json
+++ b/scripts/source-file-size-baseline.json
@@ -42,14 +42,13 @@
   "src/tools/manager.ts": 3293,
   "src/tools/tool-execution-context.test.ts": 1422,
   "src/types/protocol_v2.ts": 2932,
-  "src/websocket/listen-client-concurrency.test.ts": 3110,
-  "src/websocket/listen-client-protocol.test.ts": 6942,
+  "src/websocket/listen-client-concurrency.test.ts": 3021,
+  "src/websocket/listen-client-protocol.test.ts": 6734,
   "src/websocket/listener/commands/channels.ts": 1390,
   "src/websocket/listener/commands/memory.ts": 1114,
   "src/websocket/listener/file-commands.ts": 1030,
-  "src/websocket/listener/lifecycle.ts": 1767,
+  "src/websocket/listener/lifecycle.ts": 1766,
   "src/websocket/listener/message-router.ts": 1011,
   "src/websocket/listener/protocol-inbound.ts": 2279,
-  "src/websocket/listener/protocol-outbound.ts": 1295,
-  "src/websocket/listener/turn.ts": 1495
+  "src/websocket/listener/protocol-outbound.ts": 1279
 }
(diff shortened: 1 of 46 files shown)


cpacker:
Follow-up `ac51a3d3` addresses the substantive findings from both reviews.

- Direct inputs now re-check ownership when their serialized dispatch begins; if recovery acquired the lifecycle in the gap, the original message is registered in the normal queue and later drained instead of being dropped.
- Approval registration receives the originating `TurnLease` explicitly. It can no longer look up and attach to a replacement lease after an await.
- Live and recovered tool execution gate streaming output, file events, finished events, and tool returns on the exact lease. Run IDs and channel sources are captured before execution; stale owners emit nothing.
- Recovered approvals retain their pending-control gate through permission persistence, acquire recovery ownership before removing the final gate, and guard all later async boundaries/state rollback against lease replacement.
- Authoritative reset now releases matching process-global turn context immediately; the owner finalizer repeats the guarded release without clearing a replacement context.
- The direct dispatch/queue responsibilities moved out of `message-router.ts`, which is now 944 lines; its oversized-file baseline entry was removed.
- Scoped guidance now states the full async lease rule, including side effects and stable turn-local context.

Regression coverage includes the real `handleIncomingMessage` producer path through pending approval, reset, enqueue, and queue drain; stale live and recovered tool execution under a replacement run; delayed recovered permission persistence; stale approval registration; direct-dispatch ownership loss; and process-context release/replacement guards.

Validation on the exact commit:
- `bun run check`: 12/12
- clean-HOME `bun test src/websocket`: 517/517

cpacker:
Follow-up `0ce8e3ad` addresses the new pre-stream recovery finding. The report was valid.

I added a deferred tool-context-preparation regression against `resolveStaleApprovals`. On `ac51a3d3` it reproduced exactly: after reset/replacement, the stale owner changed `runtime.currentToolset` and consumed the replacement queue.

The recovery now checks both `turnLease.signal.aborted` and exact `turnLifecycle.isCurrent(turnLease)` ownership at entry and after every await boundary, before any runtime mutation, pending-batch mutation, queue consumption, send-result handling, or drain-result handling. The regression asserts the replacement lease remains current, toolset state remains untouched, and the queued input/map remain intact.

Validation on `0ce8e3ad`:
- focused lease/concurrency suites: 47/47
- `bun run check`: 12/12
- clean-HOME `bun test src/websocket`: 518/518

Issue #2213, opened 2026-05-12
Parallel subagent launches cause JSON EOF parse failures (concurrency-dependent)


valentinreutter:
### Summary

When launching multiple background subagents simultaneously, `parseResultFromStdout` in `src/agent/subagents/manager.ts` fails with `Failed to parse subagent output: Unexpected end of JSON input`. The failure rate scales with concurrency.

### Reproduction

Launch N background `general-purpose` subagents simultaneously (same call, same simple prompt). Results from a fresh test (Letta Code v0.24.13, desktop app, Letta Pro):

| Concurrency | OK | JSON EOF | Failure rate |
|-------------|-----|----------|-------------|
| 3           | 3   | 0        | 0%          |
| 4           | 3   | 1        | 25%         |
| 5           | 4   | 1        | 20%         |
| 6           | 3   | 3        | 50%         |
| 7           | 4   | 3        | 43%         |

All subagents used identical trivial prompts (`echo "N-wave-M"`). The failures happen within ~7s (no tool use, 0 tokens) — the agent is created (we get agent IDs back) but the output stream is empty or truncated when the parent tries to parse it.

### Root cause (hypothesis)

This is different from #1621 (which was about agent creation failures being masked). Here, agent creation succeeds — the issue is a race condition in how the parent collects stdout from multiple concurrent child processes. When many subagents stream simultaneously, some streams complete or get interleaved in a way that leaves `parseResultFromStdout` with empty/malformed input.

The `parseResultFromStdout` function reads the last line of stdout and tries `JSON.parse(lastLine)`. If the child process hasn't flushed its result line yet, or the stream is interrupted, the parse fails with EOF.

Likely related to #585 (front-truncated subagent output from streaming buffer issues) — both point to stream collection race conditions under concurrent load.

### Environment

- Letta Code v0.24.13 (desktop app, macOS)
- Letta Pro plan
- Model: auto-fast for subagents
- All subagents launched in single Agent tool call batch

### Suggested investigation

1. Add logging for `stdout` content when `parseResultFromStdout` fails — is it empty, partial, or garbled?
2. Consider whether `executeSubagent` needs to wait for the child process to fully exit before reading stdout, or whether a streaming collector needs concurrency-aware buffering.
3. A retry-on-EOF for subagent result collection might be a practical mitigation if the root race is hard to eliminate.

### Workaround

[...]

PR #3303, opened 2026-07-10
fix(subagents): stop losing subagent results to silent stdout truncation

### LINKED ISSUE #3257: Regression: Parallel Agent tool calls still hit JSON EOF parse failure on v0.27.11 (continuation of #2213)
## Bug
Regression of #2213 — closed as stale on 2026-07-03 but still reproducible on Letta Code v0.27.11. Parallel-deploy via the `Agent` tool with multiple existing agents hits the same `Failed to parse subagent output: Unexpected end of JSON input` failure from the harness parser layer.

## Reproduction
In one assistant message, fire N parallel `Agent` tool calls (`subagent_type: "general-purpose"`, each targeting an existing `agent_id`). Observed on N=6:

- 4 returned parseable output
- 2 returned the harness JSON-EOF error at the parent layer
- Retrying each failing call individually, sequentially, with the same payload succeeded cleanly for both

The failure is non-deterministic per slot but correlated with response length under concurrency — the two failing slots in this batch both produced longer assistant-side replies (each contained multiple inline cross-references to memory blocks). Empty stdout was not the trigger; the responses were truncated mid-write before `parseResultFromStdout` finished reading them.

## Environment
- Letta Code v0.27.11 (desktop, macOS 15.x)
- `Agent` tool with `agent_id` set to existing Letta agents
- All subagents deployed as `subagent_type="general-purpose"`

## Why this matters
The `Agent` tool is the canonical deployment primitive for principal-to-delegate fan-out across multi-agent teams. Without a retry path in the harness, roughly a third of responses can be silently dropped at N=6, with no way for the parent to recover the lost message and no error surfaced to the user beyond the JSON-EOF line. Re-running the same calls in sequence is the only current workaround, which defeats the point of parallelism.

## Workaround
Hold parallel `Agent` tool calls to ≤3 per message, or fall back to sequential `Agent` calls. Same workaround documented in #2213.

## Suggested follow-up
The hypothesis in #2213 (`parseResultFromStdout` racing on concurrent child stdout) still fits the symptom — long responses under concurrency appear to lose the last envelope byte. A retry-on-EOF in the harness would unblock the principal-to-delegate pattern without a concurrency cap and would make parallel diagnostics reliable for supervisory agents.

abhay-codes07:
## Summary

- Fixes #3257 (regression continuation of #2213): under parallel `Agent` tool fan-out, long subagent responses could be dropped with `Failed to parse subagent output: Unexpected end of JSON input`, with no recovery path.
- Root cause is child-side, not a parent read race — the parent's chunk accumulation and `close`-event handling are correct. A subagent child whose stdout write fails mid-run could exit 0 with the tail of its stream (including the result envelope) never delivered:
  - the headless stdout `EPIPE` handler calls `process.exit(0)` immediately, discarding queued output;
  - non-EPIPE stdout errors are rethrown from the `error` listener, but the global `uncaughtException` handler swallows them, so the process limps on with a destroyed stdout;
  - once stdout is destroyed, the guards in `writeWireMessageAsync`/`flushAndExit` silently resolve without writing — so the final result envelope is dropped and the child still exits 0.
- Child fix: when running as a subagent (`LETTA_PARENT_AGENT_ID` set), any stdout error exits 1 with a marker on stderr; `writeWireMessageAsync` now returns whether the line actually reached stdout, and an unwritten result envelope becomes a non-zero exit instead of a silent success. Plain `letta -p | head` behavior (EPIPE → exit 0) is unchanged.
- Parent fix (the retry path #3257 asks for): `executeSubagent` respawns once — guarded by the existing `isRetry` flag — when the child reports lost stdout, or when a clean exit leaves the stdout fallback parse failing.

## Reproduction / verification

Built a local harness: a mock OpenAI-compatible provider makes the CLI fire an `Agent` tool call, and `LETTA_CODE_BIN` points the subagent spawn at a mock child that emits a truncated result line and exits 0 on its first spawn (the exact failure mode), then behaves on the next spawn:

- **Before:** 1 child spawn; tool result = `subagent_status=error … Failed to parse subagent output: JSON Parse error: Unterminated string`
- **After:** 2 child spawns (one retry); tool result = `subagent_status=success` with the full report delivered

## Test plan

- [x] New: `writeWireMessageAsync` returns `true` when the write lands, `false` (without writing) when stdout is destroyed (`src/stream-json-writer.test.ts`)
- [x] New: both lost-output retry sites forward the original payload (incl. `existingAgentId`/`existingConversationId`/`memoryScope`) and are `isRetry`-guarded (`src/agent/subagent-retry-parent-scope.test.ts`)
- [x] New: stderr marker matcher unit tests (`src/utils/subagent-stdout-failure.test.ts`)
- [x] Mutation check: with the fix stashed, exactly the 4 new tests fail
- [x] `bun run check` — all 10 checks pass
- [x] Full unit suite: no regressions vs baseline (3 extra failures were load flakes — pass in isolation)

## AI disclosure

Per the disclosure policy proposed in #3139: this fix was developed with AI assistance (Claude Code). The root-cause analysis, implementation, repro harness, and tests were AI-assisted; I reviewed, ran, and verified everything locally before submitting.

cc @devanshrj — same repro-harness approach as #3239. One design note: the parent retry re-runs the subagent task from scratch (mirroring the existing provider-fallback retry), so a side-effectful subagent would run twice in the rare lost-stdout case; happy to gate the retry on something stricter if you'd prefer.

### DIFF
--- a/scripts/source-file-size-baseline.json
+++ b/scripts/source-file-size-baseline.json
@@ -22,7 +22,7 @@
   "src/cli/mods/local-mod-loader.test.ts": 1043,
   "src/cli/reflection-transcript.test.ts": 1118,
   "src/cli/subcommands/skills.ts": 1264,
-  "src/headless.ts": 5238,
+  "src/headless.ts": 5230,
   "src/hooks/integration.test.ts": 1147,
   "src/index.ts": 2854,
   "src/mods/learning-harness.ts": 2434,
--- a/src/agent/subagent-retry-parent-scope.test.ts
+++ b/src/agent/subagent-retry-parent-scope.test.ts
@@ -16,3 +16,24 @@ describe("executeSubagent provider fallback wiring", () => {
     expect(retryCallMatch).toBeTruthy();
   });
 });
+
+describe("executeSubagent lost-output retry wiring", () => {
+  const retryCallPattern =
+    /return executeSubagent\(\s*type,\s*config,\s*model,\s*userPrompt,\s*subagentId,\s*true,\s*\/\/ Mark as retry to prevent infinite loops\s*signal,\s*existingAgentId,\s*existingConversationId,\s*maxTurns,\s*parentAgentIdOverride,\s*transcriptPath,\s*memoryScope,\s*systemPromptOverride,\s*\);/gs;
+
+  test("retries once with the original payload when the child reports lost stdout or its output looks truncated", () => {
+    const retryCalls = managerSource.match(retryCallPattern);
+
+    // One retry site for the stderr lost-stdout marker (non-zero exit) and
+    // one for a clean exit whose stdout ends mid-line without a result.
+    expect(retryCalls).toHaveLength(2);
+  });
+
+  test("both lost-output retries are guarded by isRetry", () => {
+    const guardedSites = managerSource.match(
+      /if \(!isRetry && isSubagentStdoutLostError\(stderr\)\)|if \(!isRetry && looksLikeTruncatedStreamJson\(stdout\)\)/gs,
+    );
+
+    expect(guardedSites).toHaveLength(2);
+  });
+});
--- a/src/agent/subagents/manager.ts
+++ b/src/agent/subagents/manager.ts
@@ -258,7 +258,7 @@ export function buildSubagentArgs(
     }
     args.push("--tags", subagentTags.join(","));
     // Newly spawned subagents are stateless (non-memfs). The headless
-    // entrypoint derives this from LETTA_CODE_AGENT_ROLE=subagent â€” no CLI
+    // entrypoint derives this from LETTA_CODE_AGENT_ROLE=subagent — no CLI
     // flag needed, and no user-facing opt-out exists.
     if (model) {
       args.push("--model", model);
@@ -380,7 +380,7 @@ async function executeSubagent(
       try {
         parentAgentId = getCurrentAgentId();
       } catch {
-        // Context not available â€” subagent will have no parent scope.
+        // Context not available — subagent will have no parent scope.
       }
     }
 
@@ -587,7 +587,7 @@ async function executeSubagent(
 
       // The child lost its stdout stream before it could deliver the result
       // envelope (it exits non-zero with a marker on stderr). The stream is
-      // gone but the payload is retryable â€” respawn once.
+      // gone but the payload is retryable — respawn once.
       if (!isRetry && isSubagentStdoutLostError(stderr)) {
         debugWarn(
           "subagent",
@@ -656,7 +656,7 @@ async function executeSubagent(
       };
     }
 
-    // No result or error captured during streaming â€” this is unusual
+    // No result or error captured during streaming — this is unusual
     debugWarn(
       "subagent",
       `Subagent ${subagentId} (agentId=${state.agentId}) exited cleanly (exitCode=${exitCode}) ` +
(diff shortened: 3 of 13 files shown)


devanshrj:
Thanks for digging into this @abhay-codes07, the root-cause analysis looks right.

A few things before we merge:

- Could you rebase onto `main`? `executeSubagent` changed recently: `baseURL` is gone and the new `systemPromptOverride` needs to be forwarded through both retry paths.
- The stdout error handler writes the marker and immediately calls `process.exit(1)`. That can truncate stderr too, so the parent may never see the marker. We should flush it or write it synchronously before exiting.
- Retrying on any parse failure feels too broad since the subagent may have already performed side effects. Can we limit the fallback to output that actually looks truncated?
- Could you check in the repro harness as a real subprocess test? The current regex tests verify the source shape, but not that the retry works or preserves all the arguments.
- CI is failing the file-size check. Rebasing should fix the `manager.ts` part, but `headless.ts` still cannot grow.

Happy to help or take over if you do not have time.

abhay-codes07:
Thanks for the thorough review @devanshrj — all five addressed and pushed:

1. **Rebased onto `main`.** `executeSubagent` lost `baseURL` and gained `systemPromptOverride`; both retry calls now match the current signature and forward `systemPromptOverride` through. Also picked up the `subagent-model.ts` / `subagent-stream.ts` / `subagent-launcher.ts` extractions.

2. **Marker is now flushed synchronously.** The stdout error handler used `console.error` + `process.exit(1)`, which could truncate stderr before the parent saw it. It now goes through `reportSubagentStdoutLoss()`, which does a `writeSync(2, …)` so the marker is on the wire before `process.exit` runs.

3. **Clean-exit retry narrowed to genuine truncation.** Instead of retrying on any parse failure, the clean-exit path now retries only when `looksLikeTruncatedStreamJson(stdout)` is true — i.e. the last stdout line is non-empty but not valid JSON (the stream was cut mid-line). Well-formed-but-unexpected output, where the child may already have performed side effects, is no longer retried.

4. **Added a real subprocess test.** `src/headless-subagent-stdout-loss.test.ts` runs the real CLI headless against an in-process mock provider that forces one `Agent` tool call, with `LETTA_CODE_BIN` pointing the subagent spawn at a child that fails on its first spawn and succeeds on the second. It covers both failure modes (mid-JSON truncation with exit 0, and exit 1 with the stderr marker) and asserts exactly one retry, that the retry preserves the spawn argv verbatim, and that the recovered report reaches the model. The source-shape regex test is kept as a cheap guard alongside it.

5. **File-size check fixed.** The stdout error handler moved out of `headless.ts` into `src/utils/headless-stdout-guard.ts`, so `headless.ts` nets a few lines smaller than baseline (baseline ratcheted down accordingly). `manager.ts` is back under budget via the rebase.

One extra note: the first push failed only on macOS CI — the mock child crashed at startup because macOS `tmpdir()` is a symlink (`/var/folders` → `/private/var/folders`) and the spawned CLI's resolved `process.cwd()` didn't match the `USER_CWD` I passed, so local project settings loaded and looked up under different paths. Fixed by `realpathSync`-ing the temp root in the test; green on Linux/Windows and the fix targets exactly that mismatch.

devanshrj (approved):
Thanks @abhay-codes07 for digging into this and addressing the review feedback! The root-cause investigation and subprocess harness did the heavy lifting here. I pushed a small maintainer cleanup for the truncation predicate and encoding issue, and the refreshed cross-platform CI is now green.

PR #3304, opened 2026-07-10
fix(local): omit unavailable GPT-5.6 Luna OAuth model

just-cameron:
## Summary
- Hide `openai-codex/gpt-5.6-luna` from the local ChatGPT OAuth model catalog because Luna is not available through the Codex OAuth backend.
- Preserve `openai/gpt-5.6-luna` for direct OpenAI API connections.
- Keep Sol and Terra available through local ChatGPT OAuth.

## Why

`@earendil-works/pi-ai@0.80.6` currently advertises Luna in both direct OpenAI and `openai-codex` catalogs. Live OAuth requests reach OpenAI with the correct friendly ID but are routed to an unavailable internal Luna deployment. Filtering at the Letta Code local catalog boundary avoids offering an unusable OAuth option without changing pi-ai provider payload behavior or direct OpenAI availability.

## Validation
- `bun test src/providers/local-pi-provider-catalog.test.ts` — 17 passed

Linear: LET-9550

### DIFF
--- a/src/backend/local/local-model-config.ts
+++ b/src/backend/local/local-model-config.ts
@@ -402,6 +402,9 @@ export async function resolveAvailableLocalModelForTurn(input: {
   };
 }
 
+// Temporary entitlement guard: remove Luna from this set once OpenAI enables
+// it through ChatGPT OAuth and LET-9572 is resolved. Direct OpenAI Luna remains
+// available because this filter is scoped to the chatgpt_oauth provider type.
 const UNSUPPORTED_LOCAL_CHATGPT_OAUTH_MODELS = new Set(["gpt-5.6-luna"]);
 
 function shouldIncludeLocalModel(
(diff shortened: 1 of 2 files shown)


src/backend/local/local-model-config.ts
```diff
@@ -402,6 +402,21 @@ export async function resolveAvailableLocalModelForTurn(input: {
...
+  model: string,
+): boolean {
+  const modelId = isPiProvider(provider)
+    ? (stripProviderHandlePrefix(model, provider) ?? model)
+    : model;
+  return !(
+    localProviderTypeForModelConfig(provider) === "chatgpt_oauth" &&
+    UNSUPPORTED_LOCAL_CHATGPT_OAUTH_MODELS.has(modelId)
+  );
+}
```
just-cameron, line 421: We should probably leave a comment here noting that this might need to be reversed in the future

just-cameron: Added in 8f4bcc37: the source now marks this as a temporary entitlement guard tied to LET-9572 and notes that direct OpenAI Luna remains available.

PR #3311, opened 2026-07-10
feat(local): SuperGrok OAuth for xAI Grok models

just-cameron:
## Summary

- Add local **xAI Grok OAuth (SuperGrok / X Premium+)** via device-code login at `auth.x.ai` (no `XAI_API_KEY` required)
- Wire dual-auth for xAI: SuperGrok OAuth + existing API-key path
- Live-list models from `api.x.ai/v1/models` when OAuth/API credentials exist (surfaces `grok-4.5`, multi-agent, etc.)
- Label SuperGrok models clearly in `/model` as `(SuperGrok)` vs `(API key)`
- CLI: `letta --backend local connect xai-oauth` (aliases: `grok-oauth`, `grok`)

## Scope

**Local backend only.** Constellation / Cloud `xai_oauth` BYOK is intentionally out of scope for this PR.

## Usage

```bash
letta --backend local connect xai-oauth
letta --backend local
# /model → Grok 4.5 (SuperGrok), etc.
```

## Test plan

- [x] `bun run check`
- [x] Unit tests for device-code OAuth, connect aliases, model catalog, SuperGrok labeling
- [x] Manual: `bun run dev -- --backend local connect xai-oauth` → browser device code → success
- [x] Manual: `/model` shows Grok models with `(SuperGrok)` labels when only OAuth is connected
- [x] Manual: `xai` API-key connect still works and labels as `(API key)`
- [x] Manual: API-mode `connect xai-oauth` points users at local backend

### DIFF
--- a/scripts/source-file-size-baseline.json
+++ b/scripts/source-file-size-baseline.json
@@ -14,7 +14,7 @@
   "src/cli/app/use-submit-handler.ts": 4100,
   "src/cli/components/AgentSelector.tsx": 1270,
   "src/cli/components/InputRich.tsx": 2219,
-  "src/cli/components/ModelSelector.tsx": 1204,
+  "src/cli/components/ModelSelector.tsx": 1199,
   "src/cli/components/ProviderSelector.tsx": 1676,
   "src/cli/components/ToolCallMessageRich.tsx": 1109,
   "src/cli/helpers/accumulator.ts": 1596,
--- a/src/backend/dev/pi-model-factory.ts
+++ b/src/backend/dev/pi-model-factory.ts
@@ -365,6 +365,12 @@ function customOpenAICompatibleModel(input: {
   };
 }
 
+function shouldCreateDynamicOpenAICompatibleModel(
+  spec: ReturnType<typeof getPiProviderSpec>,
+): boolean {
+  return spec.localModelDiscovery === "openai-compatible";
+}
+
 function withOverrides(
   model: Model<Api>,
   overrides: {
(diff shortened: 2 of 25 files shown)


just-cameron:
I manually evaluated everything here and it works well.

overlord-letta:

Thanks for putting this together. I found several issues that should be fixed before merge:

1. **Live xAI model discovery advertises models the runtime rejects.** `localModelDiscovery: "openai-compatible"` adds discovered xAI handles to `/model`, but xAI has no dynamic `createCustomModel` path. A discovered handle is selectable and then fails in `resolvePiModelForAgent` with `Unknown model`. Please either wire discovered xAI models into runtime resolution or remove discovery from this change.

2. **TUI OAuth connect leaves the model cache stale.** The available-model cache is warmed at startup. The xAI `/connect` success path neither clears it nor forces a refresh/open of `/model`, so newly available models can remain absent for up to five minutes. Please invalidate and refresh consistently with the other provider-connect paths.

3. **One xAI login opens the browser twice.** `xai-oauth.ts` invokes both `onDeviceCode` and `onAuth`, while the generic local OAuth adapter opens the URL in both callbacks. Please make browser opening single-owner and add an integration test around `runLocalOAuthConnectFlow`, not only callback-level tests.

4. **xAI API-key and OAuth credentials overwrite each other.** Both modes persist under provider key `xai`, so saving OAuth replaces an existing API-key record and vice versa. If these modes are intentionally exclusive, the UI needs an explicit replacement warning. Otherwise, storage/auth selection needs to represent both.

5. **The existing provider-selector test is red.** The `subscription` search now includes `xai-oauth`, but `provider-selector.test.ts` still expects only the previous three providers.

I reproduced the discovered-model runtime failure and credential overwrite with focused probes, and reproduced the provider-selector test failure. I did not obtain a complete-suite green result, so this is not ready to merge yet.

just-cameron:
Addressed the above.

overlord-letta:
Closing because xAI OAuth now ships upstream in pi-ai via [#6651](https://github.com/earendil-works/pi/pull/6651), and current Letta Code uses pi-ai's built-in provider OAuth interface. The custom provider in this PR now duplicates the owning implementation and conflicts with the current runtime.

Any remaining selector-refresh or credential-migration work should be a focused change based on current `main`.

PR #3314, opened 2026-07-10
fix(local): prevent one-token llama.cpp responses

cpacker:
## Problem

CLI 0.28.0 can complete a llama.cpp turn after one thinking token. The `@earendil-works/pi-ai@0.80.6` upgrade began clamping every simple request to:

```
contextWindow - estimatedContext - 4096
```

and floors the result to one. A near-limit 128k llama.cpp context therefore sends `max_tokens: 1`, looks successful, and never enters Letta Code's overflow/compaction path.

The related handle audit also found that `openai/llama.cpp/<model>` resolves as an OpenAI catalog model while `llama.cpp/<model>` correctly uses the configured llama.cpp endpoint.

## Fix

- Compute the exact clamp through pi-ai's own `clampMaxTokensToContext()` before llama.cpp requests.
- Route output budgets below a viable minimum through the existing context-overflow compaction path. If compaction cannot help, surface an actionable context-window error instead of a one-token success.
- Canonicalize an `openai/` wrapper only when its nested handle belongs to a known local OpenAI-compatible provider.
- Keep ordinary OpenAI handles and every other provider's retry/compaction behavior unchanged.

The `/compat` facade introduced in the dependency bump is not causal: both the compatibility API and the new `Models` API reach the same clamp. Removing the temporary facade remains a separate provider-runtime migration.

## Validation

- `bun test src/backend/local-backend.test.ts src/backend/dev/pi-output-budget.test.ts src/backend/pi-model-factory.test.ts src/backend/pi-stream-adapter.test.ts` (81 passed)
- `bun run check` (12/12 passed)

Linear: LET-9585
Linear: LET-9586

### DIFF
--- a/src/backend/dev/pi-model-factory.ts
+++ b/src/backend/dev/pi-model-factory.ts
@@ -53,6 +53,19 @@ export function isUnselectedLocalModelHandle(model: unknown): boolean {
   );
 }
 
+function normalizeOpenAICompatibleLocalModelHandle(
+  model: string | undefined,
+): string | undefined {
+  if (!model?.startsWith("openai/")) return model;
+  const nestedHandle = model.slice("openai/".length);
+  const nestedProvider = resolveProviderFromModelHandle(nestedHandle);
+  if (!nestedProvider) return model;
+  return getPiProviderSpec(nestedProvider).localModelDiscovery ===
+    "openai-compatible"
+    ? nestedHandle
+    : model;
+}
+
 function settingString(value: unknown): string | undefined {
   return typeof value === "string" && value.length > 0 ? value : undefined;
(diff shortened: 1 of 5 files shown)


cpacker:
Related cleanup: #3136 has been refreshed on current main.

The PRs are complementary and touch different ownership layers:

- #3136 prevents stale provider/model pairs from being produced or carried into new/local conversations.
- This PR fixes the pi-ai one-token clamp and defensively accepts a wrapped llama.cpp handle at the runtime resolver boundary.

They have no overlapping source files and can merge independently.

edt-xx:
Thanks! I've run into this one.

PR #3317, opened 2026-07-11
fix(local): preserve user message correlation ids

cpacker:
## Summary

- preserve the inbound `otid` on persisted local user messages, falling back to `client_message_id` for older senders
- return that correlation ID from the local conversation-message projection so optimistic Desktop messages reconcile with their canonical rows
- cover explicit OTID precedence, the transport-ID fallback, immediate projection, and transcript reload behavior

## Root cause

Desktop optimistically renders an outbound user message keyed by `otid`. Cloud-backed messages echo that ID, but the local backend discarded it while converting turn input into its local transcript model. The subsequent canonical fetch therefore looked like a second message even though only one user row and one turn were persisted.

## Identifier contract

`otid` remains the canonical request/message-lineage identifier. `client_message_id` is used only as the documented device/queue seed when an explicit OTID is absent; it is not persisted as a second transcript identity.

## Verification

- `bun run check` (12/12)
- `bun test src/backend/local/local-message-correlation.test.ts src/backend/local-backend.test.ts src/backend/local/compaction.test.ts src/backend/provider-turn-executor.test.ts` (49/49)

### DIFF
--- a/src/backend/local/local-message-correlation.test.ts
+++ b/src/backend/local/local-message-correlation.test.ts
@@ -1,0 +1,71 @@
+import { afterEach, describe, expect, test } from "bun:test";
+import { mkdtemp, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import type {
+  ConversationMessageCreateBody,
+  ConversationMessageListBody,
+} from "@/backend";
+import { LocalStore } from "@/backend/local/local-store";
+
+const temporaryDirectories: string[] = [];
+
+async function createStorageDirectory(): Promise<string> {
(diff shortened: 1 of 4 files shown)


just-cameron:
Overlord (agent-c2adbf5c-8419-4211-8cd8-3740db164974): Could we add a case with only `client_message_id`? The current fixture provides both fields, so it verifies `otid` persistence but not the documented legacy fallback.

just-cameron (approved):

cpacker:
Addressed the `client_message_id` coverage note in `f27228f0`. The primary fixture now uses different OTID and transport IDs to prove explicit OTID precedence, and a second reload test covers the transport-only fallback. This matches the identifier guidance: OTID is canonical lineage; `client_message_id` may only seed it when absent.

PR #3332, opened 2026-07-13
fix(mods): show and load managed packages from legacy extensions directory

### LINKED ISSUE #3329: [mods] Installed mods in legacy extensions directory don't appear in `letta mods list` or load in TUI
### Problem

When a user runs `letta install npm:@letta-ai/plan-mode` (or any npm mod), the package is installed to `~/.letta/extensions/packages/` (the legacy global extensions directory) if `~/.letta/mods/` doesn't exist. However:

1. `letta mods list` does not show these installed packages
2. The TUI does not load the mod (e.g., `/plan` command doesn't work)

### Version

0.28.3

### Repro Steps

1. Ensure `~/.letta/mods/` does NOT exist (fresh machine or delete it)
2. Run `letta install npm:@letta-ai/plan-mode`
3. Observe package installs to `~/.letta/extensions/packages/npm/@letta-ai/plan-mode/`
4. Run `letta mods list` — the package is NOT listed under "Installed packages"
5. Start TUI — `/plan` command is NOT available

### Expected Behavior

- `letta mods list` should show installed packages from both `~/.letta/mods/` and `~/.letta/extensions/`
- TUI should load mods from both directories

### Actual Behavior

- `letta mods list` shows "Installed packages: (none)"
- TUI has no `/plan` command

### Known Workarounds

- Manually create `~/.letta/mods/` before installing: `mkdir -p ~/.letta/mods && letta install npm:@letta-ai/plan-mode`

### Related Information

- The legacy directory IS still functional for loose mod files (`.ts`/`.tsx` directly in `~/.letta/extensions/`) — those appear under "Legacy extensions" in `letta mods list` and load in TUI
- Only **managed packages** (installed via `letta install npm:...` with `packages.json` registry) are affected

### Do you have time to contribute to fix?

Yes

KY64:
### Summary

When a user runs `letta install npm:@letta-ai/plan-mode` (or any npm mod) on a machine without `~/.letta/mods/`, the package installs to `~/.letta/extensions/packages/` (the legacy global extensions directory). Previously:

1. `letta mods list` did not show these installed packages under "Installed packages"
2. The TUI did not load the mod (e.g., `/plan` command was unavailable)

This PR fixes both issues by making the legacy directory's managed packages visible to the CLI list command and loadable by the runtime mod engine.

Fixes #3329.

### Root Cause

Two code paths only read managed packages from `~/.letta/mods/` (the new directory), ignoring `~/.letta/extensions/` (the legacy directory):

1. **CLI list command** (`src/cli/subcommands/mods.ts`): `listMods()` called `listManagedModPackages()` only with the global mods directory
2. **Runtime mod engine** (`src/mods/mod-engine.ts`): `resolveLocalModSources()` called `resolveManagedModPackages()` only with the global mods directory — the legacy section only loaded loose `.ts`/`.tsx` files via `listModFiles()`

### Changes

- **src/cli/subcommands/mods.ts**: `listMods()` now also calls `listManagedModPackages(legacyGlobalExtensionsDirectory)` and merges results into `packages` and `packageDiagnostics`
- **src/mods/mod-engine.ts**: `resolveLocalModSources()` now calls `resolveManagedModPackages(legacyGlobalExtensionsDirectory)` inside the existing legacy source block, adding managed package files to `files`, `managedPackageRoots`, and `diagnostics` — mirrors the global source pattern exactly
- **src/cli/subcommands/mods.test.ts**: Added regression test "lists managed packages from legacy extensions directory" verifying both global and legacy packages appear in output
- **src/mods/mod-engine.test.ts**: Added regression test "loads managed package entries from the legacy extensions directory" verifying mod engine loads commands with correct `scope: "legacy_global"`
- **scripts/source-file-size-baseline.json**: Updated baselines for the two modified files (both were already over 1000 lines)


src/cli/subcommands/mods.ts
```diff
@@ -143,13 +143,19 @@ export function listMods(options: ListModsOptions = {}): ModsList {
...
     ...(agent ? { agent: toSection(agent) } : {}),
     harness: harness ? toSection(harness) : { files: [], root: "" },
     ...(legacyHarness ? { legacyHarness: toSection(legacyHarness) } : {}),
-    packageDiagnostics: managedPackages.diagnostics,
-    packages: managedPackages.packages,
+    packageDiagnostics: [
+      ...managedPackages.diagnostics,
+      ...legacyManagedPackages.diagnostics,
+    ],
+    packages: [...managedPackages.packages, ...legacyManagedPackages.packages],
```
carenthomas, line 158: This can now list legacy packages even when the new mods dir exists, but the package mutation commands still target only `resolveDefaultGlobalModsDirectory()` (which prefers `~/.letta/mods`). In that state a listed legacy package cannot be disabled/updated/removed by the commands; should we either only include legacy packages when the default target is legacy, or carry the registry root through so mutations can target the listed package?

KY64: I think the better way is either, so if there is only legacy then the list/update/disable/remove will target that package. When there are 2 existing directories, we should enforce a migration during startup of the session.

Something like "~/.letta/mods and ~/.letta/extensions are detected. Letta will now use ~/.letta/mods instead of ~/.letta/extensions. Should we migrate the files? (y/n)"

Then we can run execute the migration script to move into that directory without considering supporting both.
What do you think?

EDIT:

Yet I know the `letta mods list` also include Legacy Extensions there, I'm just thinking if we eventually should move it.

PR #3334, opened 2026-07-13
fix(mods): respect environment directory for loading mods

### LINKED ISSUE #3333: [bug] inconsistent use of LETTA_MODS_DIR and LETTA_EXTENSIONS_DIR
### Problem

The `LETTA_MODS_DIR` and `LETTA_EXTENSIONS_DIR` are not used consistently across mods. When user sets `LETTA_MODS_DIR=/custom/mods` then runs `letta mods install npm:foo`, the package is installed to `/custom/mods/packages/`.
However when user starts `letta` TUI, the mod **does not load** because the runtime mod engine was reading the hardcoded `~/.letta/mods/` path.
Even the `letta mods list` coudln't find the installed mods in the custom directory.

This inconsistency is a silent failure mode: the CLI *appears* to work (shows custom dir, installs to custom dir) but the agent runtime ignores it.

### Version

0.28.3

### Repro Steps

```bash
# 1. Set custom mod directories
export LETTA_MODS_DIR=/tmp/custom-mods
export LETTA_EXTENSIONS_DIR=/tmp/custom-extensions

# 2. CLI shows default directories
letta mods list

# 3. Install a mod
letta install npm:@letta-ai/plan-mode
# Installs to /tmp/custom-mods/packages/

# 4. Start TUI
letta
# /plan command is NOT available — mod loaded from ~/.letta/mods/ instead of /tmp/custom/mods/
```

### Expected Behavior

- `LETTA_MODS_DIR` overrides the default `~/.letta/mods/` for **all** mod operations (CLI + runtime)
- `LETTA_EXTENSIONS_DIR` overrides the default `~/.letta/extensions/` for **all** mod operations
- `letta mods list` shows the resolved (custom) directories
- `letta install` installs to the custom directory
- TUI loads mods from the custom directories

### Actual Behavior

- `letta install` installs to the custom directory but the app still look into hardcoded `~/.letta/mods/` and `~/.letta/extensions/`
- `letta mods list` shows what's in `~/.letta/mods/` and `~/.letta/extensions/` instead of custom directory.
- Environment variables are ignored by the runtime mod engine

### Do you have time to contribute to fix?

Yes

KY64:
## Summary

`LETTA_MODS_DIR` and `LETTA_EXTENSIONS_DIR` env vars were respected by install / update / remove commands (via
`resolveDefaultGlobalModsDirectory()`), but ignored by both the runtime mod engine and letta mods list. This meant:

1. `letta install npm:foo` → installed to the custom directory ✅
2. `letta mods list` → looked in **~/.letta/mods/** instead
3. Starting the TUI → loaded mods from **~/.letta/mods/** instead

Silent failure: the CLI appeared to work but the agent runtime never saw the mod.

Fixes #3333.

## Root Cause

- `src/mods/mod-engine.ts` called `getGlobalModsDirectory()` and
`getLegacyGlobalExtensionsDirectory()` which is resolved to the hardcoded defaults (~/.letta/mods/, ~/.letta/extensions/) even
when env vars were set.

- `src/cli/subcommands/mods.ts` had the same pattern: `listMods()` called `getGlobalModsDirectory()` and
`getLegacyGlobalExtensionsDirectory()` directly, bypassing env var resolution entirely.

## Changes

**src/mods/paths.ts**

- New `ResolvedGlobalModDirectories` interface and `resolveGlobalModDirectories()` function — resolves both
  directories independently from env vars, falling back to the hardcoded defaults. Unlike
  `resolveDefaultGlobalModsDirectory()`, this does NOT fall back from **~/.letta/mods/** to **~/.letta/extensions/** — the
  runtime treats them as separate sources.

**src/mods/mod-engine.ts**

- `resolveLocalModSources()` now respect env vars for mods directory.
- `getLegacyExtensionMigrationTarget()` now resolves the migration target root from env vars instead of the
  hardcoded default, so the diagnostic message points to the correct custom directory.

**src/cli/subcommands/mods.ts**

- `listMods()` now respect env vars for mods directory.


src/mods/paths.ts
```diff
@@ -33,6 +61,33 @@ export function resolveDefaultGlobalModsDirectory(
...
+ *   LETTA_EXTENSIONS_DIR env → ~/.letta/extensions/ default.
+ */
+export function resolveGlobalModDirectories(
+  homeDirectory = homedir(),
+  env: NodeJS.ProcessEnv = process.env,
+): ResolvedGlobalModDirectories {
+  const envDirs = getModsDirectoryFromEnv(env);
+  return {
+    globalModsDirectory:
+      envDirs.globalModsDirectory ?? getGlobalModsDirectory(homeDirectory),
```
carenthomas: This keeps an empty string from `getModsDirectoryFromEnv` because it uses `??`. That regresses the existing env parsing where `LETTA_MODS_DIR=`/whitespace is treated as unset: runtime/list now resolve the global mods root to `""` and stop loading the default mods dir. Can we normalize empty strings to undefined, or use the same `||` fallback here?

KY64: Right, let me update.

KY64: Updated https://github.com/letta-ai/letta-code/pull/3334/commits/a408d9ce7c863b33f9645bf872afb3a2e101ba47

PR #3337, opened 2026-07-13
fix(listener): reuse OAuth auth during reconnect

overlord-letta:
## Context
LET-9624: listener reconnects could bypass the OAuth-aware listener auth resolver and read only `LETTA_API_KEY` directly from env/settings. If the saved access token was missing or expired but a refresh token existed, reconnect could fail with the raw `Missing LETTA_API_KEY` path instead of refreshing.

## Root cause
`connectWithRetry()` in `src/websocket/listener/lifecycle.ts` owned the reconnect WebSocket creation but performed its own direct API-key lookup. Initial registration and explicit re-registration used the richer resolver in `listen.tsx`, so reconnect had a different auth path.

## What changed
- Moved listener auth resolution into `src/websocket/listener/auth.ts` so initial registration, re-registration, and reconnect share the same env/settings/OAuth refresh semantics.
- Added non-interactive reconnect auth resolution: reconnect refreshes saved Cloud credentials when possible, never starts device OAuth itself, and routes missing credentials after a prior connection to the existing re-registration callback.
- Added refresh coalescing keyed by refresh token/device/name to avoid duplicate concurrent refresh calls for the same listener credentials.
- Added an explicit `ListenerTokenRefreshError` for non-interactive refresh failures so revoked/invalid refresh credentials surface once instead of silently looping or falling into device OAuth.
- Added a small WebSocket factory seam for lifecycle tests without broad module mocking.

## Changed-behavior matrix
- Env `LETTA_API_KEY`: still wins through the shared resolver.
- Saved Cloud access token: still used when valid.
- Missing saved access token + refresh token on ordinary reconnect: refreshes and reconnects with the new bearer token.
- Expired saved access token + rotated refresh token on reconnect: refreshes, persists the rotated refresh token, and reconnects.
- Missing key + no refresh token after abnormal reconnect: invokes existing re-registration flow instead of throwing raw `Missing LETTA_API_KEY` from lifecycle.
- Invalid/revoked refresh token during reconnect: reports one actionable error and does not start interactive OAuth or retry silently.
- Self-hosted/Desktop proxy startup classification remains in the CLI startup mode path; the shared resolver still only starts OAuth for Cloud URLs.

## Tests
- `bun test src/cli/subcommands/listen-auth.test.ts src/websocket/listener/auth-reconnect.test.ts`
- `bun run check`

## Limits / risk
- Tests use a fake WebSocket factory and OAuth dependency overrides, not a live Cloud WebSocket or real OAuth device flow.
- Invalid refresh credentials are intentionally not auto-recovered by reconnect; the user gets an actionable sign-in/API-key error instead.
- The old helper tests still cover initial CLI auth precedence and self-hosted/Desktop startup mode behavior; the new tests cover production `startListenerClient` close/reconnect flow for normal and abnormal closes.

### DIFF
--- a/scripts/source-file-size-baseline.json
+++ b/scripts/source-file-size-baseline.json
@@ -46,7 +46,7 @@
   "src/websocket/listener/commands/channels.ts": 1390,
   "src/websocket/listener/commands/memory.ts": 1114,
   "src/websocket/listener/file-commands.ts": 1030,
-  "src/websocket/listener/lifecycle.ts": 1766,
+  "src/websocket/listener/lifecycle.ts": 1760,
   "src/websocket/listener/protocol-inbound.ts": 2279,
   "src/websocket/listener/protocol-outbound.ts": 1279
 }
--- a/src/cli/subcommands/listen.tsx
+++ b/src/cli/subcommands/listen.tsx
@@ -10,12 +10,6 @@ import { Box, render, Text } from "ink";
 import TextInput from "ink-text-input";
 import type React from "react";
 import { useState } from "react";
-import {
-  LETTA_CLOUD_API_URL,
-  pollForToken,
-  refreshAccessToken,
-  requestDeviceCode,
-} from "@/auth/oauth";
 import { isLocalBackendEnvEnabled } from "@/backend/local/paths";
 import {
   type ChannelRestoreAgentScope,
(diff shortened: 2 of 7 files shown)


cpacker:
Closing in favor of a smaller replacement. The reconnect auth bug is real, but this implementation does not reproduce the keychain/cache failure, treats every refresh failure as terminal, leaves the in-app listener re-registration path on its captured stale key, and can create a socket after the listener is stopped during an awaited refresh. The local refresh map also does not coordinate with the API client, where the real token-rotation race exists.

I’ll replace this with a cache-first shared resolver, retryable refresh handling, fresh auth in both registration consumers, a post-auth runtime guard, and focused regression tests.

cpacker:
Replacement: #3341

PR #3340, opened 2026-07-13
feat(whatsapp): WhatsApp channel enhancements — LID discovery, config keys, typing, debounce, attachment policy

signo:
## WhatsApp Channel Enhancements

Cumulative WhatsApp channel work, rebased onto `main` (v0.28.2).

### New config keys (all optional, all default to backward-compatible behavior)

1. `message_prefix` — prepend a string to outbound agent replies (e.g. emoji for multi-agent identity). Default: none.
2. `inbound_debounce_ms` — batch rapid-fire messages into one agent turn. Default: `0` (disabled). Range `0..10000`. Voice notes and attachments bypass.
3. `waiting_behavior` — UX feedback during agent turns: `off` (default), `typing_indicator`.
4. `attachment_filter` — when `true`, outbound media checked against MIME/recipient/path allowlists. Default: `false` (disabled, all sends allowed).
5. `attachment_mime_types` — allowed MIME types. `[]` denies all. `["*"]` allows all.
6. `attachment_allowed_recipients` — allowed recipients (phone numbers or JIDs).
7. `attachment_allowed_paths` — allowed source directories (absolute paths, directories only).
8. `attachment_path_recursive` — when `true`, files in subdirectories of allowed paths are also permitted.

### Audio behavior

`.ogg`/`.oga`/`.opus` files are always sent as voice memos (`ptt: true`). All other audio formats (`.mp3`, `.m4a`, `.wav`, etc.) are always sent as documents. This is unconditional and not configurable.

### LID discovery and identity

- **LidDesk** — persistent PN↔LID mapping module. Learns from inbound messages and Baileys signal repository. Survives restarts.
- **On-demand LID lookup** via `sock.onWhatsApp` for phone-only contacts.
- **Outbound route auto-bootstrap** — creates routes for proactive sends to contacts with no prior inbound.
- **Route alias normalization** — phone and LID chatIds treated as aliases. Legacy phone-keyed routes migrated to LID on startup.
- **Allowlist dual-identity** — `allowedUsersIncludes` resolves LID→phone via LidDesk when checking allowlist membership.

### Channel infrastructure

- **Read receipt** — marks inbound messages as read immediately on receipt.
- **Typing indicator** — lifecycle-driven typing loop (12s refresh, 5min max) with pre-reply clear.
- **Inbound debounce** — mirrors Telegram's debounce pattern with bypass for voice notes/attachments.
- **Runtime watchdog** — periodic scan for stuck `isProcessing` states. Env-gated (`LETTA_RUNTIME_WATCHDOG_ENABLED=1`).
- **Disconnect guardrail** — stops reconnecting after 5 rapid disconnects in 60s.
- **Cron routing guidance** — CLI usage text clarifies `--conversation <id>` for context-bound tasks vs `--conversation new` for autonomous.

### `/status` enhancement

- Shows active model for the routed conversation.

## Backward compatibility

All new keys are optional. Missing keys preserve existing behavior:
- `attachment_filter` defaults to `false` (no filtering)
- `inbound_debounce_ms` defaults to `0` (disabled)
- `waiting_behavior` defaults to `off`

Existing audio behavior is preserved: `.ogg`/`.oga`/`.opus` continue to go as voice memos.

## Config conventions

All keys use `snake_case` in storage, `camelCase` at runtime via the existing `SNAKE_TO_CAMEL` map. Wired through types, accounts validator, account-config round-trip, and adapter runtime.

## Baileys v7 readiness

The identity-resolution, message-key access, and sender-id code paths have been refactored to centralize Baileys-version-specific logic behind dedicated modules (`inbound-identity.ts`, `message-key.ts`, `jid.ts`). This means the eventual migration to Baileys v7 should be a small set of file-local changes instead of a scattered refactor across the adapter.

## Test plan

```bash
bun test ./src/channels/whatsapp-media.test.ts
bun test ./src/channels/whatsapp-attachment-policy.test.ts
bun test ./src/channels/whatsapp/jid.test.ts
bun test ./src/channels/whatsapp/lid-desk.test.ts
bun test ./src/channels/whatsapp/inbound-identity.test.ts
bun test ./src/channels/whatsapp/message-key.test.ts
bun test ./src/channels/whatsapp/outbound-bootstrap.test.ts
bun test ./src/channels/whatsapp-typing.test.ts
bun test ./src/channels/whatsapp-adapter.test.ts
bun test ./src/channels/commands.test.ts
bun test ./src/channels/registry-inbound-status.test.ts
```

157 focused tests covering config parsing, attachment policy, LID resolution, allowlist dual-identity, typing, media handling, outbound bootstrap, read receipt, and status output. All pass.

Pre-existing failures (not introduced by this PR): `byok-providers.ts` missing optional dependency, `whatsapp-message-channel.test.ts` mock isolation.

## Known limitations

- `allowed_tools` / `blocked_tools` not implemented (deferred).
- LID-only contacts with no phone mapping anywhere will still fail allowlist checks (graceful degradation — same as current behavior).
- Baileys version remains `6.7.x`; v7 migration is separate future work, but the seams are in place.


overlord-letta:

Thanks for contributing this. There are several useful WhatsApp improvements here, especially outbound route bootstrap, typing lifecycle, debounce, media handling, and the reconnect guardrail. The current PR is too broad to review or merge safely, though: it combines 6k+ lines of WhatsApp work with unrelated runtime-watchdog, cron, status, and local-model changes.

Could you please split this into focused PRs, starting with LID identity/routing? That first PR should:

1. Target the pinned Baileys 6.7.21 API. `signalRepository.lidMapping` is a Baileys v7 surface and does not exist in 6.7.21. For v6, mapping should use the actual message-key fields (`senderPn`, `senderLid`, `participantPn`, `participantLid`) plus `onWhatsApp`.
2. Wire `LidDesk` through the active `registry-inbound.ts` -> `registry-routes.ts` path for allowlist and route-alias checks. The desk-aware implementation currently added to `registry.ts` is not the path used by inbound routing and reintroduces roughly 1,500 lines of duplicated registry logic.
3. Avoid creating a route for an unresolved LID that will later throw when replying. Either resolve it through a proven v6 producer or decline that inbound message.
4. Keep persisted routes stable initially and use alias lookup rather than destructive startup migration. The migration helper currently has no production caller.
5. Use Baileys `readMessages()` for read receipts rather than the more fragile `chatModify({ markRead })` path.
6. Test through the real registry path with captured Baileys v6 message shapes and, if possible, a live WhatsApp smoke test—not only helper-level fixtures.

After that, the remaining work would be much easier to review as separate PRs:

- prefix + debounce + typing, including complete account create/update/save/reload plumbing;
- attachment/audio policy, with `attachment_filter: false` consistently preserved across reloads;
- the reconnect-loop guardrail;
- generic `/status` behavior, if still desired.

Please remove the runtime watchdog, cron guidance, local-provider commits, dead helpers/build-number remnants, and large documentation addition from this branch. The watchdog in particular needs a separate listener-runtime design rather than directly clearing processing state.

This is not a rejection of the WhatsApp work—the smaller pieces are worth upstreaming. Splitting the change and fixing the active LID path will give us something we can validate and merge.

signo:
Closing in favor of a focused 6-PR split for easier review:

1. **#3395** — LID identity and routing for Baileys v6
2. **#3396** — Message prefix, inbound debounce, typing indicator
3. **#3397** — Outbound attachment policy
4. **#3398** — Reconnect guardrail for session-conflict loops
5. **#3399** — Show active model in channel /status output
6. **#3400** — Inbound reactions, SNAKE_TO_CAMEL fixes, read receipt timing, session lock

Total: 35 files, 207 tests. Same scope as this PR but split by concern for manageable review.

PR #3344, opened 2026-07-13
fix(listener): load agent-scoped MemFS mods [LET-9646]

overlord-letta:
## Problem

The websocket listener loaded only global mods. Agent-scoped mods under `$MEMORY_DIR/mods` appeared in discovery but their tools and commands were absent from Slack and other channel turns.

## What changed

- Load a separate MemFS mod adapter per agent after listener memory sync.
- Compose global and agent-scoped tools, permissions, events, and commands into each turn without registering agent capabilities process-wide.
- Refresh the current agent adapter on `/reload` so the next turn gets a fresh tool snapshot.
- Keep transpilation caches outside the git-backed memory directory.
- Preserve the existing global-mod path and source priority behavior.

Agent-scoped provider registration remains disabled in listener mode because provider registration is process-global and cannot be isolated safely.

## Before / after

Before: an agent could install a MemFS mod and `letta mods list --agent` would find it, but listener/channel turns exposed only globally installed mods.

After: each listener turn sees global mods plus only that agent’s MemFS mods. Two agents sharing one listener cannot see or execute each other’s tools or commands.

## Risk

This touches shared mod registration and listener tool-snapshot construction. The isolated adapters retain local definitions without mutating process registries; existing adapters continue using process-global registration by default. Regression coverage exercises two simultaneous agents, reload behavior, command isolation, global mod preservation, and a clean MemFS git status.

## Validation

- Focused mod/listener suite — 73 tests pass
- Full `bun test` attempted — 5,053 pass, 26 skip, 28 fail in unrelated live/environment-sensitive tests (including insufficient Cloud credits); a representative tool-context failure reproduces on unchanged `main`

Linear: [LET-9646](https://linear.app/letta/issue/LET-9646/load-agent-scoped-memfs-mods-in-listener-runtimes)

### DIFF
--- a/src/websocket/listener/mod-adapter.ts
+++ b/src/websocket/listener/mod-adapter.ts
@@ -1,3 +1,4 @@
+import { createHash } from "node:crypto";
 import { join } from "node:path";
 import type Letta from "@letta-ai/letta-client";
 import { getScopedMemoryFilesystemRoot } from "@/agent/memory-filesystem";
@@ -5,6 +6,7 @@ import { getBackend } from "@/backend";
 import { buildModInvocationContext } from "@/mods/context";
 import type { ModEvents } from "@/mods/event-emitter";
 import { createModAdapter, type ModAdapter } from "@/mods/mod-adapter";
+import { getModCacheDirectory } from "@/mods/paths";
 import type {
   ModCapabilities,
   ModContext,
@@ -137,7 +139,15 @@ function resolveListenerAgentModsDirectory(agentId: string): string | null {
   }
 }
 
+function resolveListenerAgentModCacheDirectory(agentId: string): string {
+  // A distinct import path gives each agent its own ESM module instance while
+  // keeping generated files outside the git-backed MemFS repository.
+  const cacheKey = createHash("sha256").update(agentId).digest("hex");
+  return join(getModCacheDirectory(), "listener-agents", cacheKey);
+}
+
 let resolveAgentModsDirectory = resolveListenerAgentModsDirectory;
+let resolveAgentModCacheDirectory = resolveListenerAgentModCacheDirectory;
 
 export async function ensureListenerAgentModAdapter(
   runtime: ListenerRuntime,
@@ -159,6 +169,7 @@ export async function ensureListenerAgentModAdapter(
   const load = (async () => {
     const adapter = createListenerModAdapter({
       agentModsDirectory,
+      cacheDirectory: resolveAgentModCacheDirectory(agentId),
       capabilities: LISTENER_AGENT_MOD_CAPABILITIES,
       includeGlobalMods: false,
       registerCapabilitiesGlobally: false,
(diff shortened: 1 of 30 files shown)


src/websocket/listener/mod-adapter.ts
```diff
@@ -95,14 +126,150 @@ export function ensureListenerModAdapter(runtime: ListenerRuntime): ModAdapter {
...
+  if (existing) return existing;
+
+  const pending = agentModAdapterLoads.get(agentId);
+  if (pending) return pending;
+
+  const agentModsDirectory = resolveAgentModsDirectory(agentId);
+  if (!agentModsDirectory) return null;
+
+  const load = (async () => {
+    const adapter = createListenerModAdapter({
```
carenthomas: This still shares the default mod import cache across all agent adapters. Since the import cache key is basename+source hash and the import URL only adds the source mtime, two agents with the same mod content can hit the same ESM module and share top-level state, so the isolated adapters are not fully isolated. Can we use a per-agent cache dir outside MemFS, or include the agent/source identity in the import URL?

overlord-letta: Fixed in 759a042d: each agent now gets a hashed cache directory under ~/.letta/mod-cache/listener-agents, outside MemFS. The regression loads byte-identical, same-mtime mods with top-level mutable state for two agents and verifies independent activation state plus clean MemFS repos. — Overlord (agent-c2adbf5c-8419-4211-8cd8-3740db164974)

src/websocket/listener/protocol-outbound.ts
```diff
@@ -935,46 +938,39 @@ export function emitQueueUpdateIfOpen(
...
-  Map<string, string>
->();
+export function emitDeviceStatusUpdateIfChanged(
+  socket: ListenerTransport,
+  runtime: RuntimeCarrier,
+  scope: RuntimeScope,
+  options?: { force?: boolean },
+): boolean {
+  const deviceStatus = buildDeviceStatus(runtime, scope);
+  if (!shouldEmitDeviceStatus(socket, scope, deviceStatus, options?.force)) {
```
carenthomas: should this use `getScopeForRuntime(runtime, scope)` before checking the cache too? `emitDeviceStatusUpdate` records with the resolved scope, so keeping this cache key normalized seems safer.

overlord-letta: Overlord (agent-c2adbf5c-8419-4211-8cd8-3740db164974): Yes — the cache key needed the same normalization as the unconditional emitter. `emitDeviceStatusUpdateIfChanged` now resolves through `getScopeForRuntime` before building, comparing, and emitting status. The regression proves `{}` and the explicit agent/conversation scope dedupe on one transport, while a second agent and a second transport remain isolated. Rebased and pushed in c2dc485b; please re-review.

carenthomas (approved):

PR #3347, opened 2026-07-13
fix: chatgpt provider alaises

frezbo:
PR #2971 added support but was not fully wired in, this wires in full support and shows proper model name with the profile alias name too.

Also wire the same for Anthropic so that multiple aliases could be used.


frezbo:
This also adds a `X-Letta-Provider-Alias` header now, so proxy tools can distinguish b/w profiles

frezbo:
Addressed via a custom mod

PR #3357, opened 2026-07-14
fix: enforce send-boundary image normalization [LET-9618]

cpacker:
## Summary

- remove the caller-controlled `skipImageNormalization` escape hatch and make the backend send boundary always normalize the fully composed message array
- canonicalize approval results first, then normalize both top-level user content and nested approval tool-return images before persistence/provider dispatch
- carry channel best-effort image failure policy by message `otid` through queued approval and recovery continuations
- prevent channel and interactive messages from coalescing across different image-failure policies
- centralize queued continuation composition and remove the listener-side normalization helpers/test-only exports
- retain the old mod option as a deprecated, ignored source-compatibility shim

## Root cause

Listener input was normalized before turn composition, then sent with `skipImageNormalization: true`. Messages queued during an active tool/approval turn were appended after that pre-pass, so an oversized image could be persisted unchanged and later rejected by Anthropic's 2,000 px many-image limit.

This change establishes one invariant: every final request is canonicalized and image-normalized immediately before backend persistence/provider dispatch. Source-specific policy can decide whether a failed channel attachment is dropped, but no caller can bypass normalization.

## Review follow-ups

Two independent fresh-context reviews caught and verified fixes for:

- images nested in `ApprovalCreate.approvals[].tool_return`
- mixed channel/interactive queue batches losing failure-policy provenance
- legacy `tool_return: null` payloads requiring canonicalization before image traversal
- typed mod compatibility for the removed skip option

## Verification

- `bun run check` (12/12 checks pass)
- serialized backend request asserts a 3200×1800 user image is clamped
- serialized backend request asserts an oversized, TIFF-labeled approval tool-return image is canonicalized and clamped
- runtime regression asserts even a stale `skipImageNormalization` property cannot bypass the send boundary
- queue regression asserts channel and interactive messages are split at the policy boundary
- legacy approval regression asserts `tool_return: null` is canonicalized before image traversal

Linear: https://linear.app/letta/issue/LET-9618/gpt-fable-size-error


cpacker:
Post-merge fresh-context review found one retry/recovery lifetime issue: channel image failure policy was keyed to OTIDs that post-stop retry replaces. Follow-up fix: #3360.

The follow-up couples current input with its policy, remaps policy atomically across OTID changes/stale-denial rebuilds, and preserves it through queued approval recovery. Independent re-review reported no findings; `bun run check` passes 12/12.

PR #3369, opened 2026-07-15
fix(agent): opt out of assistant message conversion

overlord-letta:

## Summary
- Adds use_assistant_message: false to the single Letta Code conversation request-body builder.
- Covers CLI, Desktop environment listeners, and channel turns that all send through this boundary.
- Adds a focused regression on the generated request body.

## Why
Core still defaults the deprecated assistant-message conversion mode to true. If a modern agent has the legacy send_message tool attached, omitting this field can cause the tool payload to be projected as another assistant response.

## Scope
This is the Letta Code half of LET-9712 PR0. It does not remove the Core/API field or change public SDK defaults. Companion Cloud/UI containment: https://github.com/letta-ai/letta-cloud/pull/13037

## Validation
- bun test src/agent/message-client-skills.test.ts
- bun run check: 12 of 12 checks passed

👾 Generated by Overlord (agent-c2adbf5c-8419-4211-8cd8-3740db164974)

### DIFF
--- a/src/agent/message-client-skills.test.ts
+++ b/src/agent/message-client-skills.test.ts
@@ -33,3 +33,16 @@ describe("buildConversationMessagesCreateRequestBody client_skills", () => {
     ]);
   });
 });
+
+describe("buildConversationMessagesCreateRequestBody use_assistant_message", () => {
+  test("opts out of assistant message conversion", () => {
+    const body = buildConversationMessagesCreateRequestBody(
+      "conversation-1",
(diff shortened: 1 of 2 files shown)


overlord-letta:
Superseded by https://github.com/letta-ai/letta-cloud/pull/13037, which centralizes the default at the Cloud API ingress and covers the known internal direct-Core callers in the owning repository. The separate Letta Code client patch is no longer needed.

PR #3377, opened 2026-07-16
fix(reflection): honor configured model on local backends

oculairmedia:
## Summary
- allow automatic reflection/sleep-time agents to use `LETTA_REFLECTION_MODEL`
- honor explicit reflection model recommendations before local parent-model inheritance
- preserve existing inheritance behavior when no override is configured

## Why
Local-backend subagents currently return the active parent model before resolving reflection-specific configuration. This makes sleep-time agents consume the same model as the primary agent even when operators configure a cheaper memory model.

## Test plan
- [x] `bun test src/agent/subagent-model-resolution.test.ts` (65 tests)
- [x] `bun run typecheck`
- [x] `git diff --check`
- [x] live deployment with `LETTA_REFLECTION_MODEL=minimax-m3` on local App Server and listener services

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -219,6 +219,7 @@ also rejects staged parent-relative imports (`../`); use the `@/` alias.
 |----------|--------|
 | `LETTA_DEBUG=1` | Verbose debug output (default in `bun run dev`) |
 | `LETTA_DEBUG=0` | Suppress debug output even in dev mode |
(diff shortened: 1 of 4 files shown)


oculairmedia:
Validated locally and deployed in production-like local-backend services: 65 focused model-resolution tests pass, TypeScript typecheck passes, and `LETTA_REFLECTION_MODEL=minimax-m3` correctly routes automatic reflection agents without changing default inheritance. Requesting maintainer review/merge; branch permits maintainer edits.

PR #3386, opened 2026-07-17
fix: never omit context_window_limit on model-bearing updates [LET-9786]

cpacker:
## Summary
- **Root cause (LET-9786)**: the server treats an omitted `context_window_limit` on a model-bearing update as "re-derive the llm_config from the handle", which clamps the window to a legacy 128k global default. Our `avoidOverwritingExistingContextWindow` option "preserved" the current window by **omitting** the field — the server then re-clamped it. Worse, a server-clamped 128k matches no registry preset, so the preserve heuristic classified it as a user customization and kept omitting on every later switch: **the corruption was self-perpetuating** (this is exactly how the Overlord agent ran gpt-5.6 at 128k for two months). Reasoning cycles and resume refreshes omitted unconditionally, so they could also poison a healthy >128k agent.
- **New contract**: preserve = re-send the current value explicitly; never omit.
  - `/model` switch (CLI + WS listener, agent & conversation scope): resets the context window to the selected catalog entry's preset, sent explicitly. A custom `/context-limit` is expected to be clobbered by `/model`.
  - Reasoning-effort cycle (Tab), startup resume refresh, and conversation model carryover: preserve the current window (keeps 1M dual-listing variants and `/context-limit` values) by re-sending it explicitly.
  - `updateAgentLLMConfig`/`updateConversationLLMConfig` now always include an explicit `context_window_limit` (from `updateArgs.context_window` when supplied, else derived from the models API); the omission option is deleted.
- **Effect**: poisoned 128k agents self-heal on their next `/model` switch, and no letta-code path can create new poisonings — the server clamp becomes unreachable from this client regardless of when the server-side fix lands.

## Notes
- The `shouldPreserveContextWindowForModelSelection` heuristic is kept, but it now decides *which value to send* (current vs preset), not *whether to send*.
- New `modify-context-window-explicit.test.ts` pins the invariant; `listen-model-update.test.ts` updated to assert the field is never omitted.
- File-size baselines bumped for comment growth (per repo precedent).
- Pre-existing biome error in `prompt-assets.test.ts` (untouched) also fails on clean origin/main.

## Test plan
- [x] `bun test src/agent/ src/websocket/listen-model-update.test.ts` (742 pass)
- [x] `bun run check` (typescript, file-size pass; biome failure pre-exists on main)
- [ ] Live check: `/model` switch on a 128k-poisoned agent repairs the window to the preset


cpacker:
Review addressed in 284caadd — all three findings were legitimate:

**[P1] Local backends discarded the explicit value.** Confirmed: the local-catalog gate (from #2767) forced `updateArgs.context_window` to `undefined`, so preserve paths actively reset custom `/context-limit` values on local backends. Preserve paths (reasoning cycle, resume refresh, carryover, same-variant `/model`) now pass a new `contextWindowOverride` option that wins on **every** backend; catalog presets in `updateArgs` remain local-catalog-gated as before (the pi catalog owns local token limits). Covered by the new `local backends (localModelCatalog)` tests.

**[P1] Uncatalogued handles could still omit.** Confirmed. `resolveContextWindowForUpdate` now falls back, on API backends: override → `updateArgs.context_window` → models API → registry preset → **the entity's current server value re-sent as-is** (last resort — a no-op for the server, but keeps the field present so the clamp path is unreachable). Local backends still omit absent an override, deliberately: no clamp exists there and omission means "keep current". Covered by the empty-models-listing and custom-handle tests.

**[P2] Same-variant changes preserved poisoned 128k.** Confirmed — the heuristic only compared presets, never the current value. New `preservableContextWindow()` in `model.ts` treats a value of exactly 128,000 that matches **no registry preset for the handle** as server-clamp poisoning and returns undefined, so all preserve paths (tier change, reasoning cycle, resume refresh, carryover) fall back to the preset and heal the agent. Models whose presets legitimately include 128k (`gpt-5.3-codex-spark`) are preserved normally — pinned by test.

766 tests pass; `bun run check` fully green (including the previously-failing biome error, which was fixed on main and is resolved here via the rebase-free formatter pass).

cpacker:
Re-review addressed in c47e29e1 — both findings were valid:

**[P1] Local carryover dropped the preserved limit.** Confirmed: carryover put the preserved window in `updateArgs.context_window`, which local-catalog resolution deliberately ignores. `buildConversationModelCarryoverUpdate` now returns the preserved value separately as `contextWindowOverride`, and `AppCoordinator` passes it through — so it survives on all backends. Preset windows stay in `updateArgs` only (local backends own those via the pi catalog). Pinned by three new carryover tests, including one asserting a poisoned 128k active window is NOT carried into new conversations.

**[P1] Inheriting conversations could still omit.** Confirmed: a conversation without its own override has `context_window_limit: null` and inherits from its agent, so the last-resort fallback returned undefined. The conversation fallback now walks up to `conversation.agent_id` → agent `llm_config.context_window` when no conversation-level value exists. Pinned by the new agent-walk test.

Also merged `origin/main` into the branch (conflicts were the size baseline only; `modify.ts`/`headless.ts` auto-merged) — the PR should no longer show conflicting. 780 tests pass, `bun run check` green.

PR #3390, opened 2026-07-17
feat(models): refresh curated model catalog from the cloud catalog endpoint [LET-9792]

cpacker:
## Summary

Makes the cloud canon for curated model data on the cloud backend, so model additions/metadata fixes no longer require a CLI release.

- New `src/agent/remote-model-catalog.ts`: fetches `GET /v1/models/catalog`, maps entries back into the bundled models.json shape (typed `contextWindow`/`maxOutputTokens` + `config` recombine into `updateArgs`), and swaps the live `models` catalog contents **in place** so every existing consumer — selector, subagent model resolution, agent-create defaults, carryover, listener `list_models` — picks up live data with zero call-site changes.
- Bundled models.json remains the seed and the fallback (this also covers LET-9793): non-API backends (self-hosted, local/pi-ai) never fetch; HTTP/network/parse failures and payloads failing a sanity gate (empty, or no default entry) leave the current catalog untouched.
- Successful refreshes persist to `~/.letta/cache/model-catalog.json` and reload at next startup, so offline sessions keep the freshest known data instead of regressing to the release snapshot.
- Refresh triggers ride the existing availability fetches: startup prefetch, ModelSelector load/refresh (force on explicit refresh), listener `list_models` (force on user-driven refresh). TTL-throttled (5 min) + inflight-deduped.

## Verification

- 12 new unit tests (mapping, in-place swap + sanity gate, every failure mode, TTL throttle, disk persist/reload)
- E2E against a local cloud-api running the paired branch: catalog refreshed 209/209 entries, correct featured/default flags, preset down-pins intact (Opus 200k over 950k ground truth), and the live models.json typo (Sonnet 4.6 1M at 9.5M) corrected by server data at runtime
- `bun run check` all 12 green; model/listener/component suites pass
- File-size baseline bumps for AppCoordinator (+4) and ModelSelector (+9) per established practice

## Test plan

- [ ] After letta-cloud #13101 merges + cloud-api deploy: run this branch against prod, open /model, verify list matches DB contents
- [ ] Verify offline start (no network) still shows full catalog from cache/snapshot


mdsubojsarker5-lgtm (dismissed):

cpacker:
Re-review found three correctness issues, all fixed in `ae22cbb1`:

1. **Malformed responses could partially replace the catalog.** The old validator checked only `id` / `handle` / `label`, dropped invalid rows, and applied the remainder. A mixed or schema-drifted response could therefore erase most presets while reporting success. Remote rows now validate all required and optional mapped fields, and any invalid row rejects the entire response. The final mapped catalog also rejects empty data, duplicate IDs, invalid cached rows, and ambiguous/missing defaults.

2. **The live catalog, TTL, in-flight request, and disk cache were global across API servers.** Switching from Cloud to a custom/self-hosted URL could leave the Cloud catalog active (or reload it from `~/.letta/cache`) after the new server returned 404. Cache records are now schema-versioned and scoped to the normalized base URL plus the bundled-catalog fingerprint. Source changes restore the bundled snapshot and reset refresh state; generation guards prevent a late response from the previous server from overwriting the active catalog.

3. **Catalog refresh had no deadline.** Both `/model` and listener `list_models` await it, so an endpoint that accepted the connection but never answered could hang those commands indefinitely. Requests now carry a 5-second abort signal.

Regression coverage is 17/17 for the remote catalog, including cross-server disk cache rejection and a late-response race. The 91 focused catalog/model-selector/listener tests pass, `bun run typecheck` is clean, and all 12 `bun run check` gates pass.

One scope/documentation note remains: self-hosted API URLs do attempt the endpoint and safely fall back; they are not a strict no-fetch path as the PR body currently says. Also, the `App` startup prefetch happens after initial startup agent resolution, so a first-run agent creation is not guaranteed to see remote metadata. I left creation non-blocking rather than adding up to five seconds to startup; the PR body should narrow the `agent-create defaults` claim, or that ordering can be handled separately if it is a requirement.

PR #3395, opened 2026-07-17
fix(whatsapp): canonicalize Baileys LID identities

signo:
## Summary

This rebuild canonicalizes Baileys LID identities before WhatsApp messages reach routing, pairing, or access control.

- Direct chats use the strict phone JID as `chatId` and phone digits as `senderId`.
- Groups retain the group JID as `chatId` and use canonical phone digits for the participant `senderId`.
- Unresolved, malformed, or conflicting LID evidence fails closed.
- Validated observations persist in a single-direction LID → phone-JID store.
- Existing LID-form outbound targets resolve through that store; unresolved targets still fail.
- Read receipts are bulked per accepted upsert batch, best-effort, non-blocking, and bound to the socket that emitted the batch.

## Design

`resolveInboundIdentity()` is pure with respect to persistence. It collects role-consistent Baileys identity fields (`remoteJid`, `senderLid`, `senderPn`, `participant`, `participantLid`, and `participantPn`), normalizes and deduplicates them, rejects conflicting phone candidates or stored mappings, and returns canonical identity plus unmapped observations.

The adapter applies those observations and batch-flushes them through an atomic temp-file-and-rename store. The store persists only LID → phone JID; there is no reverse map or generic alias layer.

## Cameron's review findings addressed

1. **Shared-suite mock isolation** — new tests use dependency injection; no process-global `mock.module()` replacements were added. The complete WhatsApp suite passes together.
2. **Route/pairing identity convergence** — canonical ingress prevents raw LIDs from becoming route, pairing, dedupe, or authorization identities. Real-adapter + `ChannelRegistry` tests prove LID and PN forms use one route/conversation and one pairing/access identity.
3. **Dead proactive bootstrap** — `tryBootstrapOutboundRoute()` and route-less proactive outbound are not part of this rebuild.
4. **Blocking read receipts** — delivery never awaits `readMessages()`; one bulk call is made after each accepted batch, with synchronous and asynchronous failures contained.
5. **Group participant canonicalization** — participant LID/PN fields resolve to canonical phone digits, including persisted LID-only follow-up messages.
6. **Stale reverse mappings** — the persisted reverse map was removed. Existing LID mappings never silently rebind; conflicts preserve the stored value and drop the ambiguous message.

## Deliberately out of scope

- Generic routing, access-control, pairing, or `MessageChannel` changes
- `onWhatsApp` bootstrap or `signalRepository.lidMapping` fallback
- Generic alias migration
- Proactive route-less WhatsApp outbound

## Validation

Rebased onto upstream `0.29.9` (`45d76e7d`):

- Complete WhatsApp suite: **107 passed**
- Focused adapter/registry/pairing/access suite: **50 passed**
- `bun run check`: **12/12 passed**
- Pre-commit checks: passed

No production deployment or live-traffic claim is included in this PR.


just-cameron:
Reviewed at `3613648544d9d429409dd6a6ed70404b1cd8625b`. I think this needs changes before merge.

### Blockers / major issues

1. **The new module mocks break the shared WhatsApp test suite.** Both `whatsapp-adapter.test.ts` and `outbound-bootstrap.test.ts` install process-global `mock.module()` replacements without restoring them, and the mocked session module omits existing exports. I reproduced this with `bun test whatsapp`: 157 tests pass, then `whatsapp-session.test.ts` fails to import `acquireWhatsAppSessionLease`. The PR's 132 new tests only pass when run in isolation. CI's test-mock-isolation failure is valid here.

2. **PN/LID aliases are not integrated with route lookup.** `resolveWhatsAppAlias()` says it is used by route lookup, but it is only referenced by tests. If an initial message creates a route under an unresolved LID and a later message resolves that contact to a PN via `senderPn`, the registry performs an exact lookup under the new PN and creates a second route/conversation. Pairing identity can shift similarly. Route lookup/migration needs to understand aliases, or the adapter needs one stable canonical chat ID.

3. **`tryBootstrapOutboundRoute()` has no production caller.** It is not on the base `ChannelAdapter`, `MessageChannel` requires an existing exact route before obtaining the adapter, WhatsApp has no explicit-target resolver, and this method explicitly does not create a route. The advertised proactive outbound-routing flow is therefore exercised only by the direct adapter unit test, not by the application.

4. **Read receipts can block inbound delivery indefinitely.** The adapter awaits `readMessages()` before `adapter.onMessage`. In Baileys 6.7.21 this performs network work (privacy lookup plus receipt send), so a slow or never-settling receipt blocks the message and all later messages in the upsert batch. It also makes one sequential call per message even though the API accepts bulk keys. Delivery should not wait on best-effort receipts.

5. **Group participant identity is still not canonicalized.** For a group key with `participant: "...@lid"` and `participantPn: "...@s.whatsapp.net"`, the desk mines the mapping, but `senderJid` still chooses `participant` first and `senderIdFromJid()` emits the numeric LID. `participantPn` and the desk mapping are ignored. The new `normalizeMessageKey()` accessor that exposes these fields is itself never used by production code.

6. **`LidDesk.record()` does not preserve a one-to-one mapping.** Remapping one LID from PN A to PN B updates the forward entry but leaves `jidToLid[A]` behind, so both PNs resolve to the same LID. Persisted alias/presence/bootstrap lookups can therefore return stale identities. Loading also trusts the two maps independently rather than reconciling them.

### Current checks

`Lint & Type Check` is red for PR-owned issues: the unsafe mocks above, an unused `messagesUpsertHandler`, adapter formatting, and the resulting TypeScript failure. Downstream test/integration jobs were skipped. The separate `review` job failed because its workflow did not set `GH_TOKEN`; that one appears infrastructure-related rather than a problem with this patch.

signo:
@just-cameron Thanks for the detailed review. I rebuilt the branch from current upstream rather than layering fixes onto the rejected implementation. Current head: `2c801028`.

All six blockers now have direct closure evidence:

1. **Mock isolation:** no new process-global module mocks; the complete WhatsApp suite passes together.
2. **Route/pairing aliases:** raw LIDs are canonicalized before registry ingress. Real-adapter + `ChannelRegistry` regressions prove LID and PN forms share one route/conversation and one pairing/access identity.
3. **Dead proactive bootstrap:** removed from scope; no route-less proactive outbound or generic route mutation.
4. **Read receipts:** one bulk, fire-and-forget `readMessages()` call per accepted batch; sync/async failures are contained and receipts stay bound to the emitting socket across reconnects.
5. **Group sender identity:** `participant`, `participantLid`, `participantPn`, `senderLid`, and `senderPn` are resolved through one strict canonical identity path.
6. **Mapping consistency:** persistence is single-direction LID → phone JID. Conflicts never rebind, and ambiguous persisted entries are omitted; no stale reverse map exists.

The branch is rebased onto upstream `0.29.9` (`45d76e7d`). Validation: WhatsApp **107 passed**, focused adapter/registry/pairing/access **50 passed**, and `bun run check` **12/12 passed**.

The PR body now states the narrowed design and exclusions explicitly. Please take another look when convenient.

signo:
Combined downstream validation update (no change to this PR's head):

Canonical identity checkpoint `2c801028` was used as the base for six independently rebuilt WhatsApp slices and combined locally at `d7c60a59`. The exact source-built image passed 193/193 WhatsApp tests, 12/12 repository checks, and a live Arrakis gate covering:

- direct LID → canonical phone route convergence
- two-message debounce and delayed receipt ownership
- lifecycle typing and exact outbound prefix
- outbound reaction and structured inbound LID reaction
- allowed and denied outbound attachment policy

The final retained runtime is healthy with zero restarts/OOM after infrastructure correction. This is combined-stack evidence, not a claim that downstream features belong in #3395. No downstream branch was pushed, and #3395 remains the required merge base before any replacement PR is published.

PR #3399, opened 2026-07-17
feat(channels): show active model in channel /status output

signo:
> [!NOTE]
> **DISCARDED / DEFERRED — THIS PR WILL NOT BE PURSUED IN THE CURRENT WHATSAPP SERIES.**
>
> Active-model status output is unrelated to the canonical identity → debounce/typing/prefix → attachment → reconnect → reaction dependency chain. Keeping it in that chain increases review and integration risk without contributing to the validated WhatsApp behavior. The idea may be reconsidered later as an independent generic channel-status change.

## Historical description

Displays the active model in the `/status` command output for WhatsApp and other channels using the runtime model-status source.


just-cameron:
Maintainer repair pushed: `204797322543bb27522c2c2f658e1469ce44f6c1`.

Summary:
- Add a status-context gate so pending-pairing senders keep read-only `/status` access without runtime model enrichment.
- Avoid invoking the model-status handler while `senderAccess === "pair"`.
- Add regressions for pending-pairing `/status`, paired `/status`, and explicit disabled enrichment.

Validation:
- `bun test ./src/channels/registry-inbound-status.test.ts ./src/channels/registry-status-access.test.ts ./src/channels/registry-command-routing.test.ts ./src/channels/commands.test.ts ./src/websocket/listen-model-update.test.ts`
- `bun test ./src/channels/discord-service.test.ts`
- `bun run check`

Note: a parallel local run of `bun test ./src/channels` across both PR worktrees timed out in unrelated Discord service tests; rerunning that file standalone passed.

signo:
Closing as discarded/deferred from the current WhatsApp series. If revived, this should return as an independent generic channel-status change rather than a dependency in this stack.

PR #3407, opened 2026-07-18
feat(cron): route letta cron through durable Cloud schedules by default [LET-9692] [LET-9820] [LET-9821]

cpacker:
## Summary

- `letta cron` previously always wrote runtime-local tasks to `~/.letta/crons.json`, which silently die with the device/sandbox that created them (LET-9692). Cloud agents now default to durable Cloud schedules (`/v1/agents/:id/schedule`), which fire from the cloud and execute inside the agent's managed cloud sandbox — surviving local shutdown.
- New `--runner local|cloud` flag on `add`/`list`/`get`/`runs`/`delete`. The default is `cloud` for cloud agents everywhere (laptop, VPS, managed sandbox); `local` remains the default for local-backend agents (`agent-local-*`) and servers without the Cloud schedule routes, and is the explicit opt-in for schedules that must run on a specific machine.
- New `--target-device <deviceId>` on `add` (cloud runner only, LET-9821): keeps the durable cloud-owned timer but executes on a named registered device (from `letta environments list`), with automatic fallback to the agent's sandbox if the device is offline at fire time. Backend support landed in the receiving service and is deployed. With `--runner local` the flag errors — a local task already runs on the device that owns it.
- Cloud support is detected by probing the schedule route rather than matching the base URL: managed sandboxes and Desktop sessions point `LETTA_BASE_URL` at a localhost proxy that forwards to the Letta API, so URL shape says nothing about capability.
- `list`/`get`/`runs`/`delete` operate across both runners and tag every task with a `runner` field (plus `target_device_id`/`execution_target` for device-targeted schedules); cloud `runs` map to the schedule history endpoint. `delete --all` reports per-runner counts.
- Failure to create a Cloud schedule is a hard error with no silent fallback to ephemeral local storage (the exact failure mode this ticket fixes).
- Recurring cloud cron expressions surface a note that they are currently interpreted in UTC until the schedule contract gains an IANA timezone field (LET-9815).
- Agent guidance updated (LET-9820): the scheduling-tasks skill gains a "Where Schedules Run" section, and device-pinned work is steered toward `--target-device` Cloud schedules, reserving `--runner local` for when the sandbox fallback is unacceptable. The system prompt's cron section matches. Local-runner-specific notes (local timezone, session-must-be-running, 24h GC) are labeled as such.

## Verification

- `bun run check` (all 12 pass); 18 unit tests for runner resolution + payload mapping (`cron-runner.test.ts`) incl. device-target mapping/notes/whitespace handling; existing cron suite and prompt-asset tests pass.
- Live smoke tests against the Letta API through a Desktop localhost proxy:
  - default `add` created a Cloud schedule (`runner: "cloud"`, sandbox target), `list` merged local + cloud tasks, `get`/`runs`/`delete` round-tripped, `--runner local` still writes device-local tasks, `--runner cloud` errors for `agent-local-*` agents, deleted IDs return not-found.
  - `--target-device` with a registered org device created a device-targeted Cloud schedule (`execution_target` = deviceId, fallback note in output), `get` echoed the target, delete cleaned up; an unregistered deviceId surfaced the server's 404 with no schedule created; `--target-device` + `--runner local` errors.

### DIFF
--- a/src/backend/api/schedules.ts
+++ b/src/backend/api/schedules.ts
@@ -28,12 +28,19 @@ export interface CreateCloudScheduleInput {
   conversation_id?: string;
   messages: CloudScheduleMessage[];
   schedule: CloudScheduleSpec;
+  /**
+   * Optional execution target: a registered environment deviceId (e.g. a
+   * Railway/VPS listener). When the device is offline at fire time, the
+   * cloud worker falls back to the agent's managed sandbox.
+   */
+  target_device_id?: string;
 }
 
 export interface CreateCloudScheduleResponse {
   id: string;
   next_scheduled_at?: string;
   use_sandbox: boolean;
+  target_device_id?: string | null;
 }
 
 export interface CloudSchedule {
@@ -46,6 +53,7 @@ export interface CloudSchedule {
   schedule: CloudScheduleSpec;
   next_scheduled_time: string | null;
   use_sandbox: boolean;
+  target_device_id?: string | null;
   created_at?: string;
 }
 
--- a/src/backend/api/schedules.ts
+++ b/src/backend/api/schedules.ts
@@ -1,0 +1,149 @@
+/**
+ * Cloud schedules client (`/v1/agents/:agent_id/schedule` endpoints).
+ *
+ * These routes are served by the Letta API (cloud-api), not the OSS core
+ * server. Cloud schedules are durable: the schedule definition lives in the
+ * Letta API's database and fires from a cloud worker, which injects the
+ * scheduled turn into the agent's managed cloud sandbox (`use_sandbox: true`).
+ * They survive local process/device/sandbox termination, unlike runtime-local
+ * crons in ~/.letta/crons.json.
+ */
+
+import { apiRequest } from "./request";
+
+// ── Wire types (mirror the cloud scheduledMessages contract) ────────
+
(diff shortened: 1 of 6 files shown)


4shub (approved):

src/backend/api/schedules.ts
```diff
@@ -0,0 +1,157 @@
...
+): Promise<CreateCloudScheduleResponse> {
+  return apiRequest<CreateCloudScheduleResponse>(
+    "POST",
+    schedulePath(agentId),
+    {
+      ...input,
+      // The only harness-capable execution target the cloud worker supports
+      // today is the agent's own managed sandbox. Named remote targets are
+      // future work (LET-9821).
+      use_sandbox: true,
```
amelia-letta, line 101: This comment is stale — "Named remote targets are future work (LET-9821)" is exactly what this PR implements. The `target_device_id` field now flows through `...input`, so `use_sandbox: true` is the fallback, not the only target.

cpacker: Fixed in f7741895 — reworded to describe the current behavior (sandbox is the fallback; `target_device_id` names a device target, and `use_sandbox: true` is required by the server for device targets either way), and refreshed the module header which had the same stale sandbox-only wording.

PR #3425, opened 2026-07-20
fix(auth): handle transient OAuth responses

overlord-letta:
## Problem

The `/login` device flow assumed every OAuth response was JSON. When an edge or upstream service returned an HTML error page, the CLI exposed the parser exception (`Unexpected token '<'`) instead of the HTTP failure.

## What changed

- Parse device-code and token-poll responses defensively without surfacing response bodies.
- Include only safe diagnostics: auth host, HTTP status, media type, and a sanitized request ID when available.
- Retry device-code creation once and tolerate two consecutive transient token-poll responses before showing the actionable error.
- Preserve `authorization_pending`, `slow_down`, denial, expiry, timeout, transport-error, and cancellation behavior.

## Scope and risk

This is client defense-in-depth for the login device flow; it does not claim to identify or fix the transient upstream HTML producer. Refresh and revoke paths are unchanged. A retried device-code request can leave one unused short-lived code if the first request succeeded upstream but its response was corrupted.

## Validation

- `bun test src/auth/oauth-network-errors.test.ts` — 12 passed
- `bun run check` — 12/12 checks passed
- The five new regression cases fail against `origin/main` and pass on this branch.
- Real TUI fault injection: two HTML `502 text/html` device-code responses triggered one bounded retry, then rendered the safe HTTP/media/request-ID error; the response-body sentinel and raw parser text were absent.
- Full `bun test` — 5,267 passed, 26 skipped; 12 unrelated environment/existing failures. Eleven reproduced on the base commit, and the remaining API parity probe passed on rerun.

Linear: [LET-9888](https://linear.app/letta/issue/LET-9888/tui-login-exposes-raw-html-json-parse-errors-instead-of-an-actionable)

### DIFF
--- a/src/auth/oauth-network-errors.test.ts
+++ b/src/auth/oauth-network-errors.test.ts
@@ -14,6 +14,35 @@ function makeFetchFailure(message: string, code?: string): Error {
   return new TypeError("fetch failed", { cause });
 }
 
+function makeHtmlGatewayError(headers?: Record<string, string>): Response {
+  return new Response(
+    "<!DOCTYPE html><html><title>Bad Gateway body marker</title></html>",
+    {
+      status: 502,
+      headers: {
+        "Content-Type": "text/html; charset=utf-8",
+        ...headers,
+      },
+    },
+  );
+}
+
+function makeJsonTransientError(headers?: Record<string, string>): Response {
+  return new Response(
+    JSON.stringify({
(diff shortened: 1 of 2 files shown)


just-cameron:
@cpacker quick review request -- this adds some fault-tolerance for the `/login` command to improve the user experience when the serve doesn't provide correct JSON.

PR #3427, opened 2026-07-20
feat: OpenAI-compatible API for App Server (--openai-api)

sarahwooders:
## Summary

Adds an OpenAI-compatible HTTP API to the App Server, enabled with a new flag:

```bash
letta server --backend local --listen ws://0.0.0.0:4500 --openai-api
```

This lets any OpenAI-compatible frontend or SDK — Open WebUI, LibreChat, LobeChat, the OpenAI Python/JS SDKs — talk to a self-hosted Letta server by pointing the base URL at `http://host:4500/v1`.

### Endpoints

- **`GET /v1/models`** — lists agents as models. Unique agent names are advertised as model ids (so pickers show `memo`, `Tutor`, …); duplicate names fall back to the agent id so every listed id resolves unambiguously.
- **`POST /v1/chat/completions`** — routes the **last user message** into a Letta conversation on the resolved agent. The client-resent history is not replayed into the agent (server-side memory and state are authoritative), but it is fingerprinted to keep each client-side chat pinned to its own Letta conversation: a first message starts a fresh conversation, and after each turn the transcript-plus-reply is mapped to that conversation — exactly the prefix the client resends next. Supports `"stream": true` with standard `chat.completion.chunk` SSE events, non-streaming aggregation with usage stats, and `image_url` content parts (data: URLs → base64 image sources, http(s) → url sources) for vision input.

### Architecture: a seam over the v2 runtime

The bridge is the HTTP analogue of a messaging channel — it converts chat-completions inputs into v2 runtime turns and v2 protocol events back into chat-completions outputs:

- **Inbound**: requests dispatch through the same path as WebSocket clients (`dispatchInboundMessageWhenReady` → queue pump → turn processor), so **tools, mods, skills, permissions, and per-conversation queueing all apply**. Tool calls execute inside the harness; verified live with real shell-tool round-trips.
- **Outbound**: a new in-process **stream-observer seam** (`stream-observers.ts`): `emitProtocolV2Message` notifies registered observers with each outbound protocol message and its resolved runtime scope, independent of socket routing. The bridge subscribes per-turn and translates `stream_delta` / `update_loop_status` events into SSE chunks. This seam is reusable by any future in-process protocol consumer.
- **Runtime lifecycle**: with no WS control session, the bridge starts a socket-free runtime via `startLocalChannelListener` (same as `letta server --channels`). When a control client is attached, its runtime is reused — and the client sees bridge-initiated turns on its stream channel.
- **Approvals**: bridge conversations run with permission mode `unrestricted` — there is no interactive approver on this surface, and the API is opt-in and bearer-gated (full toolset behind the API key). `requires_approval` stream segments are non-terminal (the turn processor evaluates and continues); if a turn does end up `WAITING_ON_APPROVAL`, the request completes with an explanatory reply instead of hanging.
- **Auth**: `/v1` routes reuse the existing WebSocket bearer policy (capability-token / signed-bearer-token) via `authorizeUpgrade`. Off by default; without `--openai-api` the routes 404.

## Testing

- 12 tests across `app-server-openai.test.ts` (wire format: flag off ⇒ 404, model listing, auth accept/reject, non-streaming aggregation, conversation stickiness/isolation, SSE shape, `model_not_found`, missing user text) and `app-server-openai-sdk.test.ts` (conformance through the real `openai` client: models.list, non-streaming + streaming create, 401 as APIError).
- Verified live from source against the local backend with real inference:
  - **shell tool execution through the API returns real command output** (`echo harness-test-$((6*7))` → `harness-test-42`)
  - streaming SSE, conversation continuity (code-phrase recall) and isolation across chats
  - image input: data-URL PNG/JPEG through the API (correct vision answers), including a 10 MB photo over streaming
  - concurrent new chats: three simultaneous first messages all complete with no conversation-creation races
  - wrong bearer token ⇒ 401
  - **Open WebUI (current release) end-to-end, user-confirmed**: agents in the model picker, streamed chats, tool-using turns, image uploads (vision replies), per-sidebar-chat Letta conversations, persistence across refresh
- `bun run check`: all checks pass except pre-existing biome errors on `main` in two untouched files; the websocket suite's `remote-settings.test.ts` failure is pre-existing on `main` (real-HOME settings cache pollution across test files) and passes in isolation.

## Follow-ups (out of scope)

- `/v1/responses` (OpenAI's stateful API) maps more faithfully onto Letta conversations (`previous_response_id` ↔ conversation) than fingerprinting.
- Optional custom SSE events for tool progress (e.g. `letta.tool.progress`) so clients can render live tool activity — the observer seam already surfaces the events; only the SSE translation is missing.
- Interactive approval bridging (render approval requests as assistant text and parse replies, like the channels' interactive flow) if `unrestricted` is too permissive for some deployments.
- Heuristic for frontend task requests (Open WebUI title generation creates one-off conversations; configuring a separate task model avoids it).
- Docs: the self-hosting page in letta-docs should gain an Open WebUI section once this lands.

https://claude.ai/code/session_01GBzTAACqizoyQcAZyrPoz8

### DIFF
--- a/src/websocket/app-server-openai-turn.ts
+++ b/src/websocket/app-server-openai-turn.ts
@@ -2,9 +2,15 @@ import { randomUUID } from "node:crypto";
 import { settingsManager } from "@/settings-manager";
 import { getOrCreateScopedRuntime } from "@/websocket/listener/conversation-runtime";
 import { dispatchInboundMessageWhenReady } from "@/websocket/listener/inbound-dispatch";
-import { startLocalChannelListener } from "@/websocket/listener/lifecycle";
+import {
+  startLocalChannelListener,
+  stopRuntime,
+} from "@/websocket/listener/lifecycle";
 import { getOrCreateConversationPermissionModeStateRef } from "@/websocket/listener/permission-mode";
-import { getActiveRuntime } from "@/websocket/listener/runtime";
+import {
+  getActiveRuntime,
+  setActiveRuntime,
+} from "@/websocket/listener/runtime";
 import {
   type ListenerTransport,
   LocalListenerTransport,
@@ -88,6 +94,23 @@ const OPENAI_TURN_TIMEOUT_MS = 15 * 60 * 1000;
 
 const bridgeTransport = new LocalListenerTransport();
 let bridgeRuntimeStart: Promise<void> | null = null;
+let bridgeOwnedRuntime: ListenerRuntime | null = null;
+
+/**
+ * Stop the listener runtime the bridge started, if it still owns it. Called
+ * from AppServerHandle.close(): a WS-session runtime is closed by its own
+ * socket lifecycle, but a bridge-started (socket-free) runtime has no other
+ * owner and would otherwise outlive the server, keeping cron and channel
+ * lifecycles running.
+ */
+export function closeOpenAiBridgeRuntime(): void {
+  const runtime = bridgeOwnedRuntime;
+  bridgeOwnedRuntime = null;
+  if (!runtime || runtime.intentionallyClosed) return;
+  if (getActiveRuntime() !== runtime) return;
+  stopRuntime(runtime, true);
+  setActiveRuntime(null);
+}
 
 /**
  * Reuse the active listener runtime when one exists (a connected WS control
@@ -119,6 +142,7 @@ async function ensureListenerRuntime(
   if (!runtime || runtime.intentionallyClosed) {
     throw new Error("failed to start listener runtime for OpenAI-compat API");
   }
+  bridgeOwnedRuntime = runtime;
   return runtime;
 }
 
@@ -314,6 +338,10 @@ async function runTurnViaListenerRuntime(
       type: "message",
       agentId: params.agentId,
       conversationId: params.conversationId,
+      // Each HTTP request settles from its own turn: batching two requests
+      // would strand the second observer (only the first OTID survives a
+      // merged batch) and answer the first with both prompts.
+      noCoalesce: true,
       messages: params.messages,
     };
     try {
--- a/src/websocket/app-server-openai-turn.ts
+++ b/src/websocket/app-server-openai-turn.ts
@@ -1,0 +1,340 @@
+import { randomUUID } from "node:crypto";
+import { settingsManager } from "@/settings-manager";
+import { getOrCreateScopedRuntime } from "@/websocket/listener/conversation-runtime";
+import { dispatchInboundMessageWhenReady } from "@/websocket/listener/inbound-dispatch";
+import { startLocalChannelListener } from "@/websocket/listener/lifecycle";
+import { getOrCreateConversationPermissionModeStateRef } from "@/websocket/listener/permission-mode";
+import { getActiveRuntime } from "@/websocket/listener/runtime";
+import {
+  type ListenerTransport,
+  LocalListenerTransport,
+} from "@/websocket/listener/transport";
+import { handleIncomingMessage } from "@/websocket/listener/turn";
+import { registerTurnObserver } from "@/websocket/listener/turn-observers";
+import type {
+  IncomingMessage as ListenerIncomingMessage,
+  ListenerRuntime,
+  ListenerStreamObserver,
+  ProcessQueuedTurn,
+  StartListenerOptions,
+} from "@/websocket/listener/types";
+
+export interface OpenAiUsage {
+  prompt_tokens: number;
+  completion_tokens: number;
+  total_tokens: number;
+}
+
+export type UserContentPart =
+  | { type: "text"; text: string }
+  | {
+      type: "image";
+      source:
+        | { type: "base64"; media_type: string; data: string }
+        | { type: "url"; url: string };
+    };
+
+export interface TurnOutcome {
+  text: string;
+  usage: OpenAiUsage;
+  error: string | null;
+}
+
+export interface BridgeTurnMessage {
+  role: "user" | "assistant";
+  content: UserContentPart[];
+  otid: string;
+}
+
+interface RunTurnParams {
+  agentId: string;
+  conversationId: string;
+  /** Full input for the turn: the newest message in stateful mode, or the
+   * replayed client transcript in stateless mode. */
+  messages: BridgeTurnMessage[];
+  /** OTID of a message in this turn, used to correlate the listener turn
+   * lifecycle with this request. */
+  correlationOtid: string;
+  onAssistantText?: (text: string) => void;
+  onLog?: (message: string) => void;
+}
+
+type RunTurnImpl = (params: RunTurnParams) => Promise<TurnOutcome>;
+
+let runTurnImpl: RunTurnImpl = runTurnViaListenerRuntime;
+
+/** @internal Test seam mirroring __testSetBackend. */
+export function __testSetRunTurnImpl(impl: RunTurnImpl | null): void {
+  runTurnImpl = impl ?? runTurnViaListenerRuntime;
+}
+
+/** Stable entry point used by the HTTP handler; tests swap the impl. */
+export function runBridgeTurn(params: RunTurnParams): Promise<TurnOutcome> {
+  return runTurnImpl(params);
+}
+
+// ---------------------------------------------------------------------------
+// Turn execution over the listener v2 runtime.
+//
+// The bridge runs turns through the same dispatch path as WebSocket clients
+// (queueing → turn processor → tools/mods/permissions) and observes the
+// resulting v2 protocol stream via the runtime's in-process stream
+// observers, converting chat-completions requests into protocol inputs and
+// protocol events back into chat-completions outputs — the HTTP analogue of
+// a messaging channel.
+// ---------------------------------------------------------------------------
+
+const OPENAI_TURN_TIMEOUT_MS = 15 * 60 * 1000;
+
+const bridgeTransport = new LocalListenerTransport();
+let bridgeRuntimeStart: Promise<void> | null = null;
+
+/**
+ * Reuse the active listener runtime when one exists (a connected WS control
+ * session or channels runtime); otherwise start a socket-free local runtime,
+ * exactly like `letta server --channels` does. If a WS control client
+ * connects later it replaces the bridge-owned runtime, and subsequent
+ * requests transparently use the new active runtime.
+ */
+async function ensureListenerRuntime(
+  onLog?: (message: string) => void,
+): Promise<ListenerRuntime> {
+  const active = getActiveRuntime();
+  if (active && !active.intentionallyClosed) return active;
+
+  bridgeRuntimeStart ??= startLocalChannelListener({
+    connectionId: `openai-api-${randomUUID()}`,
+    deviceId: settingsManager.getOrCreateDeviceId(),
+    connectionName: "openai-api",
+    onConnected: () => {},
+    onError: (error) => {
+      onLog?.(`OpenAI-compat runtime error: ${error.message}`);
+    },
+  }).finally(() => {
+    bridgeRuntimeStart = null;
+  });
+  await bridgeRuntimeStart;
+
+  const runtime = getActiveRuntime();
+  if (!runtime || runtime.intentionallyClosed) {
+    throw new Error("failed to start listener runtime for OpenAI-compat API");
+  }
+  return runtime;
+}
+
+function extractDeltaText(delta: unknown): string {
+  const content = (
+    delta as { content?: string | Array<{ text?: string }> | null }
+  ).content;
+  if (typeof content === "string") return content;
+  if (!Array.isArray(content)) return "";
+  return content
+    .map((part) =>
+      part && typeof part === "object" && typeof part.text === "string"
+        ? part.text
+        : "",
+    )
+    .join("");
+}
+
+async function runTurnViaListenerRuntime(
+  params: RunTurnParams,
+): Promise<TurnOutcome> {
+  const listener = await ensureListenerRuntime(params.onLog);
+  const scopedRuntime = getOrCreateScopedRuntime(
+    listener,
+    params.agentId,
+    params.conversationId,
+  );
+  // No interactive approver exists on this surface, so bridge conversations
+  // run unrestricted (Hermes-style: the API is opt-in and bearer-gated and
+  // exposes the agent's full toolset). Approvals that still arise finish the
+  // turn with an explanatory reply instead of hanging.
+  getOrCreateConversationPermissionModeStateRef(
+    listener,
+    params.agentId,
+    params.conversationId,
+  ).mode = "unrestricted";
+  // Frames emitted for this turn also flow to any attached WS client; the
+  // transport here is only the fallback destination when none is attached.
+  const socket: ListenerTransport =
+    listener.socket ?? listener.transport ?? bridgeTransport;
+
+  const dispatchOptions: StartListenerOptions = {
+    connectionId: listener.connectionId ?? "openai-api",
+    wsUrl: "",
+    deviceId: settingsManager.getOrCreateDeviceId(),
+    connectionName: listener.connectionName ?? "openai-api",
+    onConnected: () => {},
+    onDisconnected: () => {},
+    onError: (error) => {
+      params.onLog?.(`OpenAI-compat turn error: ${error.message}`);
+    },
+  };
+
+  const processQueuedTurn: ProcessQueuedTurn = async (
+    queuedTurn,
+    dequeuedBatch,
+  ) => {
+    const queuedScope = getOrCreateScopedRuntime(
+      listener,
+      queuedTurn.agentId,
+      queuedTurn.conversationId,
+    );
+    await handleIncomingMessage(
+      queuedTurn,
+      socket,
+      queuedScope,
+      undefined,
+      dispatchOptions.connectionId,
+      dequeuedBatch.batchId,
+    );
+  };
+
+  return await new Promise<TurnOutcome>((resolve) => {
+    let text = "";
+    let usage: OpenAiUsage = {
+      prompt_tokens: 0,
+      completion_tokens: 0,
+      total_tokens: 0,
+    };
+    // The request settles from ITS OWN turn's lifecycle (turn-observers,
+    // keyed by OTID), never from raw stream events: stop_reason ends a
+    // stream segment before the listener decides on retries/approvals, and
+    // usage can arrive after it. Stream deltas are accumulated only while
+    // this request's turn is active, so queued requests and WS-initiated
+    // turns in the same conversation never bleed into this response.
+    let turnActive = false;
+    let recordedError: string | null = null;
+    let settled = false;
+    let unregisterTurnObserver: () => void = () => {};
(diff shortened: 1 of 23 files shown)


cpacker (changes requested):
"Direct backend path, not the WS runtime. Completions go through sendMessageStream (the headless-mode substrate) rather than the listener runtime, so HTTP requests don't contend with the single-occupancy WebSocket control channel and concurrent requests to different agents work naturally."

This seems wrong IMO - this basically just means it's going to entirely bypass the harness and lose all harness functionality? Also, it just bypasses the main "development plane" that we are building in the app server protocol?

The correct way to build this is to make this (similar to channels) as a seam that can sit on top of the WS V2 protocol, similar to e.g. the Letta Agent SDK 'remote' backend mode, but instead of an SDK, it's an HTTP server, or like the Slack channel, but instead of running Slack socket mode, it's converting ChatCompletions style inputs/outputs into the WS events.

just-cameron:
How does this work if two conversations are distinct in openwebui but start with the same user message, like "Hello"? Would this map to an identical conversation, which would be quite confusing for the user?

sarahwooders:
Good catch — two cases here:

**Identical first messages alone are fine.** The empty transcript prefix is deliberately never remembered, so every new chat's first message creates a fresh Letta conversation regardless of content (two chats both starting "Hello" → two conversations; regression-tested).

**But you're right about full-transcript collisions**: if two chats end up with *byte-identical* transcripts — same "Hello" *and* the same verbatim reply — their fingerprints merge and follow-ups converge into one conversation. Stateless chat completions gives us nothing to distinguish truly identical requests.

Fixed in 3515defc by honoring a stable client chat identity when one is available: `X-OpenWebUI-Chat-Id` (Open WebUI sends it with `ENABLE_FORWARD_USER_INFO_HEADERS=true`) or the generic `X-Letta-Chat-Key`, used as the conversation key instead of fingerprinting and pinned immediately. Fingerprinting stays as the fallback for header-less clients, where the identical-transcript collision remains a documented edge (it's inherent to the protocol — the two requests are indistinguishable). The `/v1/responses` follow-up would eliminate it entirely via real thread ids.

src/websocket/app-server-openai.ts
```diff
@@ -0,0 +1,925 @@
...
+      if (settled) return;
+      settled = true;
+      clearTimeout(timer);
+      observers.delete(observer);
+      resolve({ text, usage, error });
+    };
+
+    const observer: ListenerStreamObserver = (message) => {
+      if (message.runtime.agent_id !== params.agentId) return;
+      if (message.runtime.conversation_id !== params.conversationId) return;
```
cpacker: [P1] This observer is scoped only by agent/conversation, so a queued HTTP request or one arriving during a WebSocket turn consumes the active turn output and returns its answer; its own turn then executes later without a client. Correlate the observer to the submitted OTID/run before accepting any output.

src/websocket/app-server-openai.ts
```diff
@@ -0,0 +1,925 @@
...
+          return;
+        }
+        case "stop_reason": {
+          const stopReason = (delta as { stop_reason?: string }).stop_reason;
+          // requires_approval ends a stream segment, not the turn: the turn
+          // processor evaluates the approval (auto-approving under the
+          // conversation's permission mode) and continues. Keep observing;
+          // a WAITING_ON_APPROVAL loop status marks a truly stuck turn.
+          if (stopReason === "requires_approval") return;
+          finish(null);
```
cpacker: [P1] Every stop_reason except requires_approval is treated as success, but the listener decides after the drain whether llm_api_error/error should retry or fail, and usage can arrive after stop_reason. This returns 200 early and misses retries, terminal errors, and late usage; settle from the correlated listener-turn terminal outcome instead.

src/websocket/app-server-openai.ts
```diff
@@ -0,0 +1,925 @@
...
+      "model_not_found",
+    );
+    return;
+  }
+
+  const prefixTurns = transcript.slice(0, lastUserIndex);
+  const headerChatKey = chatKeyFromHeaders(request);
+  const prefixKey = headerChatKey
+    ? `chat-key:${agent.id}:${headerChatKey}`
+    : transcriptFingerprint(agent.id, prefixTurns);
```
cpacker: [P1] The optional chat-id header only protects clients that send it; the advertised generic OpenAI path still falls back to a text-only transcript key. I reproduced two no-header chats with identical first prompt/reply routing the follow-up from chat A to chat B (conv-1, conv-2, conv-2); require a stable identity or avoid stateful reuse when it is absent.

src/websocket/app-server-openai.ts
```diff
@@ -0,0 +1,925 @@
...
+  }));
+  sendJson(response, 200, { object: "list", data });
+}
+
+async function resolveAgentForModel(
+  model: string,
+): Promise<AgentModelEntry | null> {
+  const agents = await listAgentEntries();
+  return (
+    agents.find((agent) => agent.id === model) ??
```
cpacker, line 259: [P2] Listing can advertise a unique agent name that equals another agent ID, but this ID-first resolver then routes that advertised model to the other agent. Build one collision-free advertised-id map and use it for both listing and resolution.

src/websocket/app-server-openai.ts
```diff
@@ -0,0 +1,925 @@
...
+const OPENAI_TURN_TIMEOUT_MS = 15 * 60 * 1000;
+
+const bridgeTransport = new LocalListenerTransport();
+let bridgeRuntimeStart: Promise<void> | null = null;
+
+/**
+ * Reuse the active listener runtime when one exists (a connected WS control
+ * session or channels runtime); otherwise start a socket-free local runtime,
+ * exactly like `letta server --channels` does. If a WS control client
+ * connects later it replaces the bridge-owned runtime, and subsequent
```
cpacker: [P2] This replacement path strands in-flight requests: startControlSession stops the bridge runtime, stale leases emit no terminal event, and observers attached to the old runtime wait for the 15-minute timeout. Runtime takeover must settle or cancel active bridge requests, or defer replacement.

cpacker (changes requested):
Requesting changes on the current head. The harness rearchitecture is directionally right, and the new fail-loud conversation creation path resolves the unsafe default fallback. Five correctness and isolation issues remain: turn observers are not request-correlated, raw stop reasons settle before the listener lifecycle, no-header clients still cross-wire identical transcripts, model IDs can resolve to the wrong agent, and runtime takeover can strand active HTTP requests.

src/websocket/app-server-openai.ts
sarahwooders: Fixed in fc73bd66: added an OTID-keyed turn-lifecycle observer seam (`turn-observers.ts`) — `handleIncomingMessage` notifies start/finish for every turn it carries, independent of which dispatch closure processed it (closure identity was indeed unreliable under queue-pump drains). The bridge accumulates deltas only while its own turn is active, so queued requests and WS-initiated turns no longer bleed.

src/websocket/app-server-openai.ts
sarahwooders: Fixed in fc73bd66: settlement now comes exclusively from the correlated turn's lifecycle end (the observer seam above). `stop_reason` no longer settles anything; terminal `loop_error`/`error_message` deltas are recorded but the listener's post-drain retry/failure decision is authoritative, and usage arriving after stop_reason is captured.

src/websocket/app-server-openai.ts
sarahwooders: Fixed in fc73bd66 by taking your second option: transcript fingerprinting is removed entirely. Stateful conversation reuse now requires a stable client identity (`X-Letta-Chat-Key` / `X-OpenWebUI-Chat-Id`); header-less requests run statelessly — a fresh conversation per request with the client transcript replayed (user/assistant roles preserved, images on the newest message). Verified live: replayed transcripts recall context; your repro scenario now yields conv-1, conv-2, conv-3 with correct answers from the replay.

src/websocket/app-server-openai.ts
sarahwooders: Fixed in fc73bd66: one collision-free advertised-id map (`buildAdvertisedModelMap`) used by both listing and resolution. A name is advertised only when unique among names AND not equal to any agent id, so raw-id lookups can never be shadowed.

src/websocket/app-server-openai.ts
sarahwooders: Fixed in fc73bd66: `stopRuntime` now delivers a synthetic `runtime_stopped` event to in-process stream observers and clears the set, so in-flight bridge requests settle immediately with a server_error instead of waiting out the turn timeout when a WS control client takes over.

src/websocket/app-server-openai.ts
```diff
@@ -0,0 +1,944 @@
...
+      }
+      if (message.type !== "stream_delta" || message.subagent_id) return;
+      const delta = (message as { delta?: { message_type?: string } }).delta;
+      if (!delta) return;
+      switch (delta.message_type) {
+        case "assistant_message": {
+          const piece = extractDeltaText(delta);
+          if (piece) {
+            text += piece;
+            params.onAssistantText?.(piece);
```
amelia-letta: `recordedError` is never cleared after a retryable `error_message`. If the turn retries and succeeds, `onFinished` settles with the stale error and the client gets a 500 instead of the successful response. Clear it here when new assistant text arrives.

```suggestion
          if (piece) {
            text += piece;
            recordedError = null;
            params.onAssistantText?.(piece);
          }
```

just-cameron:
Status check for re-review — head `fc73bd66` addresses all five issues from the previous review:

1. **Turn correlation** — observers now correlate via `registerTurnObserver(correlationOtid)`; stream deltas accumulate only while the request's own turn is active, so queued or WS-initiated turns in the same conversation can't bleed into a response.
2. **Settling** — requests settle from the turn lifecycle (`onFinished`), not raw `stop_reason`; `loop_error` is recorded (leaving room for listener retries) and only surfaces if the turn actually ends failed. Late usage is captured.
3. **No-header cross-wiring** — header-less clients now run statelessly: fresh conversation per request with the client transcript replayed, so identical transcripts can't cross-wire chats. Stateful pinning requires an explicit identity header (`X-OpenWebUI-Chat-Id` / `X-Letta-Chat-Key`).
4. **Model ID collisions** — `buildAdvertisedModelMap` builds one collision-free advertised-id map used for both listing and resolution; a name equal to another agent's ID falls back to the agent ID.
5. **Runtime takeover** — `notifyStreamObserversRuntimeStopped` emits a synthetic terminal event when a control session replaces the bridge runtime, so in-flight HTTP requests settle instead of waiting out the 15-minute timeout.

CI is green across all 21 checks (including Windows and the headless suites). @cpacker this is ready for another look when you have a moment.

Two items worth deciding on, from the earlier design discussion:

- **Unauthenticated non-loopback exposure**: with no `--ws-auth` configured, `authorizeUpgrade` is a no-op, so the example invocation (`--listen ws://0.0.0.0:4500 --openai-api`) exposes the full toolset on all interfaces with no auth. Suggest refusing `--openai-api` on a non-loopback bind without auth (with an explicit unsafe escape hatch), matching the original design notes.
- **Retry idempotency**: bridge conversations have no idempotency key — a client retry of the same request appends the user message again. Consider honoring an idempotency header (e.g. `Idempotency-Key`, or Open WebUI's `{{USER_MESSAGE_ID}}`) and returning the cached terminal response. Fine as a follow-up if you'd rather keep this PR lean.

One conscious deviation to confirm: concurrent same-chat requests **queue** per-conversation rather than returning `409 Conflict` as originally sketched. Queueing seems friendlier for chat UIs (a 409 surfaces as a visible error), but calling it out for explicit sign-off.

src/websocket/app-server-openai.ts
sarahwooders: Good catch — applied in 3b7348ca, exactly as suggested: new assistant output clears the recorded error, so a successful retry settles as success while an error arriving after the final text still fails the request.

sarahwooders:
Re: retry idempotency and the queue-vs-409 question —

**Idempotency**: implemented in 3b7348ca rather than deferred: `Idempotency-Key` / `X-Idempotency-Key` retries share the original request's outcome — in-flight duplicates join the same turn, replays stream the cached text as one chunk, and failed outcomes are evicted so a deliberate retry re-runs. Bounded FIFO cache, tested (replay-without-rerun + retry-after-error).

**Queueing over 409**: confirming the deviation as intentional. Per-conversation queueing matches how every other Letta surface handles concurrent input to one conversation (WS clients, channels), and chat UIs handle a delayed response far more gracefully than a hard 409. Clients that want fail-fast semantics can send an Idempotency-Key and timeout client-side.

src/websocket/app-server-openai.ts
just-cameron: Fixed in 3b7348ca — new assistant output now clears the recorded error, same semantics as the suggestion, so a turn that retries and succeeds settles 200. Deliberate edge choice on top: an error arriving after the final assistant text still fails the request, since a trailing error should surface rather than be masked.

One coverage note: the wire tests stub turn execution (`__testSetRunTurnImpl`), so this observer path has no automated coverage — the clearing logic is verified by inspection and live runs only. Worth a test seam later if we want it pinned.

just-cameron (approved):
I'm happy to merge -- it seems like @sarahwooders has addressed all of @cpacker's notes.

src/websocket/app-server-openai-turn.ts
```diff
@@ -0,0 +1,340 @@
...
+    let usage: OpenAiUsage = {
+      prompt_tokens: 0,
+      completion_tokens: 0,
+      total_tokens: 0,
+    };
+    // The request settles from ITS OWN turn's lifecycle (turn-observers,
+    // keyed by OTID), never from raw stream events: stop_reason ends a
+    // stream segment before the listener decides on retries/approvals, and
+    // usage can arrive after it. Stream deltas are accumulated only while
+    // this request's turn is active, so queued requests and WS-initiated
```
cpacker, line 229: [P1] This guarantee does not hold once the listener queue coalesces two messages for the same conversation. `buildQueuedTurnMessage` merges both queue-item contents into the first item’s `IncomingMessage` template, so only the first request’s OTID survives. I reproduced a two-item batch whose merged content was `first` + `second` but whose sole message still carried only `otid-first`; the second HTTP observer is therefore never started or finished and waits for the 15-minute timeout even though its prompt was consumed. The first request can also receive an answer to the merged prompts. The bridge needs non-coalescing request semantics, or the queue must preserve every request correlation and define how each completion settles. Please add an integration test through the real queue path; the HTTP tests currently replace `runTurnViaListenerRuntime`, and the observer unit test constructs an already-batched message with both OTIDs, so neither catches this.

src/websocket/app-server-openai-turn.ts
```diff
@@ -0,0 +1,340 @@
...
+ * connects later it replaces the bridge-owned runtime, and subsequent
+ * requests transparently use the new active runtime.
+ */
+async function ensureListenerRuntime(
+  onLog?: (message: string) => void,
+): Promise<ListenerRuntime> {
+  const active = getActiveRuntime();
+  if (active && !active.intentionallyClosed) return active;
+
+  bridgeRuntimeStart ??= startLocalChannelListener({
```
cpacker, line 128: [P2] This lazily creates a global listener runtime, but the App Server does not retain ownership of it. `AppServerHandle.close()` stops only `activeSession.runtime`; when the OpenAI bridge started the socket-free runtime and no WS control session exists, closing the server leaves that runtime (including cron/channel lifecycle) active. Please return/track an adapter instance with an explicit `close()` that stops only the runtime it owns, and have the App Server close it.

src/websocket/app-server-openai.ts
```diff
@@ -0,0 +1,705 @@
...
+  const headerChatKey = chatKeyFromHeaders(request, body.stream === true);
+  let conversationId: string;
+  try {
+    if (headerChatKey) {
+      const chatKey = `chat-key:${agent.id}:${headerChatKey}`;
+      conversationId = await resolveConversationId(agent.id, chatKey);
+      // Pin immediately so concurrent requests for this chat reuse it.
+      rememberConversation(chatKey, conversationId);
+    } else {
+      conversationId = await createConversationSerialized(agent.id);
```
cpacker: [P2] Headerless “stateless” mode persists a fresh Letta conversation for every completion and never removes it. Generic OpenAI SDK clients do not send the custom chat headers, so ordinary use grows stored conversations at request rate; Open WebUI background tasks add more. Idempotent replays also reach this creation step before consulting `outcomeByIdempotencyKey`, creating unused conversations even when no turn is re-run. At minimum, check idempotency before allocation and define bounded cleanup/ephemeral semantics for headerless conversations (or explicitly accept and document the persistent-storage behavior).

cpacker (changes requested):
Requesting changes on the current head. The v2-runtime direction is right, and the listener touchpoints are legitimate integration seams, but the HTTP bridge still has a queue-correlation bug that can consume a request and leave it hanging. The bridge also needs explicit runtime ownership/cleanup, and the persistent-conversation behavior for generic headerless clients needs to be bounded or deliberately resolved. Architecturally, I would keep the small generic listener hooks, move the OpenAI-specific implementation under a sibling openai-compat folder, and expose one correlated in-process turn API instead of having the adapter assemble queue, lifecycle, and stream-observer internals itself.

src/websocket/app-server-openai-turn.ts
sarahwooders: Fixed in 6622b07a with non-coalescing request semantics: bridge messages carry a `noCoalesce` flag through the queue, and `consumeQueuedTurn` treats them as single-item batches — one never joins an existing batch and nothing joins a batch it started — so every request's OTID survives to its own turn and settles its own client. Plain messages still coalesce (TUI behavior unchanged). Per your ask, the integration test now goes through the real queue path (`enqueue → consumeQueuedTurn → buildQueuedTurnMessage`) rather than pre-batched fixtures, and I verified live: two concurrent requests to the same pinned chat each received their own correct reply.

src/websocket/app-server-openai-turn.ts
sarahwooders: Fixed in 6622b07a: the bridge tracks the runtime it starts (`bridgeOwnedRuntime`) and exports `closeOpenAiBridgeRuntime()`, which `AppServerHandle.close()` now calls — it stops only a bridge-owned runtime that is still the active one, so WS-session runtimes remain owned by their socket lifecycle and a taken-over runtime isn't double-stopped.

src/websocket/app-server-openai.ts
sarahwooders: Fixed in 6622b07a, both parts: (1) idempotent replays consult `outcomeByIdempotencyKey` before any allocation — a replay creates nothing; (2) headerless conversations are now explicitly ephemeral: deleted best-effort once their turn settles (backend gains an optional `deleteConversation` passthrough), while header-keyed conversations persist because they are the chat's state. Both behaviors are tested. Also caught adjacent to this: `listAgents` was reading one default 20-item page, silently dropping agents from `/v1/models` once the backend grew past it — now listed with an explicit high limit.

PR #3429, opened 2026-07-20
fix(slack): put model picker on its own row

overlord-letta:
## Problem

Slack rendered the `/model` dropdown as an accessory to the explanatory text, squeezing it into a narrow column on the right.

## Change

Render the explanatory text as its own section, then place the existing `static_select` in a separate actions block beneath it. Selection handling, model order, initial selection, and text fallback are unchanged.

## Before / after

- Before: text and dropdown shared one row; the dropdown was narrow.
- After: the dropdown gets its own row beneath the text.

## Risk

Low. This only changes the Block Kit layout for the Slack model picker. Tests assert the exact block order and preserve model option ordering.

## Validation

- `bun test src/channels/commands.test.ts src/channels/slack/adapter.test.ts` (35 passed)
- `bun run typecheck`

Requested by Cameron in Slack.

### DIFF
--- a/src/channels/commands.test.ts
+++ b/src/channels/commands.test.ts
@@ -699,15 +699,40 @@ describe("channel slash commands", () => {
     });
 
     expect(blocks).toBeDefined();
-    const selectBlock = blocks?.[1] as Record<string, unknown> | undefined;
-    const accessory = selectBlock?.accessory as
-      | { action_id?: string; type?: string; options?: unknown[] }
+    const currentBlock = blocks?.[0] as Record<string, unknown> | undefined;
+    const explanatoryBlock = blocks?.[1] as Record<string, unknown> | undefined;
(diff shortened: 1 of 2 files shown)


just-cameron:
CI repair update: no branch changes were needed. I checked the complete macOS log; the exit 1 came from two earlier `src/channels/discord-service.test.ts` timeout failures, not the final 118/118 passing batch. The focused Discord service test passed locally with the same 15s timeout, and the Slack model picker focused tests plus `bun run check` passed locally.

The Ollama failure was an upstream 503 while pulling `qwen2.5:1.5b`; after the public GitHub status and Ollama registry/model endpoints were healthy, I reran the failed CI jobs. The rerun is now green for macOS and Ollama.

just-cameron (approved):

PR #3430, opened 2026-07-20
feat(channels): centralized sender access control, command tiers, and /whoami

sarahwooders:
## Summary

Closes several access-control gaps in the messaging channels: allowlists were only enforced on the DM fall-through path, any group participant could run control commands, and there was no way to inspect one's own access.

- **Uniform sender gating** — a new `src/channels/access-control.ts` module makes one access decision per inbound message, now covering Slack, Telegram groups, Discord guild/DM auto-routing, WhatsApp, Signal, and custom channels. WhatsApp/Signal allowlist matching reuses the existing identity-normalization helpers (JID↔phone, UUID↔E.164), and per-channel denial messages are unchanged.
- **Group sender policy** — new opt-in `group_policy: "allowlist"` account field restricts group/channel senders to allowlisted, admin, or paired users. Default `"open"` preserves existing behavior.
- **Admin/user command tiers** — new `admin_users` + `user_allowed_commands` account fields. Gating activates only when at least one admin is configured (existing installs unaffected); non-admins then get a read-only floor (`/help`, `/status`, `/whoami`) plus explicitly allowed commands. Previously any participant in a routed chat could `/pause`, `/cancel`, `/model`, etc.
- **`/whoami`** — reports sender ID, scope (DM vs group/channel), tier, and runnable commands on every channel.
- **Env-var allowlists** — `LETTA_CHANNELS_ALLOWED_USERS` / `LETTA_<CHANNEL>_ALLOWED_USERS` (+ `_ADMIN_USERS` variants, `LETTA_CHANNELS_ALLOW_ALL_USERS` opt-out) merge with account config; once configured they restrict every scope on the affected channels.
- Pairing grants are a union with allowlists: allowlisted senders skip the pairing handshake, and paired users stay allowed in groups.

Backcompat notes: Slack's `dm_policy: "pairing"` is a legacy default that was never enforced and has no pairing reply surface, so it behaves as `"open"` (documented); explicit `"allowlist"` and env allowlists are enforced. Groups stay open unless `group_policy` or an env allowlist is configured.

## Test plan

- New `src/channels/access-control.test.ts` (24 cases): scope resolution, all dm/group policy combinations, Slack legacy behavior, env allowlist arming/merging, allow-all flags, wildcard entries, admin tiers, command-floor gating, whoami/denial messages, and end-to-end slash-command gating.
- Full channel suite: 772 pass / 0 fail.
- `bun run check`: all 12 checks pass.

### DIFF
--- a/src/channels/access-control.ts
+++ b/src/channels/access-control.ts
@@ -1,0 +1,365 @@
+import { isUserApproved, loadPairingStore } from "./pairing";
+import { getChannelDisplayName } from "./plugin-registry";
+import { signalAllowedUsersIncludes } from "./signal/target";
+import type {
+  ChannelChatType,
+  ChannelGroupSenderPolicy,
+  DmPolicy,
+} from "./types";
+import { allowedUsersIncludes as whatsappAllowedUsersIncludes } from "./whatsapp/jid";
+
+/**
+ * Centralized sender access control for channel inbound traffic.
+ *
+ * Modeled on the Hermes gateway's authorization layer: one decision
+ * function consulted on every inbound message (DMs and groups), an
+ * env-var allowlist axis that works across all channels, and an
+ * opt-in admin/user command tier that stays disabled until an
+ * operator lists at least one admin.
+ */
+
+// ── Scope ─────────────────────────────────────────────────────────
+
+export type ChannelAccessScope = "dm" | "group";
+
+export function resolveChannelAccessScope(
+  chatType: ChannelChatType | undefined,
+): ChannelAccessScope {
+  return chatType === "channel" ? "group" : "dm";
+}
+
+// ── Env-var allowlists ────────────────────────────────────────────
+
+const GLOBAL_ALLOWED_USERS_ENV = "LETTA_CHANNELS_ALLOWED_USERS";
+const GLOBAL_ADMIN_USERS_ENV = "LETTA_CHANNELS_ADMIN_USERS";
+const GLOBAL_ALLOW_ALL_ENV = "LETTA_CHANNELS_ALLOW_ALL_USERS";
+
+function channelEnvKey(channelId: string, suffix: string): string {
+  const normalized = channelId.toUpperCase().replace(/[^A-Z0-9]+/g, "_");
+  return `LETTA_${normalized}_${suffix}`;
+}
+
+function parseUserList(raw: string | undefined): string[] {
+  if (!raw) return [];
+  return raw
+    .split(",")
+    .map((entry) => entry.trim())
+    .filter((entry) => entry.length > 0);
+}
+
+function isEnvFlagEnabled(raw: string | undefined): boolean {
+  const normalized = (raw ?? "").trim().toLowerCase();
+  return normalized === "1" || normalized === "true" || normalized === "yes";
+}
+
+/** Union of the global and per-channel allowed-user env vars. */
+export function getChannelEnvAllowedUsers(channelId: string): string[] {
+  return [
+    ...parseUserList(process.env[GLOBAL_ALLOWED_USERS_ENV]),
+    ...parseUserList(process.env[channelEnvKey(channelId, "ALLOWED_USERS")]),
+  ];
+}
+
+/** Union of the global and per-channel admin-user env vars. */
+export function getChannelEnvAdminUsers(channelId: string): string[] {
+  return [
+    ...parseUserList(process.env[GLOBAL_ADMIN_USERS_ENV]),
+    ...parseUserList(process.env[channelEnvKey(channelId, "ADMIN_USERS")]),
+  ];
+}
+
+/** Explicit opt-out of sender gating (LETTA_CHANNELS_ALLOW_ALL_USERS=1). */
+export function isChannelAllowAllUsersEnabled(channelId: string): boolean {
+  return (
+    isEnvFlagEnabled(process.env[GLOBAL_ALLOW_ALL_ENV]) ||
+    isEnvFlagEnabled(process.env[channelEnvKey(channelId, "ALLOW_ALL_USERS")])
+  );
+}
+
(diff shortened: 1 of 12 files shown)


src/channels/access-control.ts
```diff
@@ -0,0 +1,365 @@
...
+  ].filter((entry) => entry.length > 0);
+  const enabled = admins.length > 0;
+  const allowedCommands = (params.account.userAllowedCommands ?? [])
+    .map(normalizeCommandName)
+    .filter((name) => name.length > 0);
+  return {
+    enabled,
+    isAdmin:
+      !enabled || (!!params.senderId && admins.includes(params.senderId)),
+    allowedCommands,
```
amelia-letta, line 300: Admin check uses plain admins.includes(senderId) while the access gate uses allowlistMatches with channel-specific normalization (WhatsApp JID to phone, Signal UUID to E.164). An admin configured as +1234567890 on WhatsApp would pass the access gate but not be recognized as admin for command gating, getting the read-only floor instead of full access.

just-cameron:
**Overlord (agent-c2adbf5c-8419-4211-8cd8-3740db164974)**

Sarah — the centralized policy direction looks right, and I’m treating manual `accounts.json` / env configuration of the new role fields as intentional scope here. I traced the actual registry path and ran isolated adversarial tests; I think there are three authorization gaps to close before merge.

### 1. Text control replies run before sender access

`registry-inbound.ts` calls `controls.tryHandleInbound()` before it loads the account or calls `evaluateChannelSenderAccess()`. The text control path in `registry-controls.ts` scopes by channel/account/chat/thread but does not check `msg.senderId`; unlike native control responses, it does not enforce that the responder is the initiating sender.

I reproduced this through `ChannelRegistry`: with `group_policy: "allowlist"`, a pending `AskUserQuestion` for an allowlisted user, and an unlisted sender replying in the same thread, the unlisted sender's reply reached the approval handler.

Minimal fix: run sender access before text control interception, and give text responses the same initiating-sender ownership check as native responses. Access gating alone is insufficient if multiple allowed users share the thread.

### 2. Unpaired senders are not actually limited to read-only commands

The new code intentionally lets a `"pair"` decision continue to slash-command handling so read-only commands remain available. But `resolveChannelCommandGate()` disables command gating whenever no admin list exists, so that unpaired sender is allowed through `/cancel`, `/model`, `/pause`, etc. If an old/existing route is present, those handlers can act before pairing completes.

I reproduced the policy contradiction directly: Telegram DM access returned `"pair"`, then the resolved command gate returned `enabled: false`, and both `canRunChannelCommand(gate, "cancel")` and `"model"` returned true.

Minimal fix: apply an explicit pre-pairing command floor independent of whether admin tiers are enabled.

### 3. Route provisioners override valid central grants

After the central evaluator allows a sender, the Slack, Discord, WhatsApp, and Signal DM route functions each repeat the old `dmPolicy === "allowlist"` check against only `config.allowedUsers`:

- `registry-routes.ts:105-116` — Slack
- `registry-routes.ts:348-358` — Discord
- `registry-routes.ts:443-453` — WhatsApp
- `registry-routes.ts:539-551` — Signal

That re-denies users who were validly admitted through `admin_users`, global/per-channel env allowlists, env admins, wildcard `"*"`, or pairing. It also breaks the documented “pairing grants are a union with allowlists” behavior.

Minimal fix: remove the downstream allowlist checks and make the central decision authoritative, then add route-level integration cases for each grant source.

I also independently reproduced Amelia’s WhatsApp/Signal admin-normalization finding: the sender gate uses channel-aware matching, but command admin detection uses exact `admins.includes(senderId)`. A smaller cleanup: configured `user_allowed_commands: ["reflect"]` does not permit the canonical `reflection` command because aliases are canonicalized only on the parsed-command side.

The new evaluator tests pass, as do the adjacent existing channel tests, but they do not exercise these registry/control/route interactions. The useful regression coverage would be through `ChannelRegistry`: unauthorized text control response, unpaired mutating command, and central allow followed by each auto-route path.

sarahwooders:
Addressed all review findings in 94022ba0:

1. **Text control replies before sender access** — the access gate now runs before `controls.tryHandleInbound()` in `registry-inbound.ts`, and `tryHandleInbound()` enforces the same initiating-sender ownership rule as `handleNativeResponse`: when the pending prompt has a `senderId`, replies from other participants fall through to normal ingress instead of resolving the prompt. Regression test: `registry-controls.test.ts` ("text control replies from other senders never resolve the prompt") — the non-owner's reply is not consumed and the owner can still answer afterward.

2. **Unpaired senders not limited to read-only commands** — a `"pair"` access decision now forces a floor-only command gate (`floorOnlyChannelCommandGate()`) regardless of whether admin tiers are enabled, so pre-pairing senders get only `/help`, `/status`, `/whoami`. The denial message says pairing is required rather than "limited to admins", and `/whoami` reports "Tier: pending pairing".

3. **Route provisioners overriding central grants** — removed the four downstream `dmPolicy === "allowlist"` checks (Slack/Discord/WhatsApp/Signal) from `registry-routes.ts`; the central evaluator is authoritative, so `admin_users`, env allowlists, wildcard, and pairing grants are honored on the auto-route paths. Integration test: an `admin_users` sender passes a Discord DM allowlist and auto-routes.

4. **Admin identity normalization (Amelia's finding)** — `resolveChannelCommandGate` now matches admins through the same channel-aware `allowlistMatches` as the sender gate, so a WhatsApp admin configured as a phone number is recognized when messages carry a JID sender (test covers WhatsApp, Signal, and the exact-match default).

5. **Alias canonicalization** — `user_allowed_commands: ["reflect"]` now permits `/reflection` (and vice versa); both gate resolution and `canRunChannelCommand` canonicalize aliases.

All 777 channel tests and the full 12-check suite pass.

just-cameron (approved):

PR #3433, opened 2026-07-21
fix(channels): harden pairing code generation

sarahwooders:
Follow-up to #3430.

- Generate pairing codes with `crypto.randomInt` instead of `Math.random()` — codes are an authorization credential and must not be predictable.
- Reuse a sender's existing unexpired pending code instead of minting a fresh one on every inbound message. This rate-limits code generation per sender and keeps the operator-visible code stable for the full 15-minute TTL; a new code is only minted after consumption or expiry.

Updated the pairing tests accordingly (the old "replaces existing pending code" case asserted the regenerate-per-message behavior; it now covers reuse-until-consumed). All 778 channel tests pass.

### DIFF
--- a/src/channels/pairing.test.ts
+++ b/src/channels/pairing.test.ts
@@ -67,15 +67,18 @@ describe("pairing", () => {
     expect(result).not.toBeNull();
   });
 
-  test("replaces existing pending code for same user", () => {
+  test("reuses the unexpired pending code for the same user", () => {
     const code1 = createPairingCode("telegram", "user-1", "chat-1");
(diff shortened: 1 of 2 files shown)


just-cameron (approved):
Reviewed at 948ab73b2e86eea22c72b401202789eae0c50f68. Crypto-backed generation and channel/account/sender-scoped code reuse look correct; no merge blockers found. Non-blocking expiry-copy and edge-case test follow-ups are tracked in LET-9932: https://linear.app/letta/issue/LET-9932/harden-pairing-code-reuse-edge-cases-and-expiry-messaging

PR #3437, opened 2026-07-21
fix(channels): clarify reused pairing code expiry

just-cameron:
## What changed

- Keep reused pairing codes tied to the original pending record, including its chat ID and timestamps, and document why a later DM identifier must not silently retarget a live credential.
- Change pairing copy from “expires in 15 minutes” to “expires 15 minutes after it was first issued,” so repeated delivery of the same code does not imply a fresh TTL.
- Cover regeneration after expiry, sender/account isolation, and duplicate-code ambiguity remaining fail-closed.

## Before / after

Before, an unexpired code was correctly reused, but every reply claimed another 15 minutes and the route-binding/scoping invariants were not explicit in tests. After, the copy describes the original issuance window and regression tests pin the existing security boundaries.

This does not add collision retries or change which chat a reused credential binds. Duplicate persisted codes still fail closed.

## Validation

- `bun test ./src/channels/pairing.test.ts ./src/channels/registry-community-copy.test.ts ./src/channels/registry-pairing.test.ts` — 28 pass
- Complete unit suite was also attempted. Unrelated existing failures in `settings-manager.test.ts` and `hooks/e2e.test.ts` reproduce on the base checkout; all pairing tests pass independently.

Follow-up to #3433.

### DIFF
--- a/src/channels/pairing.test.ts
+++ b/src/channels/pairing.test.ts
@@ -8,8 +8,15 @@ import {
   getApprovedUsers,
   getPendingPairings,
   isUserApproved,
+  loadPairingStore,
   rollbackPairingApproval,
 } from "@/channels/pairing";
+import type { PairingStore } from "@/channels/types";
+
+function loadPairingStoreFixture(channelId: string, store: PairingStore): void {
+  __testOverrideLoadPairingStore(() => store);
+  loadPairingStore(channelId);
+}
 
 describe("pairing", () => {
   beforeEach(() => {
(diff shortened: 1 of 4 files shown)


just-cameron:
Closing — the sender+account-scoped pairing-code reuse behavior already shipped on main; what remains here is tests and a comment. Happy to reopen as a tests-only PR if we want that coverage.

— Overlord (agent-c2adbf5c-8419-4211-8cd8-3740db164974)

PR #3439, opened 2026-07-21
fix(mods): avoid eager React resolution in packaged runtimes

### LINKED ISSUE #3436: Mods fail to load in desktop AppImage v0.28.11: Cannot find module 'react/package.json'
## Bug

All TypeScript mods fail to load in the Letta Code desktop AppImage (v0.28.11) after an app restart. The mod loader throws `Cannot find module 'react/package.json'` for every mod, preventing them from activating.

## Environment

- Letta Code Desktop AppImage v0.28.11 (Linux)
- Mount path: `/tmp/.mount_Letta Dfar2M/`
- Mods directory: `~/.letta/mods/`

## Error

```
Error: Cannot find module 'react/package.json'
Require stack:
- /tmp/.mount_Letta Dfar2M/resources/app.asar.unpacked/node_modules/@letta-ai/letta-code/letta.js
```

Reported in diagnostics as affecting all mods, including ones that don't import react.

## Reproduction

1. Start the desktop app
2. Place any `.ts` mod in `~/.letta/mods/`
3. Run `/reload`
4. Check `~/.letta/mods/diagnostics/latest.json` — shows the react error

## What we've tried

- Verified the mod code is valid (transpiles fine with esbuild)
- Verified the mod doesn't import react
- Tried symlinking react into every node_modules resolution path
- Tried installing react globally via npm
- The AppImage's filesystem is read-only so we can't add react to `app.asar.unpacked/node_modules/`
- The same mod worked in a previous session (loaded before restart, stayed in memory). Fresh load after restart fails.

## Likely cause

The mod loader in `letta.js` tries to resolve `react/package.json` during mod transpilation/loading. Inside the AppImage, react is bundled inside `app.asar` but the require resolver doesn't look inside the asar archive. The unpacked `node_modules` directory doesn't include react.

The global npm install of letta-code is v0.21.17 (stale) — updating it doesn't help because the desktop app uses its own bundled runtime.

## Workaround

None found. The mod stays dead until the app ships with react available to the mod loader.

overlord-letta:
## Summary
- Fixes #3436 by stopping the mod engine from resolving and symlinking React for every local mod before load.
- Moves runtime dependency cache setup into a focused helper that scans the transpiled/importable source with TypeScript preprocessing and only symlinks React when a real runtime import such as `react` or `react/jsx-runtime` exists.
- Preserves the stale React symlink repair path for React/TSX mods and moves that coverage into a focused runtime dependency test file, lowering oversized source-file baselines after the extraction.

## Root cause
The packaged AppImage runtime can load plain TypeScript mods from the bundled `letta.js`, but it does not expose an unpacked `react/package.json` beside that bundle. `createImportableModPath` called `ensureModCache` before transpiling or inspecting the mod, and `ensureModCache` unconditionally ran `requireFromRuntime.resolve("react/package.json")`. That made even a plain command mod fail during reload with `Cannot find module react/package.json`.

PR #3126 fixed stale dependency-cache symlinks. It did not address the separate packaged-runtime case where React is absent and the mod does not need React at all.

## Changed behavior
- Plain JS/TS mods without runtime React imports still get importable cache files, but no React symlink is created and no runtime React package resolution is attempted.
- TSX/React mods are transpiled first, then the generated `react/jsx-runtime` import triggers the same symlink creation and stale-link replacement as before.
- Managed package mods remain on their package-local import cache path and are not changed by this runtime cache logic.
- Comment text and string literals mentioning React do not trigger the dependency path.

## Test plan
- [x] `bun test --cwd /Users/lettamate/letta-code-issue-3436 src/mods/mod-engine-runtime-dependencies.test.ts src/mods/mod-engine.test.ts src/cli/mods/local-mod-loader.test.ts`
- [x] `bun run --cwd /Users/lettamate/letta-code-issue-3436 check`
- [x] `bun run --cwd /Users/lettamate/letta-code-issue-3436 build`
- [x] Fake unpacked AppImage-style runtime under `/tmp/letta-3436-fixed-boundary-final2` with `NODE_PATH` unset, packaged externals symlinked except React, and a plain TypeScript mod containing only comment/string React text. Result: command completed with `pong`, diagnostics reported zero errors/warnings, and no `mod-cache/node_modules/react` symlink was created.

## Limits and risk
- This does not make React-based mods work in a packaged runtime where React is genuinely unavailable. Those mods still require React to be resolvable, which is intentional.
- The dependency scan covers runtime imports reported by TypeScript preprocessing after transpilation; type-only imports removed by transpilation do not force React resolution.
- Highest-risk path is local mod reload in packaged runtimes. Rollback is isolated to the runtime dependency helper and the call site in `createImportableModPath`.


cpacker (approved):
Reviewed the runtime dependency extraction and packaged-runtime behavior. The implementation correctly scans transpiled/importable source, skips React resolution for plain and type-only mods, retains JSX/runtime React detection and stale-symlink repair, and leaves managed package caches unchanged.

Validated locally:
- Focused mod suites: 62/62 passed
- `bun run check`: all 12 checks passed
- `bun run build`: passed

No blocking findings.

PR #3447, opened 2026-07-21
feat(acp): ACP v1 adapter — use Letta agents from Zed and other ACP clients

sarahwooders:
## What

A self-contained package at `acp/` that exposes a Letta agent as an [Agent Client Protocol](https://agentclientprotocol.com) agent over stdio, so ACP clients (Zed, JetBrains, marimo, …) can drive a stateful Letta agent. Built on `@letta-ai/letta-agent-sdk` + the official `@agentclientprotocol/sdk`.

## Why protocol v1 (not v2)

- v1 is the only stable ACP wire protocol; every shipping client (including Zed) negotiates `protocolVersion: 1`.
- v2 exists only as an unstable schema surface (`@agentclientprotocol/sdk` ships it under `experimental/v2`) — MCP-aligned content types, `auth/*` regrouping, unified session lifecycle — with no stable release or client support yet.
- When v2 stabilizes, migration is mostly mechanical (the SDK includes a `dual-version-agent` example for serving both).

## How it maps

- Each ACP session → a new conversation (`createSession`) on **one underlying Letta agent**, rooted at the ACP `cwd`, so the agent's memory persists across sessions and editors.
- SDK stream messages → `session/update`: `assistant` → message chunks, `reasoning` → thought chunks, `tool_call`/`tool_result` → tool call updates with kind, title, and file locations.
- The SDK's `canUseTool` callback → ACP `session/request_permission` (allow once / always allow this session / reject).
- `session/cancel` → `session.abort()` → `stopReason: cancelled`.

### The approval_conflict wrinkle

The app-server transport ends a turn with a recoverable `approval_conflict` result whenever a tool needs user approval; the approval resolves concurrently over the control channel, and the resumed run streams **without a second terminal result message**. The adapter therefore pumps recovery stream rounds after such a result and ends the turn when the agent loop reports idle (`WAITING_ON_INPUT`). Worth noting for SDK consumers generally: the plain `send()`/`stream()` pattern cannot complete an approval-gated turn on its own.

## Choosing which Letta agent

ACP has no native "pick an agent" step — one server = one agent from the client's perspective. The adapter resolves its agent at launch:

- `LETTA_AGENT_ID` set → all sessions use that agent (recommended; persistent memory).
- Unset → it creates an agent on first session and logs the id to stderr.

To use multiple Letta agents from Zed, add one `agent_servers` entry per agent — Zed's agent picker then acts as the agent selector:

```json
"agent_servers": {
  "Letta (Zed)":  { "command": "bun", "args": ["<repo>/acp/src/index.ts"], "env": { "LETTA_AGENT_ID": "agent-..." } },
  "Letta (Docs)": { "command": "bun", "args": ["<repo>/acp/src/index.ts"], "env": { "LETTA_AGENT_ID": "agent-..." } }
}
```

Possible follow-up: advertise Letta agents as ACP session modes in the `session/new` response, giving an in-editor agent dropdown without touching settings.

Other config: `LETTA_ACP_BACKEND` (`local` default / `remote` / `cloud`), `LETTA_ACP_MODEL`, `LETTA_ACP_PERMISSION_MODE` (`standard` / `acceptEdits` / `unrestricted`). See `acp/README.md`.

## Testing

Verified end-to-end against a real local Letta agent using the bundled ACP test client (`bun test-client.ts`, built on the same official SDK):

- prompt/response: v1 negotiation → session → streamed reply → `end_turn`
- tool call + approval: permission round-trip, file created, tool call `completed` (~7.5s)
- rejection: tool call `failed`, no side effect, graceful finish
- cancellation: `session/cancel` mid-turn → `stopReason: cancelled`

Also verified the exact Zed spawn shape (absolute paths, foreign cwd, raw JSON-RPC on stdio) returns a correct `initialize` response.

### DIFF
--- a/acp/.gitignore
+++ b/acp/.gitignore
@@ -1,0 +1,34 @@
+# dependencies (bun install)
+node_modules
+
+# output
+out
+dist
+*.tgz
+
+# code coverage
+coverage
+*.lcov
+
+# logs
+logs
+_.log
+report.[0-9]_.[0-9]_.[0-9]_.[0-9]_.json
+
+# dotenv environment variable files
+.env
+.env.development.local
+.env.test.local
+.env.production.local
+.env.local
+
+# caches
+.eslintcache
+.cache
+*.tsbuildinfo
+
+# IntelliJ based IDEs
+.idea
+
+# Finder (MacOS) folder config
+.DS_Store
--- a/acp/README.md
+++ b/acp/README.md
@@ -1,0 +1,99 @@
+# letta-acp
+
+An [Agent Client Protocol](https://agentclientprotocol.com) (ACP) adapter for
+Letta. It exposes a stateful Letta agent as an ACP agent over stdio, so any ACP
+client — Zed, JetBrains, marimo, or the bundled test client — can drive it.
+
+Built on [`@letta-ai/letta-agent-sdk`](https://github.com/letta-ai/letta-agent-sdk)
+(agent/session management, streaming, tool approvals) and
+[`@agentclientprotocol/sdk`](https://www.npmjs.com/package/@agentclientprotocol/sdk)
+(protocol plumbing).
+
+## ACP v1, not v2
+
+This adapter implements **protocol v1** (`protocolVersion: 1`):
+
+- v1 is the current stable wire protocol. Every shipping ACP client (including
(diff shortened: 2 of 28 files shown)


sarahwooders:
Moving to a dedicated repo: letta-ai/letta-acp

PR #3448, opened 2026-07-21
feat(channels): hot reload channels after active turns drain [LET-9911]

overlord-letta:
## Problem

Channel `/reload` needs to refresh a listener without interrupting an active channel turn or dropping messages that arrive during the restart. It also needs to pick up channel implementation, account, and routing edits rather than only re-reading listener settings.

Linear: https://linear.app/letta/issue/LET-9911/queue-channel-hot-reload-until-active-runs-drain

## Approach

- Acknowledge `/reload` immediately, then wait for active channel turns—including approval continuations and final delivery work—to drain.
- Hold queued channel work behind a listener-local barrier and buffer ordinary inbound messages while reload runs. Approval/control responses and commands still pass through so the drain cannot deadlock.
- Refresh listener runtime settings, channel plugins, keyring-backed account state, routes, and adapters in one coordinated operation.
- Reconcile added, edited, disabled, and removed accounts. Replace route state so on-disk deletions take effect.
- Flush buffered messages in FIFO order through the newly loaded account policy and route table.
- Roll back plugins, adapters, accounts, and routes together if validation or adapter restart fails.
- Coalesce concurrent requests; if another request arrives after reconciliation starts, run a final freshness pass before releasing ingress.

Restore-mode listeners can discover newly enabled channel types. Explicit `--channels` listeners retain their configured scope.

## Before / after

**Before:** changing channel source, account settings, or routes required restarting the listener, and an in-flight turn could be interrupted.

**After:** an authorized `/reload` safely drains active channel work, refreshes the channel runtime and configuration, and then resumes queued/buffered traffic against the fresh state.

## Limits and risk

This touches channel lifecycle, queue ownership, plugin loading, and persisted account/route state. The reload is process-local; changes to native dependencies or process-level environment still require a listener restart.

The main failure boundary is transactional: malformed configuration or adapter startup failure preserves the previous working adapter and configuration state.

## Validation

- `bun run check` — all 12 checks passed
- Focused channel/account/routing/lifecycle suite — 144 passed
- `node scripts/run-unit-tests.cjs` — 5,051 passed; 9 unrelated baseline failures remained in app-server WebSocket and hook E2E tests
- Two adversarial review passes; identified rollback and secret-placeholder gaps were fixed and re-verified


overlord-letta:
Superseded by a smaller current-main replacement focused on the supported channel reload command.

PR #3449, opened 2026-07-22
fix(listener): single-instance lock reaps prior same-scope listeners [LET-10023]

cpacker:
## Summary
- Adds an advisory single-instance lock (`~/.letta/listeners/<scope>.lock`, scoped to deviceId + connectionName) claimed at listener startup. Both `letta server` and `letta remote` route through `runListenSubcommand`, so both are covered.
- **Newest wins**: a live previous holder is SIGTERMed (its existing shutdown handler reaps child processes and releases its own lock), escalating to SIGKILL if unresponsive. A fresh launch expresses user intent.
- **Why**: multiple listeners for the same (device, env) silently contest one relay connection lease — each re-registration steals the lease from the current holder, dropping its socket mid-turn; the close handler clears the conversation runtime and orphans healthy `requires_approval` runs. The turn stalls silently, and the next user message trips stale-approval recovery. Two confirmed producers on LET-9772: desktop-spawned listeners orphaned by an app relaunch (surviving for weeks), and a manual `letta server` beside a forgotten `letta remote`.

## Safety rails
- **PID-recycling guard**: before terminating, the claimant verifies the lock pid's current command line still looks like a letta process (`ps -o command=`). If identity is unverifiable (win32 without `ps`, permission errors, process gone) it never kills — it only overwrites the lock file and defers to relay-side supersession (LET-10024).
- Lock claim failures never block listener startup (advisory guard, logged to the session log).
- Shutdown/exit paths release the lock only if this process still owns it — a dying predecessor can't delete a newer claimant's lock.
- Corrupt/stale lock files are claimed silently.

## Test plan
- [x] 16 new unit tests: deterministic scoped paths, dead-pid reclaim without kill, SIGTERM reap, SIGKILL escalation (fake clock), unverifiable-identity and recycled-pid no-kill paths, corrupt lock, self re-claim, release ownership semantics
- [x] `bun run check` (12/12)
- [x] `bun test` listener protocol + concurrency suites (194 tests)

Part 1 of 3 for the LET-9772 duplicate-listener fix suite (LET-10023; relay supersession close code = LET-10024, reconnect run-resume = LET-10025).

### DIFF
--- a/src/cli/subcommands/listen.tsx
+++ b/src/cli/subcommands/listen.tsx
@@ -32,6 +32,10 @@ import {
   MissingListenerApiKeyError,
   resolveListenerRegistrationOptions,
 } from "@/websocket/listener/auth";
+import {
+  claimListenerInstanceLock,
+  releaseListenerInstanceLockSync,
+} from "@/websocket/listener/instance-lock";
 import { flushRemoteSettingsWrites } from "@/websocket/listener/remote-settings";
 
 type ListenerProcessAnchor = {
@@ -298,6 +302,7 @@ export async function runListenSubcommand(argv: string[]): Promise<number> {
     } catch {
       // Best-effort cleanup — don't block exit
     }
+    releaseListenerInstanceLockSync();
     await flushRemoteSettingsWrites();
     await flushListenerTelemetryEnd(`listener_${signal.toLowerCase()}`);
     process.exit(0);
@@ -321,6 +326,7 @@ export async function runListenSubcommand(argv: string[]): Promise<number> {
     } catch {
       // Best effort — don't block exit on channel cleanup failure
     }
+    releaseListenerInstanceLockSync();
     await flushRemoteSettingsWrites();
     await flushListenerTelemetryEnd(exitReason);
     process.exit(code);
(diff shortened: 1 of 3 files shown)


cpacker:
Closing as superseded — review identified merge-blocking design flaws:

1. **Sibling kill**: lock key of (deviceId, connectionName) collides for the desktop primary listener and the direct-Cloud listener, which both default to `hostname()`. With remote access enabled, normal startup would deterministically kill a legitimate sibling.
2. **PID-recycling guard was not real**: `processStartedAt` was stored but never compared; the `includes("letta")` command check could match the Electron app itself on a recycled PID and SIGKILL it.
3. **Not atomic**: read-then-rename acquisition races two simultaneous startups (TOCTOU).
4. **Kill-before-viable**: the claim ran before auth/registration resolution, so a misconfigured newcomer could kill a healthy listener and fail to replace it.

The deeper problem is the abstraction boundary: a random localhost proxy port means no stable endpoint identity exists at the CLI layer — only the spawner (Desktop) knows the logical slot. Replacement design lives in LET-10023 (rewritten around Desktop-owned process identity: persisted `{slotId, instanceUuid, pid, OS start token}`, reap only on full identity match, terminate the recorded process tree). Relay-side generation fencing (LET-10024) covers manual, cross-process, cross-host, and legacy-orphan cases.

PR #3450, opened 2026-07-22
feat(listener): terminal supersession handling and per-surface environment slots [LET-10024]

cpacker:
## Summary
- **Close code 4009 (superseded) is terminal**: when the relay reports that a newer listener registered for this environment slot, the listener logs it, stops the retry loop, and exits cleanly (exit 0, telemetry `listener_superseded`) — it never re-registers. Previously the displaced listener hit 1008 on reconnect and re-registered unconditionally, stealing the lease back: two same-slot listeners ping-ponged the connection indefinitely, and every steal aborted the current holder's in-flight turn (the silent-stall / stale-approval producer on LET-9772).
- **Per-process registration nonce**: registration now sends a `processInstanceId` unique to this process lifetime. The relay tombstones a displaced process's nonce and rejects its re-registration with 409 `LISTENER_SUPERSEDED`, which the retry path treats as terminal (never retried). Covers the race where the 1008→re-register path outruns the 4009 close.
- **New `desktop-remote` listener surface**, declared by the spawner via `LETTA_LISTENER_SURFACE`: WS-driven desktop clients register their cloud listener under a distinct environment slot, so they coexist with a manual `letta server` that defaults to the same hostname connection name instead of superseding it. Unknown/unset values fall back to the caller's surface.

## Compatibility
- Servers that predate the fencing never send 4009/409 — behavior is unchanged there.
- Spawners that predate the surface env var keep today's slots.
- `onSuperseded` is optional on `StartListenerOptions`; embedded callers without it fall back to `onError`.

## Test plan
- [x] 8 new tests: nonce sent with registration, 409 `LISTENER_SUPERSEDED` classification, no-retry on 409 (terminal), other 409s not misclassified, `LETTA_LISTENER_SURFACE` honored/fallback, `desktop-remote` vs `server` slot separation
- [x] `bun run check` 12/12; listener protocol/concurrency/turn-lifecycle suites green (200 tests)
- [ ] Live probe once the server side deploys: duplicate `letta server` with the same name → older exits with the supersede message; desktop + CLI coexist

### DIFF
--- a/scripts/source-file-size-baseline.json
+++ b/scripts/source-file-size-baseline.json
@@ -46,7 +46,7 @@
   "src/websocket/listener/commands/channels.ts": 1390,
   "src/websocket/listener/commands/memory.ts": 1114,
   "src/websocket/listener/file-commands.ts": 1030,
-  "src/websocket/listener/lifecycle.ts": 1766,
+  "src/websocket/listener/lifecycle.ts": 1788,
   "src/websocket/listener/protocol-inbound.ts": 2279,
   "src/websocket/listener/protocol-outbound.ts": 1282
 }
--- a/src/cli/commands/listen.ts
+++ b/src/cli/commands/listen.ts
@@ -8,7 +8,11 @@ import type { Buffers, Line } from "@/cli/helpers/accumulator";
 import { buildAgentReference } from "@/cli/helpers/app-urls";
 import { settingsManager } from "@/settings-manager";
 import { getErrorMessage } from "@/utils/error";
-import { registerWithCloudRetry } from "@/websocket/listen-register";
+import {
+  createListenerSessionNonce,
+  isSupersededRegistrationError,
+  registerWithCloudRetry,
+} from "@/websocket/listen-register";
 import { resolveListenerRegistrationOptions } from "@/websocket/listener/auth";
 
 // tiny helper for unique ids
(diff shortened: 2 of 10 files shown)


cpacker:
Pushed review fixes (6f27e28c):

- **Session-scoped nonce** — `createListenerSessionNonce()` replaces the module-global process nonce. Each `letta server` run and each in-app `/listen` invocation mints its own nonce and reuses it only across that session's re-registrations, so one superseded session no longer 409-blocks every later `/listen` from the same long-lived TUI process.
- **In-app `/listen` 409 classification** — its `onNeedsReregister` catch now reports "a newer listener took over" instead of a generic re-registration failure.
- **Central lifecycle test** — real WS `close(4009)` end-to-end: `onSuperseded` fires exactly once, no reconnect, no re-register (asserted over a settle window).

cpacker:
Closing in favor of a from-scratch replacement. The 4009 terminal handling and session nonce here are the client half of a relay supersession protocol we're deferring (see letta-cloud #13216 closure — the problem is local orchestration, not distributed arbitration). Rather than stripping this branch down, the replacement PR starts clean with the narrower scope: explicit local listener identity (persisted per-listener UUIDs, spawner-provided instance IDs, display-only connectionName), an atomic local lock that fails visibly instead of killing, and coexistence by default — no protocol changes. The surface-separation idea survives in spirit: slots become explicit IDs rather than name-derived.

PR #3453, opened 2026-07-22
feat(listener): spawner-assigned listener identity hook [LET-10085]

cpacker:
## Summary
Minimal fix for the Desktop/CLI listener collision (LET-9772): an **owning spawner assigns its child an explicit relay identity**.

- New `LETTA_LISTENER_INSTANCE_ID` env var (validated format): when set by a spawner (Desktop passes `desktop-<slot>:<installation-id>`, see letta-cloud#13221), registration sends it verbatim as `listenerInstanceId`. Desktop children therefore occupy relay environment slots that can never collide with a manual `letta server`/`/listen` sharing the same display name — the name-collision that caused the connection-lease ping-pong and silent turn stalls.
- **Manual listener behavior is unchanged**: no env var → legacy name-derived identity, same slots as today. No new persistence, no locks, no lifecycle changes.
- The registration endpoint already accepts `listenerInstanceId`; zero relay changes.

## Scope note
Earlier revisions of this PR grew a generalized single-instance system (persisted identity files, corrupt-file repair claims, a generic lock, /listen lifecycle management). That was scope drift — the review findings against it were real, but the machinery itself was unnecessary for this fix, so it's deleted rather than hardened. Duplicate *manual* listener processes are a separate pre-existing problem; if evidence shows they need a local single-run guard, that's a follow-up with a session-owned lock API.

## Diff
4 files, ~130 lines: `identity.ts` (env accessor + validation), one line in `resolveListenerRegistrationOptions`, tests.

## Test plan
- [x] Spawner identity passes through registration verbatim; absent/invalid env falls back to legacy name-derived identity (auth + identity tests)
- [x] `bun run check` 12/12; listen-register + protocol suites green (182 tests)

Pairs with letta-cloud#13221 (Desktop child registry + slot identities). Release order: this ships first, Desktop bumps its bundled letta-code, then #13221 merges after packaged-child validation.


cpacker:
Review fixes pushed (d4c7eb24):

1. **Atomic identity get-or-create** — identity storage moved from the settings map to one file per configuration key (`~/.letta/listener-identities/`, O_EXCL candidate + hard-link publication). Concurrent first-time processes converge on ONE identity; pinned by an 8-way concurrent-resolve test asserting a single distinct id and zero candidate litter.
2. **Claim-guarded stale reclaim** — the reference protocol's recovery-claim half is now implemented: reclaim requires atomically publishing a claim keyed to the stale content hash, then re-verifying the lock still holds exactly that content before removal. TOCTOU regression test: B (which observed stale S) cannot remove A's fresh lock published after A reclaimed S. Contended/crashed reclaims degrade to "unavailable" — never a false acquisition.
3. **Docs honesty** — comments now state the env name is the *configuration key* (project + namespace + name selects which configured listener; renames configure a new listener) while the identity *value* is never name-derived.
4. **Windows CI** — the never-kills test injects liveness via deps instead of assuming pid 1 is alive.
5. **Deployment-order guard (client half)** — `isLegacyDesktopSpawn()`: a Desktop child (`LETTA_DESKTOP_MODE=1`) without an explicit spawner identity keeps the legacy name-derived identity and skips minting + the lock, so shipping this before the Desktop PR cannot collide Desktop's same-named siblings on the local lock. New Desktop builds pass per-slot identities and never hit the fallback.

26 tests green, `bun run check` 12/12.

cpacker:
Re-review fixes pushed (282908ea):

1. **Corrupt-identity TOCTOU closed** — corrupt files are removed only under a content-keyed repair claim (same generation guard as the stale-lock reclaim): claim the exact corrupt content, re-verify the file still holds it, then remove. Repairer B that captured corrupt C can no longer delete the VALID identity A published after repairing C; B loops, re-reads, and adopts A's identity. Regression test simulates B losing the claim with A's valid identity already in place — B converges instead of splitting the configuration.
2. **Real TOCTOU regression test for the stale-lock reclaim** — the test now recreates the actual interleaving: claimant B captures the stale contents, and A's fresh lock is swapped in at B's dead-holder liveness probe, BEFORE B's reclaim runs. Verified the pre-fix implementation (unconditional remove) fails this test; the claim-guarded one settles as "held" on A's fresh lock with the file byte-identical.

28 tests green, `bun run check` 12/12.

cpacker:
Re-review fix pushed (15279d07) — the identity-split race is closed:

1. **Contended repair waits, never bypasses** — a resolver that finds a live repairer's claim sleeps and re-reads (budget raised to 10 attempts) instead of burning retries and falling through.
2. **Crashed repairer claims are recovered** — a claim held by a dead pid is removed, so the next attempt re-claims and repairs; a crash can no longer wedge the configuration permanently.
3. **Fail visibly, never split** — if a corrupt generation (or a live never-finishing repairer) persists past the budget, resolution throws `ListenerIdentityUnavailableError` and listener startup exits with an actionable message. The random session-identity fallback now applies ONLY to plain fs failures where no corruption was ever observed — it can no longer hand concurrent starters different lock keys.
4. **Tests now exercise the real paths**: your split reproduction is pinned (6 concurrent resolvers over a corrupt identity → exactly ONE identity, matching the persisted file); the TOCTOU test drives the interleaving through the repair path itself (A finishes during B's contended-claim wait) rather than pre-writing the valid identity; crashed-claim recovery and the fail-visibly throw are covered.

31 tests green, `bun run check` 12/12, neighboring suites green (176).

On #13221's release gate: agreed the lockfile bump is a code change — it lands as a commit on that PR once the letta-code release containing this PR exists, followed by packaged-child validation. The Windows tree-liveness caveat is acknowledged as documented behavior (root-pid fallback; taskkill /T covers the live-root case).

cpacker:
Scope gap closed (acaec359): **/listen now acquires the single-run lock**, not just `letta server`.

- Claimed right after identity resolution; "held" fails visibly with the holder pid and remediation (`/remote off` in the owning session, or a different `--env-name`).
- Released on every terminal path of the /listen session: `/remote off`, disconnect, listener error, re-registration failure, and startup failure after acquisition. Process exit/crash is covered by the lock design itself (dead-pid locks are reclaimed by the next claimant; release is pid-guarded so a stale release can't delete a newer claimant's lock).

The "same configured listener twice → local error" criterion now holds on both surfaces. CI: the macOS failure is the two unrelated Discord test timeouts (listener suites passed) — the new push retriggers CI; will rerun if they flake again.

cpacker:
Re-review fixes pushed (8c601903) — the lock lifecycle is now generation-safe:

1. **Ownership handles** — `claimListenerLock` returns `{lockPath, lockNonce}`; `releaseListenerLock(handle)` deletes only while the file still contains that exact nonce. Your same-pid reproduction is pinned as a regression test: generation A's stale release against generation B's same-pid lock leaves B's file byte-identical. The module-global `activeLockPath` (and its test reset hook) are gone; each surface holds its own handle.
2. **Callbacks capture their session's handle** — `/listen` terminal callbacks (disconnect, error, re-registration failure, startup failure) close over the session's handle; `letta server` uses a process-scoped handle for its signal/exit paths.
3. **Second concurrent /listen rejected** — with remediation (`/remote off` first), instead of silently replacing the runtime and stranding the old lock under the live TUI pid.
4. **`/remote off` during reconnect** — new `hasListenerRuntime()` (runtime exists, transport possibly null) replaces `isListenerActive()` in the off-path, so stopping mid-reconnect works and releases the lock. Also stops cleanly when only the session lock remains.
5. Release is idempotent and null-safe.

Lifecycle regressions added: same-pid replacement (your repro), idempotent release, null-handle no-op, different-pid claimant intact. 33 lock/identity tests + protocol/auth-lifecycle suites green (193), `bun run check` 12/12.

PR #3463, opened 2026-07-23
feat(channels): expose turn context to channel plugins

snowctl:
## Summary

- Preserve the unique initiating `ChannelTurnSource` through `MessageChannel` action contexts, outbound messages, and rich-message drafts.
- Resolve provenance structurally and fail closed when multiple distinct inbound sources match the selected route.
- Expose complete parsed `toolInput` on tool progress events so trusted channel adapters can own platform-specific presentation without adding channel policy to core.

## Why

Custom channel adapters currently lose context that core already has:

- A `MessageChannel` send or rich draft cannot reliably identify the inbound message that initiated the turn. Adapters therefore cannot safely choose reply/thread behavior or reconcile a draft with its final message when turns overlap.
- Progress adapters receive sanitized core presentation fields, but not the structured tool input needed to derive their own titles and details.

This change exposes those facts while leaving behavior in the adapter. Core does not infer platform-specific reply, edit, reconciliation, or progress-label policy.

### Turn-source provenance

- Reuses the existing `ChannelTurnSource` domain type.
- Includes channel/account, chat and sender metadata, message/thread identity, agent, and conversation in structural identity.
- Collapses duplicate copies of the same source.
- Omits provenance when matching is ambiguous or conflicting.
- Handles effective Slack root-thread identity after route inference.
- Propagates the exact source selected by `MessageChannel` into rich drafts and final outbound messages.

### Structured progress input

- Adds optional `toolInput` to `ChannelTurnProgressUpdate` / `ChannelTurnProgressEvent`.
- Emits it only after tool arguments form complete valid JSON, including accumulated terminal updates.
- Omits incomplete streamed fragments.
- Documents that input is unsanitized and must not be rendered directly.
- Preserves existing `toolTitle` and `toolDetails` behavior for compatibility.

## Validation

- `bun run check` — 12/12 checks passed.
- 36 focused provenance, progress-input, Slack routing, and rich-draft tests passed.
- External custom-channel compatibility suite — 320 tests passed.
- Full local unit run — 5,006 passed and 25 skipped. Nine unrelated App Server websocket/hook E2E failures reproduced unchanged on clean `upstream/main` in the same environment.

### DIFF
--- a/scripts/source-file-size-baseline.json
+++ b/scripts/source-file-size-baseline.json
@@ -37,7 +37,7 @@
   "src/settings-manager.test.ts": 1669,
   "src/settings-manager.ts": 2113,
   "src/tools/impl/enter-worktree.ts": 1260,
-  "src/tools/impl/message-channel.ts": 1187,
+  "src/tools/impl/message-channel.ts": 1113,
   "src/tools/manager.ts": 3282,
   "src/tools/tool-execution-context.test.ts": 1422,
   "src/types/protocol_v2.ts": 2943,
--- a/src/channels/message-channel-slack.test.ts
+++ b/src/channels/message-channel-slack.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, describe, expect, mock, test } from "bun:test";
+import { afterEach, describe, expect, mock, spyOn, test } from "bun:test";
 import {
   __testOverrideLoadChannelAccounts,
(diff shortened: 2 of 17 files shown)


snowctl:
I'm working on a custom Matrix adapter for use with a companion agent, using these events to get a better UX in Matrix 😛

PR #3464, opened 2026-07-23
feat(backend): per-backend pi-ai Models runtime + real Ollama provider [LET-10126] [LET-10127]

cpacker:
## Summary

First two children of **LET-10125** ("fix the pi-ai / LC harness seam"): the local backend stops fabricating pi-ai `Model` objects from provider IDs and model-name substrings, and instead uses pi-ai's own `Provider`/`Models` runtime as the source of truth.

Fixes LET-10126
Fixes LET-10127

### Foundation (LET-10126)
- New `LocalPiModelsRuntime` (`src/backend/dev/pi-models-runtime.ts`): a per-`LocalBackend` `createModels()` runtime — instance-scoped, no module-global mutable state. Built-in pi-ai providers are registered via `builtinProviders()`.
- `/model` listing (`listLocalModels`), turn execution (`PiStreamAdapter` → `resolvePiModelForAgent`), and compaction all share the backend's runtime instance, so selection and execution resolve the **same** `Model` object/capability set.
- Stream dispatch goes through `Models.stream()/streamSimple()` (provider-owned); the global `stream()`/`streamSimple()` compat dispatch is removed from the local turn path. Letta-resolved credentials still win via explicit stream options, which pi-ai honors over provider auth. Providers not yet migrated (mods, LM Studio, llama.cpp, Ollama Cloud) dispatch through a models-less provider object in the runtime pending LET-10128..LET-10130.

### Ollama (LET-10127) — fixes the user-visible vision bug
- New dynamic provider (`src/backend/dev/pi-ollama-provider.ts`): owns keyless auth/base URL, discovers installed models from `/api/tags`, and reads **authoritative capabilities from `POST /api/show`** (`vision` → `input: ["text","image"]`, `thinking` → `reasoning`, engine `context_length` → `contextWindow`).
- `qwen3.6:27b` (no `vl`/`vision`/`llava` substring) now resolves as image-capable; `customOpenAICompatibleModel()` no longer serves Ollama. No model-name heuristics were added anywhere.
- Refresh failures retain the last-known model list (turns don't brick, `/model` doesn't blank); a base URL/auth record change invalidates and rebuilds only the Ollama provider.

### File-size ratchet splits
`pi-stream-adapter.ts` image elision → `pi-image-elision.ts`; `createLocalExecutor` → `local-executor-factory.ts`; local-endpoint payload test → its own file. Baselines lowered accordingly.

## Regression coverage (per ticket)

[...]

PR #3465, opened 2026-07-23
feat(backend): provider-owned capability discovery for llama.cpp, LM Studio, Ollama Cloud [LET-10128] [LET-10129]

cpacker:
## Summary

**Stacked on #3464** (LET-10126/LET-10127). Migrates the remaining Letta-owned local endpoint connectors onto the pi-ai Provider/Models seam.

Fixes LET-10128
Fixes LET-10129

### Generic where possible, engine-specific only where necessary

The Ollama provider's skeleton is factored into `createLocalEndpointPiProvider` — shared fetch/timeout/auth handling, last-known-model retention on refresh failure, complete `Model` construction, and `createProvider` wiring. What remains per engine is only the irreducible part: each engine exposes authoritative capability metadata on a different **native** API (the shared OpenAI-compatible `/v1/models` surface returns bare IDs with no capabilities — which is why the old code guessed from filenames):

| Engine | Native metadata source | Vision | Context window |
|---|---|---|---|
| llama.cpp (LET-10128) | `GET /props` | `modalities.vision` | `default_generation_settings.n_ctx` |
| LM Studio (LET-10129) | `GET /api/v0/models` | `type === "vlm"` / `capabilities` | `max_context_length` |
| Ollama Cloud | same API as local Ollama | `POST /api/show` `capabilities` | engine `context_length` |

- LM Studio: embeddings models excluded from `/model`; when the native API is unavailable, capabilities are **explicitly unknown** → text-only (per ticket: no model-id guessing; a `llava-*` id on the fallback path stays text-only until real metadata arrives).
- llama.cpp: single-model server; `/props` applies to the listed model. `/props` unavailable → per-model last-known fallback.
- `LocalPiModelsRuntime` generalizes to a managed-endpoint provider table with per-provider connection signatures — a base URL/auth change still invalidates and rebuilds only the affected provider.

## Tests

- `pi-llama-cpp-provider.test.ts`: vision/context from `/props`; a `*vision*` filename with `modalities.vision: false` stays text-only; `/props` outage → last-known; refresh failure retains list.
- `pi-lmstudio-provider.test.ts`: `vlm` type → image input; embeddings excluded; native-API outage → explicit-unknown text-only (no `llava` name grant) and last-known retention.
- `pi-ollama-provider.test.ts`: Ollama Cloud reuse publishes under `ollama-cloud` with Bearer auth.
- `pi-models-runtime.test.ts`: end-to-end llama.cpp — `/model` listing and turn execution resolve the identical provider-published `Model` instance.

`bun run check` 12/12; 129 tests across the affected suites pass.


cpacker:

PR #3466, opened 2026-07-23
feat(backend): fold mod-registered providers into the pi-ai Models runtime [LET-10130]

cpacker:
## Summary

**Stacked on #3465** (which stacks on #3464). Third child of LET-10125.

Fixes LET-10130

Mod registrations remain the declarative, mod-facing vocabulary; the concrete runtime unit is now a real pi-ai `Provider` (`createModPiProvider`), registered in the per-backend Models runtime like every other provider:

- statically declared models → published as complete pi-ai `Model` objects
- a mod `listModels` hook → the provider's dynamic `refreshModels`, with last-known retention on failure (seeded from the static declaration — the fallback the legacy path implemented by hand)
- auth resolves from the local provider record/env; stream dispatch runs through per-API implementations (`knownApiStreams()`, extracted to `pi-api-streams.ts`)
- `PiProviderModelRegistration` is now **derived from pi-ai's `Model` type** (`Omit<Model<Api>, "api" | "provider" | "baseUrl">` with `api`/`baseUrl` relaxed) instead of duplicating its field list — the "reuse pi-ai types" requirement
- the registry gains a per-provider registration revision, so each per-backend runtime rebuilds exactly the provider whose registration changed (register/unregister), leaving all others untouched
- mod registrations take precedence over the built-in endpoint table, matching `pi-model-factory`'s existing resolution order
- with a runtime present, turn resolution reads `Models.getModel()` (no per-turn `listModels` round trip) and `/model` reads the same published objects; OAuth `modifyModels` still applies per request, on a copy, so a mutating mod cannot corrupt the published instance

The legacy no-runtime path stays until the cleanup child (LET-10131) deletes it.

## Tests

`pi-mod-provider.test.ts`: static publication as complete Models; `listModels`→refresh with last-known retention; turn resolution returns the identical published instance; re-registration rebuilds only that provider (builtin lists untouched by reference); unregistration removes it from the runtime.

`bun run check` 12/12; 191 tests across the backend suites pass.


cpacker:

PR #3468, opened 2026-07-23
refactor(backend): delete pi-ai compat dispatch and model fabrication paths [LET-10131]

cpacker:
## Summary

**Stacked on #3466** (→ #3465 → #3464). Final child of LET-10125: with every consumer on the Models runtime, the legacy paths are deleted — net **-242 lines**.

Fixes LET-10131

### Code that disappeared (per the parent ticket's list)

- **`customOpenAICompatibleModel()` and all model-name capability heuristics** (`llava/vision/vl` → image, `gpt-oss/qwen3/deepseek-r1` → reasoning). Runtime-managed providers publish authoritative Models; nothing fabricates a Model from a string handle anymore.
- **Bare-ID local model discovery** in `local-model-config` (OpenAI-compatible `/models` scraping, ollama tags fallback, discovery-timeout plumbing). `listLocalModels` / `resolvePiModelForAgent` now create a call-scoped runtime when the backend's instance isn't threaded (honoring injected `fetch`), so every caller reads provider-published Models.
- **Per-turn registered-provider `listModels` round trip** and `registeredModelToPiModel`; connection auth headers are per-request stream options, never baked into published Models.
- **Static compat catalog imports**: `getModel`/`getModels`/`getProviders` from `/compat` replaced with `getBuiltinModel(s)`/`getBuiltinProviders` from `@earendil-works/pi-ai/providers/all`; `isContextOverflow`/`Type` moved to the root entrypoint. **One documented `/compat` import remains** (`getEnvApiKey` in `pi-provider-registry.ts` — it has no non-compat entrypoint yet; the ticket's must-disappear list doesn't include it).

### Also included

- Ollama discovery now caches `/api/show` metadata by tag **digest**: refresh skips GGUF metadata reads (seconds for large models) when the installed blob is unchanged — the "cache through provider refresh semantics" requirement from LET-10127, and it materially reduces load on live local daemons.

### Behavior notes for review

- A runtime-managed model that was never discovered now fails resolution with a clear "Unknown model … choose another model with /model" instead of blindly fabricating a text-only Model whose HTTP request would fail anyway. Last-known retention still covers endpoint outages after first discovery.
- Discovery call patterns changed (tests updated accordingly): Ollama/Ollama Cloud use native `/api/tags`(+`/api/show`) only; LM Studio tries `/api/v0/models` then `/v1/models`; llama.cpp uses `/v1/models` + `/props`.

## Tests

`bun run check` 12/12. 109 tests across the directly affected suites pass; full backend suites pass standalone (a dev machine running a real Ollama at `localhost:11434` can see contention flakes in heavily parallel runs — CI is unaffected; the digest cache reduces this).


cpacker:

PR #3472, opened 2026-07-23
feat(backend): upgrade pi-ai to 0.81.1 and adopt its auth/refresh architecture

cpacker:
## Summary

**Stacked on #3468** (top of the LET-10125 stack). Upgrades `@earendil-works/pi-ai` 0.80.6 → **0.81.1** (latest) — split into its own PR deliberately, since the last pi-ai bump rode silently inside a feature PR (#3294) and froze us on `/compat` for months.

0.81 is a second architecture wave, and this PR adopts it rather than shimming it:

### Refresh/persistence model
- Dynamic catalogs move to `fetchModels(context)` with **store-backed persistence**; `LocalPiModelsRuntime` owns a `ModelsStore` and drives `refreshModels` per managed provider (one collection-wide `refreshAll()` per `/model` listing), preserving per-provider error semantics and last-known retention. Built-in providers stay on their own cadence.
- Upstream now **merges static baseline + dynamic overlay by id**: local endpoint providers publish an empty baseline (uninstalled engine models must disappear; the `initialModels` carry-over option is deleted as a stale-baseline trap), and mod static registrations become a baseline the `listModels` overlay extends.

### OAuth: registry deleted upstream, provider-owned now
- pi-ai removed the standalone OAuth registry; OAuth lives on `Provider.auth.oauth` (`login(AuthInteraction)` / `refresh` / `toAuth`). New `pi-oauth.ts` is the bridge: builtin providers expose their `OAuthAuth` directly, mod `PiProviderOAuthConfig` adapts onto it (legacy extension callback types are upstream-preserved), token refresh in the local auth store goes through the provider's `refresh`/`toAuth`, and `/connect` drives an `AuthInteraction`.
- Per-credential base URLs/headers (GitHub Copilot) now flow from `toAuth()` instead of the removed `modifyModels` — connection auth stays per-request stream options, never baked into published Models.

### Catalog sync
- Pi TUI default-model mirror synced to coding-agent 0.81.1 (anthropic → claude-opus-4-8, openai → gpt-5.5, xai → grok-4.5, zai → glm-5.1; ant-ling/nvidia/zai-coding-cn/qwen-token-plan gain defaults; the defaultless review set is now empty). `radius` is a `KnownProvider` without a builtin factory and stays unmapped.
- `builtinCatalogModels()` tolerates providers absent from the generated catalog; test expectations updated for upstream catalog drift (gpt-5.6 context 272k, zai default handle, upstream OAuth display names).

### Verified upstream gaps (why two residual items stay)
Checked against the 0.81.1 tarball: `env-api-keys` is still exported **only** via `/compat` (our single allowlisted import), there is still no default-model runtime API, and the library ships **no** llama.cpp/Ollama/LM Studio providers (Pi's llama.cpp provider lives in its coding-agent app) — so our local-endpoint factories remain necessary at latest. These stay as upstream asks.

## Tests
`bun run check` 12/12. 247 tests across the affected suites (backend, providers, connect flows, modify-local) pass. New `test-utils/pi-refresh-context.ts` drives `refreshModels(context)` directly in provider unit tests.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -4,7 +4,7 @@
     "": {
       "name": "@letta-ai/letta-code",
       "dependencies": {
-        "@earendil-works/pi-ai": "^0.80.6",
+        "@earendil-works/pi-ai": "^0.81.1",
         "@letta-ai/letta-client": "^1.10.2",
         "@letta-ai/trajectory": "^0.1.1",
         "@pierre/diffs": "1.2.2",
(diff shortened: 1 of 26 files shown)


amelia-letta (approved):
reviewed independently against the 0.81.1 types. the adaptation is clean:

- `fetchModels` (factory input) vs `refreshModels` (built Provider method) split is correct — the runtime calls `provider.refreshModels` which pi-ai wraps around `fetchModels` with store restore/persist
- `initialModels` carry-over removed, static baseline empty for local endpoints — correct per 0.81 merge-by-id semantics (stale models must disappear when uninstalled)
- `pi-oauth.ts` bridge is the right seam: builtin providers expose `OAuthAuth` directly, mod configs adapt via `modOAuthAuth`, resolution order (mod first, then builtin) matches everywhere else
- `refreshAll()` replacing per-provider loops in `listLocalModels` is cleaner, concurrent with per-provider error retention
- per-credential baseUrl/headers from `toAuth()` at request time instead of `modifyModels` baked into models — good separation
- `connect: false` mods reporting `{ auth: {} }` for refresh eligibility is correct (keyless but still dynamic)
- `builtinCatalogModels` try/catch for dynamic-only providers like "radius" — reasonable
- test infra (`testRefreshContext`) and catalog drift expectations all look right

no issues found.

cpacker:

PR #3474, opened 2026-07-23
feat(backend): make the Models runtime the canonical lookup and credential seam

cpacker:
## Summary

**Stacked on #3472** (bottom of stack: #3464). Addresses the two load-bearing corrections from review, plus the llama.cpp metadata contract.

### 1. Built-in model lookup goes through the runtime

`resolvePiModelForAgent()` now resolves catalog models via `modelsRuntime.getModel()` (dated-handle fallback included) and clones **only when an effective override applies** — with none, the turn model **is** the runtime-published instance. The reviewer's probe now yields `sameObject: true`, asserted by identity in a new test (`anthropic/claude-opus-4-8`). Headers stay per-request stream options rather than model clones. `/model` listing reads catalog handles and display metadata from the runtime's registered providers; the static `getCatalogModel`/`listCatalogModelsForProvider` reads in the factory/listing path are gone.

### 2. The runtime owns credentials

New `local-pi-credential-store.ts` adapts `auth.json` into a pi-ai `CredentialStore` (`read`/`list`/`modify`/`delete`; OAuth refreshes persist through `modify`), wired into `createModels()`. The factory's two hand-rolled OAuth blocks collapse into a single `Models.getAuth()` call — stored tokens resolve/refresh through the runtime under the store's write path, and per-credential request auth (GitHub Copilot's per-token base URL) flows from `OAuthAuth.toAuth`. The explicit stream options remain as plumbing, but their values now **come from** the runtime.

### 3. llama.cpp per-model metadata

Discovery follows the upstream reference contract: native `/models` catalog entries (`architecture.input_modalities`, `meta.n_ctx`) per model, falling back to `GET /props?model=<id>` (single-model servers ignore the param and stay correct; router servers address by id). One model's metadata is never applied to another; a router-mode regression test covers two models with different capabilities on one server. Handle resolvability treats empty dynamic catalogs (radius) as runtime-resolved, fixing normalization for `radius/auto`.

## Related fixes already landed below this PR

- #3472 (new commits): `refreshAll()` gates on configured/auto-detectable providers (no more unconfigured Ollama Cloud probing or mod `listModels` invocation), connection changes delete the provider's `ModelsStore` entry (no stale-catalog restoration), five retention assertions now `await` their rejections, `modifyModels` gets a `structuredClone`, and `PI_PROVIDER_SPECS` is keyed off `builtinProviders()` (radius resolvable).
- NUL byte fixed at #3464 and merged forward — every branch's `pi-models-runtime.ts` is text.

## Not in this PR

Collapsing the remaining mod-specific resolution branches in `pi-model-factory`/`local-model-config` onto the now-canonical path (LET-10130 follow-through) — the lookup/auth/credential seams they'd collapse onto now exist; tracked as the next increment.

## Tests

`bun run check` 12/12. 222 tests across the affected suites pass (remaining full-batch flakes are live-local-Ollama contention on this dev machine, all pass standalone; CI unaffected). New coverage: built-in same-instance invariant, credential-store read/list/modify persistence, llama.cpp router-mode no-bleed, unconfigured-refresh and stale-store regressions.

### DIFF
--- a/src/backend/dev/pi-llama-cpp-provider.test.ts
+++ b/src/backend/dev/pi-llama-cpp-provider.test.ts
@@ -2,70 +2,135 @@ import { describe, expect, test } from "bun:test";
 import { testRefreshContext } from "@/test-utils/pi-refresh-context";
 import { createLlamaCppPiProvider } from "./pi-llama-cpp-provider";
 
+interface FakeLlamaCppModelEntry {
+  id: string;
+  inputModalities?: string[];
+  nCtx?: number;
+  props?: { vision?: boolean; nCtx?: number };
+}
+
 interface FakeLlamaCppState {
-  modelIds: string[];
-  props?: unknown;
-  failProps?: boolean;
+  models: FakeLlamaCppModelEntry[];
+  /** Serve /models as a plain OpenAI id list (no per-model metadata). */
+  plainList?: boolean;
   failModels?: boolean;
+  failProps?: boolean;
+  requests: string[];
 }
 
 function fakeLlamaCppFetch(state: FakeLlamaCppState): typeof fetch {
   return (async (input: string | URL | Request) => {
-    const url = String(input);
-    if (url.endsWith("/v1/models")) {
+    const url = new URL(String(input));
+    state.requests.push(url.pathname + url.search);
+    if (url.pathname.endsWith("/models")) {
       if (state.failModels) throw new Error("connection refused");
       return Response.json({
-        data: state.modelIds.map((id) => ({ id, object: "model" })),
+        data: state.models.map((model) =>
+          state.plainList
+            ? { id: model.id, object: "model" }
+            : {
+                id: model.id,
+                object: "model",
+                ...(model.inputModalities
+                  ? {
+                      architecture: {
+                        input_modalities: model.inputModalities,
+                      },
+                    }
+                  : {}),
+                ...(model.nCtx ? { meta: { n_ctx: model.nCtx } } : {}),
+              },
+        ),
       });
     }
-    if (url.endsWith("/props")) {
+    if (url.pathname.endsWith("/props")) {
       if (state.failProps) throw new Error("props unavailable");
-      return Response.json(state.props ?? {});
+      const modelId = url.searchParams.get("model");
+      // Single-model servers ignore ?model=; router servers address by id.
+      const model = modelId
+        ? state.models.find((entry) => entry.id === modelId)
+        : state.models[0];
+      if (!model?.props) return new Response("not found", { status: 404 });
+      return Response.json({
+        modalities: { vision: model.props.vision ?? false, audio: false },
+        ...(model.props.nCtx
+          ? { default_generation_settings: { n_ctx: model.props.nCtx } }
+          : {}),
+      });
     }
     return new Response("not found", { status: 404 });
   }) as typeof fetch;
 }
 
 describe("createLlamaCppPiProvider", () => {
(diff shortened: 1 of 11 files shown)


cpacker:
Re-review response landed in 48f90642 — all six blocking items:

1. **API-key auth through the runtime**: `Models.getAuth()` now resolves auth for every provider/credential kind; the factory keeps only non-credential connection config + spec-specific env fallbacks (zai's endpoint choice applies after). New test counts exactly one `getAuth` call resolving a stored Anthropic key.
2. **Contract-faithful CredentialStore**: OAuth provider-specific fields (enterpriseUrl/accountId) round-trip untouched; `modify`/`delete` serialized per provider (concurrency test asserts max 1 in-flight callback); `list()` reports pi-ai provider ids; `api_key` writes persist through the provider record (preserving base URL/timeout).
3. **Canonical refresh path**: hand-built refresh context deleted; `refresh`/`refreshAll` go through `Models.refresh()` (credentialed store-backed contexts, pi's own auth gating). The factory catalog branch uses `resolveModel`, so dynamic built-ins refresh on a miss.
4. **radius/auto end-to-end test**: configured gateway (stubbed `/v1/config`) → listed in `/model` → resolves for the turn with the stored key.
5. **llama.cpp upstream port**: native `/models` catalog distinct from `/v1/models` (test fake routes exact paths and asserts `/v1/models` is NOT used in router mode), `status.value === "loaded"` filtering, `meta.n_ctx ?? n_ctx_train`, `maxTokens = contextWindow`, `compat.maxTokensField = "max_tokens"`.
6. **Mod alias bug fixed** (adapter uses `getRegisteredPiProviderLocalNames`; regression test: mod-registered `openai-codex` still reads the `chatgpt-plus-pro` record). The broader mod-branch collapse remains open — LET-10130 is moved back to In Progress and is a blocking follow-up before the parent closes; it is not presented as complete.

Also: listing/settings metadata now reads through the runtime (static catalog only for legacy sync callers without a runtime instance).

`bun run check` 12/12; affected suites green (batch flakes remain the live-local-Ollama contention class, standalone-green).

cpacker:
Round-3 findings addressed in ad5444a6:

1. **Parallel credential path removed**: the factory no longer falls back to its own record/env lookup — `apiKey` comes solely from `Models.getAuth()` (zai's dual-record endpoint selection is the one named exception, applied after). The reviewer's probe is now a regression test: `getAuth` forced to `undefined` with a stored Anthropic record resolves **no** key.
2. **Radius account-change invalidation**: dynamic built-in catalogs invalidate on credential *identity* change (persisted catalog dropped + provider rebuilt, so closure-cached models can't survive). Identity excludes the volatile access token/expiry — routine refreshes keep last-known retention. Test: key-a → key-b against a 401ing gateway yields zero models, not account A's catalog.
3. **llama.cpp exact construction**: fallback context applies before `maxTokens` (128000/128000) and the compat set now literally includes `supportsStore/supportsUsageInStreaming/supportsStrictMode: false`.
4. **`resolveAvailableLocalModelForTurn`** threads its runtime into `localModelSettingsForHandle`.
5. **API-key writes preserve Bedrock config**: `access_key`/`region`/`profile` are re-supplied through the read-modify-write.

On the ⚠️ refresh-scope note: pi-ai 0.81 has no per-provider refresh — `refresh(providerId)` now documents that it refreshes the collection (configured dynamics only) and adds per-provider error semantics on top; happy to revisit if upstream grows a scoped refresh.

`bun run check` 12/12; all affected suites green (remaining batch flake is the live-local-Ollama git-contention class, standalone-green).

cpacker:

PR #3477, opened 2026-07-24
refactor(backend): make pi-ai Models the canonical provider runtime [LET-10125]

cpacker:
## Aggregate staging PR for the LET-10125 stack

This PR consolidates the six-PR LET-10125 stack into a single review/merge unit. It is a **branch consolidation, not a new implementation pass**: the staging branch points at the stack's exact final commit (`ad5444a6`), and its tree is byte-identical to #3474's head (tree `5a04226c`). Nothing was cherry-picked, squashed, rebased, or reconstructed.

## The architectural invariant

> The model used for a turn is the same complete `Model` object published by the provider that discovered it.

On this branch, the local backend's **listing, model resolution, authentication, refresh, and dispatch all flow through a per-backend pi-ai `Models` runtime** (`LocalPiModelsRuntime`):

- `/model` listing, turn execution, and compaction read the same provider-published `Model` instances (asserted **by identity** in tests, for built-ins and discovered local-endpoint models alike).
- `Models.getAuth()` is the sole credential path — a `CredentialStore` adapter over Letta's `auth.json` handles stored API keys and OAuth (refreshes serialized per provider, provider-specific fields like Copilot's `enterpriseUrl` preserved). The one named exception is zai's dual-record endpoint selection.
- Refresh goes through pi-ai's canonical `Models.refresh()` (credentialed store-backed contexts, auth-gated so unconfigured providers are never probed); dynamic catalogs retain last-known models on transient failure, invalidate on endpoint/credential-identity change.
- Stream dispatch runs through the provider that owns the model. No model-name capability heuristic exists anywhere; no Letta code fabricates a `Model` from a string handle.

## What each superseded portion contributed

1. **Runtime foundation + Ollama** (#3464 — LET-10126, LET-10127): instance-scoped `createModels()` runtime; real dynamic Ollama provider with authoritative `POST /api/show` capabilities (fixes the reported qwen3.6:27b silent-vision-loss bug); digest-keyed metadata caching.
2. **Local endpoints** (#3465 — LET-10128, LET-10129): shared `createLocalEndpointPiProvider` skeleton; llama.cpp, LM Studio, and Ollama Cloud as per-engine native-metadata mappers — explicit-unknown (text-only) when metadata is unavailable, never name-guessing.
3. **Mod providers** (#3466 — LET-10130): mod registrations materialize as real pi-ai `Provider`s (static models published, `listModels` → store-backed dynamic refresh, OAuth on `Provider.auth.oauth`); `PiProviderModelRegistration` derived from pi-ai's `Model` type; per-name registry revisions rebuild exactly the re-registered provider.
4. **Compatibility deletion** (#3468 — LET-10131): `customOpenAICompatibleModel()` and every capability heuristic deleted; bare-ID discovery deleted; dispatch-only fallback deleted (unknown providers fail loudly); biome rule bans `@earendil-works/pi-ai/compat` repo-wide with a single documented allowlisted import (`getEnvApiKey`, still compat-only upstream).
5. **pi-ai 0.81.1** (#3472): adopts the 0.81 architecture wave — `fetchModels(context)` with `ModelsStore` persistence, static-baseline/dynamic-overlay catalog merge, provider-owned OAuth (`AuthInteraction` login flows), Pi TUI default-model sync; plus the review-driven gating (no unconfigured probing) and store-invalidation fixes.
6. **Canonical seam** (#3474): built-in lookup through `runtime.getModel()`/`resolveModel()` with no gratuitous cloning; the `CredentialStore` adapter wired into `createModels()` with `Models.getAuth()` replacing all factory credential lookups (no parallel fallback path); llama.cpp ported to the current upstream contract (native `/models` catalog, `status.value === "loaded"` filtering, `n_ctx ?? n_ctx_train`, `maxTokens = contextWindow`, literal compat flag set); radius end-to-end; account-change catalog invalidation.

## Linear

Parent: LET-10125. Children: LET-10126, LET-10127, LET-10128, LET-10129, LET-10130, LET-10131. All remain open until this staging PR passes aggregate review and merges. LET-10130's broader mod-branch consolidation is explicitly **In Progress** and tracked as a follow-up; the parent should not close before it lands.

## Superseded stacked PRs (closed unmerged; exact combined tree preserved here)

- #3464 `charles/let-10126-pi-models-runtime-ollama`
- #3465 `charles/let-10128-9-local-endpoint-providers`
- #3466 `charles/let-10130-mod-providers-runtime`
- #3468 `charles/let-10131-compat-cleanup`
- #3472 `charles/pi-ai-0-81-upgrade`
- #3474 `charles/canonical-runtime-seam`

## Aggregate stats vs main

40 files changed, +4216 −1080 (source +2328 −868, tests +1846 −176). `bun run check` 12/12 on the tip; the branch carries the full regression suite from the stack (same-instance invariants, payload retention, refresh gating, credential-store contract, llama.cpp router mode, radius e2e).


cpacker:
CI on the initial staging tip (`ad5444a6`) surfaced two real aggregate defects — exactly the case where implementation changes were sanctioned — fixed in `a6b38209` (the staging head is no longer byte-identical to the closed #3474's original tip; the #3474 branch has been fast-forwarded to match):

1. **Google env auth regression** (Headless Local / gemini failure, "Provider is not configured: google"): the previous parallel-credential removal also dropped Letta's ambient env fallback, but upstream's google provider only reads `GEMINI_API_KEY` while Letta documents `GOOGLE_GENERATIVE_AI_API_KEY`. Restored as a *named* exception scoped to ambient env vars upstream doesn't read (plus the keyless local-daemon placeholder) — the stored-record lookup stays removed, and the masking regression test now clears ambient env so it keeps proving that property.
2. **Stale ProviderSelector expectation** (Linux/macOS unit failures): pi-ai 0.81.1 legitimately ships subscription OAuth for **radius** and **xai**, so the `/connect` subscription filter now lists five providers; expectation updated.

`bun run check` 12/12 locally on `a6b38209`; awaiting the fresh CI run.

amelia-letta (changes requested):
## Review

This is a massive, well-structured refactoring that makes pi-ai's `Models` runtime the single source of truth for model discovery, auth, and dispatch. The architectural invariant ("the model used for a turn is the same complete `Model` object published by the provider that discovered it") is cleanly enforced, and the biome lint rule banning `@earendil-works/pi-ai/compat` is a good guardrail.

Two CI failures need to be addressed before merge:

---

### 1. ProviderSelector test failure (macOS, Linux x64, Linux arm64)

`src/cli/components/provider-selector.test.ts:177` — the "subscription" filter test:

```
Expected:  ["anthropic-oauth", "openai-codex-oauth", "github-copilot"]
Received:  ["anthropic-oauth", "github-copilot", "openai-codex-oauth", "radius", "xai"]
```

pi-ai 0.81.1 added `radius` and `xai` as OAuth providers. `localOAuthProviderConfigs()` gives every OAuth provider `description: "Connect a subscription account"`, so they now match the "subscription" filter. The sort order also changed (display names shifted). The test needs to be updated to reflect the new provider set.

### 2. Headless Local / gemini-3.1-flash-lite failure

```
"Provider is not configured: google"
```

The CI workflow sets `GOOGLE_GENERATIVE_AI_API_KEY`, but pi-ai's Google provider only checks `GEMINI_API_KEY` (`envApiKeyAuth("Gemini API key", ["GEMINI_API_KEY"])` in `providers/google.js`).

The old code had a custom `apiKeyEnv` that checked both:
```typescript
apiKeyEnv: () => process.env.GOOGLE_GENERATIVE_AI_API_KEY ?? getEnvApiKey("google"),
```

The new code still calls `localProviderConnection(spec.localProviderNames, spec.apiKeyEnv?.() ?? spec.fallbackApiKey, storageDir)` which finds the key from `GOOGLE_GENERATIVE_AI_API_KEY`. But then it overwrites the result:
```typescript
const authResult = await modelsRuntime.getAuth(authProviderId);
connection = { ...connection, apiKey: authResult?.auth.apiKey };
```

When `authResult` is `undefined` (pi-ai can't find `GEMINI_API_KEY`), `connection.apiKey` is set to `undefined`, discarding the key from `GOOGLE_GENERATIVE_AI_API_KEY`. The key is lost and the turn fails with "Provider is not configured: google."

The comment says "No generic fallback to the factory's own record/env lookup: a runtime auth regression must surface, not be silently masked." That's the right principle for masking stale credentials, but here the key *is* available — just under a Letta-specific env var alias that pi-ai doesn't know about.

**Fix options:**
- Pass a custom `authContext` to `createModels()` that checks `GOOGLE_GENERATIVE_AI_API_KEY` as a fallback for the `google` provider
- Update the CI workflow to set `GEMINI_API_KEY` (but this would break existing users who set `GOOGLE_GENERATIVE_AI_API_KEY`)
- Don't override `connection.apiKey` when `authResult` is `undefined` (preserves the `localProviderConnection` key)

---

### Non-blocking observations

- **`as never` type casts** in `pi-stream-adapter.ts` and `compaction.ts` for bedrock stream options — minor type safety gap, not a runtime issue
- **`PiProviderInputType`** in `pi-provider-mod-types.ts` may be dead code — still exported but no longer used in `PiProviderModelRegistration`
- Image elision, mod provider adapter, credential store adapter, and OAuth migration all look correct
- Test coverage is comprehensive

src/backend/dev/pi-model-factory.ts
```diff
@@ -557,6 +476,41 @@ export async function resolvePiModelForAgent(
...
+      ? registeredProvider.providerName
+      : (spec?.piProvider ?? provider);
+    const authResult = await modelsRuntime.getAuth(authProviderId);
+    // No fallback to the factory's own stored-record lookup: a runtime auth
+    // regression must surface, not be silently masked by a parallel
+    // credential path. Two named exceptions remain: (1) Letta-documented
+    // ambient env vars that upstream providers do not read (e.g.
+    // GOOGLE_GENERATIVE_AI_API_KEY — upstream google only reads
+    // GEMINI_API_KEY) plus the keyless local-daemon placeholder, and
+    // (2) zai's dual-record endpoint selection below.
```
amelia-letta: When `authResult` is `undefined` (pi-ai cannot find `GEMINI_API_KEY`), this overwrites the key from `localProviderConnection` with `undefined`. The CI sets `GOOGLE_GENERATIVE_AI_API_KEY` which the old `apiKeyEnv` checked, but pi-ai only checks `GEMINI_API_KEY`. This causes the "Provider is not configured: google" failure in Headless Local / gemini-3.1-flash-lite.

cpacker:
@amelia-letta Both blocking findings were fixed in a6b38209 (pushed while this review ran against the initial tip ad5444a6):

1. **ProviderSelector subscription filter** — expectation updated to the five-provider set (radius and xai legitimately ship subscription OAuth in pi-ai 0.81.1).
2. **Google env alias** — restored as a *named* ambient-env exception: `apiKey: authResult?.auth.apiKey ?? spec?.apiKeyEnv?.() ?? spec?.fallbackApiKey`. This is deliberately scoped to ambient env vars upstream providers don't read (`GOOGLE_GENERATIVE_AI_API_KEY` vs upstream's `GEMINI_API_KEY`) plus the keyless local-daemon placeholder — the stored-record lookup stays removed so a runtime credential regression still surfaces (regression-tested with ambient env cleared). Of your fix options this is closest to option 3, but narrowed so stored credentials can never ride the fallback; a custom `authContext` mapping is a nice future refinement once upstream documents alias env vars.

Non-blocking notes (`as never` bedrock casts, possibly-dead `PiProviderInputType` export) are logged for the LET-10130 follow-up rather than churning this consolidation PR.

amelia-letta (approved):
## Update

I initially reviewed an earlier commit of this PR and flagged two CI failures. Both have been addressed in the latest push (a6b3820):

1. **ProviderSelector test** — updated to expect `["anthropic-oauth", "github-copilot", "openai-codex-oauth", "radius", "xai"]` with a comment noting pi-ai 0.81 ships subscription OAuth for radius and xai. Fixed.

2. **Headless Local / gemini-3.1-flash-lite** — the auth resolution now falls back to `spec?.apiKeyEnv?.()` when `authResult` is undefined:
   ```typescript
   apiKey: authResult?.auth.apiKey ?? spec?.apiKeyEnv?.() ?? spec?.fallbackApiKey,
   ```
   This preserves the `GOOGLE_GENERATIVE_AI_API_KEY` path that pi-ai's default `GEMINI_API_KEY` check misses. The comment documents the exception clearly. Fixed.

The non-blocking observations from my original review still stand (`as never` casts, `PiProviderInputType` potentially dead code) but neither is merge-blocking.

The refactoring is solid — the architectural invariant is cleanly enforced, the credential store adapter is correct, and the test coverage is comprehensive. Approving.

cpacker:
Re-review response, landed in 2560f557:

The reviewer's core catch was correct and important: my previous Google fix used `spec.apiKeyEnv?.()`, which exists **generically** for every built-in provider (`makePiProviderSpec` defaults it to `getEnvApiKey(provider)`), so it silently reintroduced a factory-side ambient credential path for all providers — capable of masking a `Models.getAuth()` regression whenever an env key exists.

Fixed at the runtime/provider boundary (the reviewer's option 2): `createModels()` now receives an `AuthContext` that resolves `GEMINI_API_KEY` through Letta's documented `GOOGLE_GENERATIVE_AI_API_KEY` alias — an exact, documented, single-entry alias table, resolved inside pi-ai's own auth resolution. The factory is back to `apiKey: authResult?.auth.apiKey` with **no credential fallback of any kind**: the keyless local-daemon placeholder already resolves through each endpoint provider's `auth.resolve()` (the `fallbackApiKey` leg was dead weight), and zai's dual-record endpoint selection remains the one named exception. Regression test: `GOOGLE_GENERATIVE_AI_API_KEY` set without `GEMINI_API_KEY` resolves via `runtime.getAuth("google")`.

Also correcting the record on LET-10130 framing: the mod-provider materialization (former #3466) ships **in this staging tree**; only the resolution-branch consolidation remains outstanding. The ticket stays In Progress until this PR merges.

`bun run check` 12/12 locally on 2560f557.

cpacker:
Merge-readiness package landed in 54f28906, per the settled bar (all 8 requirements):

1. **Effective-value identity gating** across all provider classes — persisted selection settings that restate published values never clone; asserted through the real selection path (`localModelSettingsForHandle`-derived settings) for mods and endpoints.
2. **Credential-change invalidation as one consistent operation**: new `resolveTurn()` returns model + auth from one provider state with identity invalidation applied on every async resolution path; lookup-path regression test proves new auth never pairs with a stale account's cached model.
3. **Built-in restoration** when a mod overriding its id unloads (regression test: `openai-codex` catalog survives).
4. **Mod refresh consumes `RefreshModelsContext.credential`** — no parallel auth read in discovery; `authHeader` moved into provider-owned auth.
5. **Dead factory auth computation deleted** — `localProviderConnection` is non-credential config only; `withAuthHeader` gone.
6. **Duplicate `lastKnown` cache removed** — retention comes solely from the pi-ai `ModelsStore`; the digest cache is now a typed, provider-instance-scoped `metadataFingerprints` mechanism (fingerprints only, invalidated with the provider connection).
7. **LET-10130 collapse complete**: the factory has a single `resolveTurn` resolution path (no registered/managed/catalog model branches — remaining named items: zai endpoint selection, the mod `modifyModels` product hook, one base-URL override-policy line); `local-model-config` lists every provider class through one runtime discovery projection and derives settings solely from runtime-published models.
8. **Scoped refresh**: explicit documented exception with honest semantics in code + test ("collection-wide across materialized dynamic providers"), upstream ask filed as earendil-works/pi#7040 (also covers the env-api-keys export and a store field for metadata fingerprints, i.e. the paths to retiring the last two documented exceptions).

`bun run check` 12/12; 289 tests across the affected surface (remaining batch flakes are the established live-local-daemon contention class, standalone-green).

cpacker:
**Re: dynamic OAuth mod isolation review — all three findings addressed in `66b31975`.**

**P1 (confirmed, fixed): OAuth account switch on a dynamic mod could pair new auth with the old account's catalog.** The diagnosis was exact: invalidation iterated only dynamic built-ins, and the mod reconstruction signature (`revision / baseURL / record apiKey`) is blind to OAuth credentials (`apiKey` is `undefined` for OAuth records). Credential-identity invalidation now covers every account-scoped catalog — `accountScopedProviderIds()` = dynamic built-ins ∪ mod providers with a `listModels` hook (a static mod catalog overriding a dynamic built-in id correctly leaves the set). On an identity change the runtime drops the persisted catalog and rebuilds the provider — `createModPiProvider` for mods, the fresh built-in otherwise. The regression test (`pi-mod-provider.test.ts`, "OAuth account switch on a dynamic mod drops the old account's catalog") reproduces the review's sequence: account A publishes `account-a-model`, refresh, switch stored OAuth to account B, `resolveTurn` without an explicit refresh → asserts `auth.apiKey === "access-b"` **and** `model === undefined` with an empty catalog. Same identity semantics as before: volatile `access`/`expires` are excluded, so routine token rotation never clears last-known retention.

**P2 (fixed): `resolveRegisteredPiProviderListModelsConnection` deleted.** It was the last remnant of the parallel credential read (superseded by `context.credential` in the mod provider's `fetchModels`); it and its now-unused `getLocalOAuthApiKey`/`PiProviderConnection` imports are gone.

**Tracking (fixed): earendil-works/pi#7040 was bot-closed as not-planned**, so it is no longer the expiry path. The three upstream-dependent exceptions (scoped refresh, `env-api-keys` outside `/compat`, store-level fingerprint field) are now tracked in [LET-10144](https://linear.app/letta/issue/LET-10144), owned by us: adopt-and-delete when upstream ships, re-raise upstream when there's an opening.

`bun run check` 12/12; `src/backend/dev` + `src/backend/local` suites green (69 tests). Both `charles/canonical-runtime-seam` and this staging branch point at the same commit.

cpacker:
**Re: credential race during resolution — confirmed and fixed in `3e073485`.**

The finding was right: `resolveTurn` validated the catalog against the invalidation-time credential read, but `Models.getAuth` performs its own independent store read, and logins are not serialized against resolution — so a switch landing between the two reads paired B's auth with A's catalog.

**Fix (detect-and-retry):** pi-ai's provider-owned auth resolution reads the credential store itself, so injecting one shared snapshot isn't possible without bypassing the provider auth seam. Instead, after auth *and* the model lookup/refresh complete, `resolveTurn` re-reads the stored identity and compares it to the invalidation snapshot the catalog was validated against; if it moved mid-flight, it invalidates and retries (bounded at 2 retries). The final identity check runs after the last await before return, which closes both orderings — a switch during auth resolution (the reproduced case) and a switch during the on-miss catalog refresh (the symmetric case, where a fresh catalog could have paired with stale auth). Only account-scoped providers carry a recorded identity, so static-catalog providers pay no extra read per turn.

**Regression test** (`pi-mod-provider.test.ts`, "account switch during turn resolution cannot pair the old catalog with new auth"): wraps the runtime's credential store so the file-backed switch to account B completes immediately after the invalidation snapshot reads account A — auth then resolves as B while the catalog was validated as A. Asserts `auth.apiKey === "access-b"` and `model === undefined`. Verified the test reproduces the exact leak (`account-a-model` + `access-b`) with the retry disabled.

Residual window: after the final identity re-read there are no further awaits before return, so what remains is the irreducible tail of any read-then-use without serializing logins against turns — the returned pair is always from one verified identity at resolution time.

`bun run check` 12/12; backend dev/local suites 70/70. Both branches at `3e073485`.

cpacker:
**Re: fail-open retry exhaustion — confirmed and fixed in `de43ccb6`.**

The off-by-one was real: `attempt < MAX && await moved(...)` short-circuited, so the final attempt performed no identity re-read at all — contradicting my previous claim that the residual window began after the final check. Exhaustion now fails closed, per the suggested shape:

1. **The identity check always runs**, on every attempt including the last.
2. **If the identity moved and the retry budget is exhausted, `resolveTurn` throws** a retryable error (`Credentials for provider "…" changed repeatedly during turn resolution; retry the request.`) instead of returning a pair whose auth and catalog may come from different accounts. The error propagates through the factory to the turn, which surfaces it as a failed, retryable request.
3. **Both regressions added** (`pi-mod-provider.test.ts`):
   - *"retry exhaustion fails closed instead of returning a mismatched pair"* — the wrapped credential store lands a different account after every read (your three-rapid-changes reproduction, generalized to permanent flapping), so no attempt observes a stable identity; asserts rejection. Verified it fails under the previous short-circuit condition.
   - *"switch during the on-miss catalog refresh resolves a consistent pair"* — the explicit switch-during-refresh ordering: the login to B completes while the on-miss refresh runs with A's auth (fresh catalog racing already-resolved auth); asserts resolution converges to a consistent `access-b`/`catalog-b` pair.

The residual-window claim is now accurate: the identity re-read is unconditionally the last await before return, and a mismatch at that point can only throw or retry — never return.

`bun run check` 12/12; backend dev/local suites 72/72. Both branches at `de43ccb6`.

PR #3481, opened 2026-07-24
feat(cron): bind channel targets before scheduled turns

overlord-letta:
## Summary
- persist optional channel destinations on local schedules and validate them on the listener protocol
- for each fire, validate that selected channel runtimes are active, create the requested conversation, then atomically move all selected root routes before the prompt can run
- preserve legacy schedules without channel targets and support replacing or clearing target selections

## Behavior
A schedule can target multiple channel destinations. `new` schedules create a fresh conversation on each fire and move those routes to it before the queued turn starts, so the turn receives the normal route-scoped `MessageChannel` tool. If an adapter is stopped, a route is outbound-disabled, the queue is full, or any route write fails, the prompt does not run and prior queue/route/account state is restored.

Thread-specific destinations are intentionally not supported; schedule targets bind root chats/channels.

## Validation
- `bun run typecheck`
- `bun test src/channels/service-route-bindings.test.ts src/cron/channel-targets.test.ts src/cron/cron-file.test.ts src/websocket/listener/protocol-inbound-cron.test.ts src/websocket/listen-client-protocol.test.ts` (206 passed)
- `bun run lint` (clean; 6 pre-existing warnings)
- `bun run check:cycles`
- `bun run check:boundaries`
- `bun run check:module-ownership`
- `bun run check:file-size`
- `bun run check:test-mock-isolation`

### DIFF
--- a/scripts/source-file-size-baseline.json
+++ b/scripts/source-file-size-baseline.json
@@ -47,6 +47,6 @@
   "src/websocket/listener/commands/memory.ts": 1114,
   "src/websocket/listener/file-commands.ts": 1030,
   "src/websocket/listener/lifecycle.ts": 1779,
-  "src/websocket/listener/protocol-inbound.ts": 2290,
+  "src/websocket/listener/protocol-inbound.ts": 2137,
   "src/websocket/listener/protocol-outbound.ts": 1128
 }
--- a/src/channels/service-accounts.ts
+++ b/src/channels/service-accounts.ts
@@ -14,16 +14,13 @@ import {
   createAccountFromPatch,
   mergeAccountPatch,
 } from "./service-account-model";
-import {
-  assertSupportedChannelId,
-  getErrorMessage,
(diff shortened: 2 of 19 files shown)


just-cameron (approved):
Rebased onto current main and revalidated locally. Typecheck, the cron/channel/protocol tests (189 passing), lint, cycle, layer-boundary, module-ownership, source-file-size, and mock-isolation checks all pass. The two conflicts were mechanical: preserve the new management protocol imports while extracting cron validators, then recalculate the file-size baseline.

just-cameron:
@cpacker @4shub This is fully rebased onto current main, locally validated across the focused cron/channel suite plus typecheck/lint/architecture gates, and all GitHub checks are green. I approved after reviewing the route-binding, persistence, protocol, rollback, and reconnect paths. Merge is now blocked only on the required CODEOWNERS approval; could one of you take the final look?

PR #3489, opened 2026-07-25
fix: detect max_tokens_exceeded as context window overflow

ninachaubal:
## Summary

- Add `"max_tokens_exceeded"` to the overflow error pattern list in `isContextWindowOverflowError`
- This is a standard OpenAI error code used by multiple providers (OpenRouter, AIsa, Umans, and OpenAI itself) but was not recognized by Letta's overflow detection
- Without this, reactive compaction doesn't trigger when a long conversation hits the context limit on providers that return this error code, leaving the conversation in a broken state until a new conversation is started

## Context

Discovered while debugging a `max_tokens_exceeded` error from the [Umans](https://umans.ai) code gateway (an OpenAI-compatible provider registered via a [community provider mod](https://github.com/ninachaubal/letta-provider-umans)). Long-running conversations would hit the context window limit, the provider would return `max_tokens_exceeded`, but Letta's reactive compaction path didn't recognize it as an overflow error — so the conversation was stuck in a broken state.

Research shows `max_tokens_exceeded` is widely used:
- **OpenRouter** – canonical `error_type` in their [API error reference](https://openrouter.ai/docs/api_reference/errors-and-debugging)
- **AIsa** – listed as a common error code under 422 Unprocessable Entity
- **Omnigent** – included in their `CONTEXT_EXCEEDED_CODES` frozenset alongside `context_length_exceeded`
- **OpenAI** – appears in their own error handling examples (September 2024 API updates)

## Test plan

- [ ] Verify existing overflow detection tests still pass
- [ ] Verify an error containing `max_tokens_exceeded` is now classified as context window overflow

### DIFF
--- a/src/backend/dev/context-window-overflow.ts
+++ b/src/backend/dev/context-window-overflow.ts
@@ -54,5 +54,6 @@ export function isContextWindowOverflowError(error: unknown): boolean {
     "request too large",
     "request_too_large",
     "model_context_window_exceeded",
+    "max_tokens_exceeded",
   ].some((marker) => haystack.includes(marker));
 }


cpacker:
Closing this because the reported failure may be real, but this patch targets the wrong layer and does not fix the normal local-provider path.

### Additional context that can be copy-pasted into another coding agent

Please investigate this as a provider-integration issue rather than adding `max_tokens_exceeded` to Letta Code’s global context-overflow fallback:

- Normal local provider failures are wrapped as `PiProviderError`. `pi-stream-adapter.ts` then calls pi-ai’s `isContextOverflow(error.assistant, contextWindow)` and bypasses the `isContextWindowOverflowError()` function changed by this PR. Both pinned pi-ai 0.81.1 and current upstream pi-ai return false for an assistant error containing only `max_tokens_exceeded`.
- `max_tokens_exceeded` is not generically equivalent to context overflow. OpenRouter’s typed-error reference distinguishes it from `context_length_exceeded`: the former means generation reached `max_tokens`/`max_completion_tokens`; the latter means input plus output exceeded the model context window. Globally treating the former as context overflow can compact and discard healthy conversation context for an output-limit failure.
- The official OSS Umans integration, `umans-ai/pi-provider-umans`, uses pi-ai’s `anthropic-messages` stack and maps `recommended_max_tokens` to the model output budget, with `max_completion_tokens - 1` only as a ceiling. For Kimi/Coder this yields 32,768 output tokens, not 262,143.
- The community Letta Umans mod currently diverges from that official integration: it uses `openai-completions` and sets `maxTokens` to `max_completion_tokens - 1`. That can request nearly the entire context window as output, so a growing prompt can make the prompt-plus-output budget invalid. Its history also shows `supportsUsageInStreaming` was only enabled immediately before this PR, which should restore proactive compaction from real usage.

Recommended next step: align the Umans provider mod with the official Umans pi provider, use `recommended_max_tokens`, and add a provider-level regression test. Capture one sanitized raw Umans failure—including status, complete error body, input/output usage, requested output budget, and whether output had started—before deciding whether any provider-specific error normalization is needed. If a generic pi-ai gap remains after that, fix it upstream in pi-ai with positive and negative tests rather than maintaining a second global classifier in Letta Code.

PR #3496, opened 2026-07-25
fix(local): show on-demand llama.cpp router models

### LINKED ISSUE #3493: fix(local): show unloaded on-demand models from llama.cpp router mode
## Summary

Letta Code 0.29.1 hides valid llama.cpp router-mode models from `/model` when the router reports them as `status.value: "unloaded"`. In router mode these presets are available and load automatically when selected/used; `unloaded` does not mean unavailable.

## Environment

- Letta Code: 0.29.1, built from git
- Backend: Local
- llama.cpp router: `127.0.0.1:8088`
- Explicit provider URL: `http://127.0.0.1:8088/v1`

## Reproduction

1. Start llama.cpp in router mode with several on-demand presets.
2. Connect Letta:
   
-----

model showed after running
MODEL='Qwen3.6-27B'
curl -sS http://127.0.0.1:8088/v1/chat/completions \
  -H 'Content-Type: application/json' \
  -d "{\"model\":\"$MODEL\",\"messages\":[{\"role\":\"user\",\"content\":\"Reply OK\"}],\"max_tokens\":1}"

cpacker:
## Summary

- show healthy unloaded llama.cpp router presets in the model picker so the router can load them on demand
- keep sleeping models selectable while continuing to hide loading, downloading, and failed entries
- preserve the provider-ownership invariant: listing and turn execution consume the same complete Model instance built from native engine metadata
- add regression coverage for every router status and preserve the existing single-model behavior

Fixes #3493.

### DIFF
--- a/src/backend/dev/pi-llama-cpp-provider.test.ts
+++ b/src/backend/dev/pi-llama-cpp-provider.test.ts
@@ -5,6 +5,7 @@ import { createLlamaCppPiProvider } from "./pi-llama-cpp-provider";
 interface FakeLlamaCppModelEntry {
   id: string;
   status?: string;
+  failed?: boolean;
   inputModalities?: string[];
   nCtx?: number;
   nCtxTrain?: number;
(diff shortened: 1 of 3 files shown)


edt-xx:
Thanks.  It now shows the unloaded models.

PR #3515, opened 2026-07-27
feat(app-server): dead-peer detection and same-credential preemption for control/stream channels

cpacker:
## Problem

The app-server accepts exactly one control-channel and one stream-channel client and rejects every extra connection with close code 1008 (`control channel already connected` / `stream channel already connected`). While building the Letta Agent SDK reference react app, three failure modes were reproduced live against this behavior:

1. **Zombie holder**: a client that dies without a close frame (crashed app, dropped mobile socket, killed proxy) keeps the slot until the dead-peer watchdog reaps it — up to 90s with the default cadence, during which every reconnect gets 1008.
2. **Close/connect race**: a closing client races its successor's connect. The close teardown has not freed `activeSession` when the successor arrives, so legitimate quick reconnects (e.g. navigation in a mobile app) get 1008.
3. **Second-tool deadlock**: any second tool in the same workflow (SDK management transport plus a session) deadlocks on the occupied slot.

## Design

**Dead-peer detection** already exists on `main` (#3143): the server pings each connected socket every 30s at the WebSocket protocol level and terminates any client without a pong for 90s, with tests. This PR does not change that cadence; it demotes the watchdog to a backstop and updates its comment accordingly.

**Newest-wins preemption** (this PR): when a new connection arrives for an occupied slot, the server closes the existing holder with **`4000` / `superseded by newer connection`** and seats the newcomer, instead of rejecting the newcomer with 1008. This matches how users actually behave — the newest client is the live one — and resolves all three failure modes without waiting on any timeout. Details:

- The 4000 close code (private-use range) lets the displaced client distinguish "another client took over — do not auto-reconnect" from transport failures (1006) and policy rejections (1008).
- A superseding **stream** connection tears down the whole paired session and is seated as pending until its own control channel arrives (5s window, pre-existing). Swapping only the stream socket under a live session would cross-pair one client's control with another's stream during two-channel reconnects.
- A newer pending stream socket likewise replaces an older pending one (previously 1008 `stream channel already pending`).
- Every preemption is logged via `onLog`.
- `onSessionClosed` now only clears `activeSession` when it still belongs to the closing connection, so the superseded session's deferred close events cannot clobber its successor's slot.

**Why preemption is unconditionally safe here (same-credential reasoning):** the auth policy is fixed for the server's lifetime and enforced in the `upgrade` handler before any connection reaches channel seating — `authorizeUpgrade` for capability-token / signed-bearer-token modes, plus the post-#3511 rule that unauthenticated servers only admit loopback clients without an Origin header. Any connection that reaches seating is therefore authenticated exactly as strongly as the current holder (or both are unauthenticated loopback). There is no accepted-but-weaker-credential case to preserve a rejection for; a code comment documents that if per-connection identities are ever introduced, preemption must start comparing credential strength.

## Behavior changes for existing clients (please review)

- **Desktop listener / SDK / any app-server client**: a second client connecting to a busy server now *displaces* the first (which receives close 4000) instead of failing fast with 1008. Two clients configured to auto-reconnect at the same server could now flap indefinitely where previously the second failed fast — clients should treat close code 4000 as "do not auto-reconnect". No in-repo code matches on the removed 1008 reasons (the 1008 handling in `listener/lifecycle.ts` is the cloud-relay path, untouched).
- The pending-stream timeout rejection (1008 `control channel did not connect`) and all upgrade-time auth rejections are unchanged.

## Tests

New tests in `src/websocket/app-server.test.ts`, following the file's existing patterns:

- quick reconnect: second control connection succeeds; first receives exactly `4000` / `superseded by newer connection`, preemption is logged
- zombie holder (client with `autoPong: false`, watchdog timeout far beyond the test window): successor connects immediately without waiting for the watchdog
- newer stream connection supersedes the paired session (both old sockets get 4000) and pairs with the newcomer's control channel
- newer pending stream socket replaces an older pending one

Verification (`HOME` isolated to a scratch dir to avoid local `~/.letta` mods):

- `bun test src/websocket/app-server.test.ts` — 16/16 pass, stable across 3 runs
- `bun test src/websocket/` — 654/655 pass; the 1 failure is pre-existing flakiness in `listen-model-update.test.ts` (5s timeouts, also fails on clean `main`, varying tests per run)
- `bun run typecheck`, `bun run lint` (19 warnings, identical on clean `main`), layer-boundary / file-size / exported-functions / test-coverage checks — all clean

### DIFF
--- a/src/websocket/app-server.test.ts
+++ b/src/websocket/app-server.test.ts
@@ -140,6 +140,26 @@ function waitForClientPing(socket: WebSocket): Promise<void> {
   });
 }
 
+function waitForCloseEvent(
+  socket: WebSocket,
+): Promise<{ code: number; reason: string }> {
+  return new Promise((resolve, reject) => {
+    const timeout = setTimeout(() => {
+      cleanup();
+      reject(new Error("Timed out waiting for websocket close event"));
+    }, TEST_TIMEOUT_MS);
+    const cleanup = () => {
+      clearTimeout(timeout);
+      socket.off("close", handleClose);
+    };
+    const handleClose = (code: number, reason: Buffer) => {
+      cleanup();
+      resolve({ code, reason: reason.toString() });
+    };
+    socket.once("close", handleClose);
+  });
+}
+
 function waitForClientClose(socket: WebSocket): Promise<void> {
   if (
     socket.readyState === WebSocket.CLOSED ||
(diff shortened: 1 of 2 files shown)


cpacker:
Closing in favor of the connection-identity / one-bidirectional-WebSocket design tracked in [LET-10235](https://linear.app/letta/issue/LET-10235/refactor-app-server-listener-to-support-one-bidirectional-ws-per). The reconnect repros here are useful, but unconditional newest-wins preemption conflates authentication with ownership and cannot safely pair the current control/stream sockets during concurrent reconnects. We are landing the narrower SDK connection fixes first and punting the server transport refactor to LET-10235.

PR #3516, opened 2026-07-27
fix(local): compact before output headroom is exhausted

### LINKED ISSUE #3508: Local llama.cpp responses persist zero usage, preventing reliable auto-compaction and surfacing near-limit `length` as `max_tokens_exceeded`
## Summary

Using a Local agent with llama.cpp, automatic compaction does not trigger reliably as the conversation approaches/exceeds the configured Letta context limit. The llama.cpp provider's normalized assistant messages persist all-zero usage, so Letta cannot compare the provider's real context usage against `context_window_limit` and instead uses a rough transcript estimate.

A provider `finish_reason: "length"` is then surfaced as `max_tokens_exceeded`, even when llama.cpp reports no context truncation/error. The affected conversation also contains an invalid `model_settings.max_tokens: 128000` override despite the llama.cpp server context being 110,240.

Discord investigation thread (requires access to the Letta Discord):
https://discord.com/channels/1161736243340640419/1530618961572397206

## Environment

- Letta Code: 0.29.2, built from a fresh git checkout (please replace with exact SHA: `git rev-parse --short HEAD`)
- Backend: Local
- Provider: llama.cpp
- Model: `llama.cpp/Qwen3.6-27B`
- llama.cpp context: 110,240 in the latest test
- Letta conversation context limit: 100,000 in the latest test
- Earlier reproduced configurations: Letta limits of 95,000 and 90,000 against a 100,000 llama.cpp context
- Host OS/distribution: unknown; please add
- llama.cpp version/commit: unknown; please add if available

## Expected behavior

Letta should compact before dispatching a turn that cannot safely fit within the configured context limit plus output headroom, or compact and retry after a near-limit provider `length` finish.

Provider usage should be captured accurately enough for automatic compaction. A generic `max_tokens_exceeded` message should not imply context overflow when the underlying provider only returned `finish_reason: "length"`.

Model changes/context-limit updates should not leave an impossible `max_tokens` override in conversation settings.

## Actual behavior

1. Automatic compaction did not trigger.
2. Letta displayed:

```text
An error occurred during agent execution
(run_id: local-run-34, stop_reason: max_tokens_exceeded)
```

3. llama.cpp reported fewer than 60,000 tokens for that turn and no provider/context error.
4. `/context` in Letta reported a divergent fallback estimate:

```text
96k/100k tokens (96%) Â· Qwen3.6-27B
```

5. The latest normalized assistant transcript row contained:

```json
{
  "stopReason": "length",
  "usage": {
    "input": 0,
    "output": 0,
    "cacheRead": 0,
    "cacheWrite": 0,
    "totalTokens": 0
  },
  "diagnostics": null,
  "provider": "llama-cpp",
  "model": "Qwen3.6-27B",
  "contentTypes": ["thinking"]
}
```

The failed thinking block was only 570 characters. Multiple preceding successful llama.cpp rows also persisted all-zero usage.

6. The conversation record contained:

```json
{
  "id": "local-conv-22",
  "model": "llama.cpp/Qwen3.6-27B",
  "context_window_limit": 100000,
  "model_settings": {
    "provider_type": "openai",
    "context_window_limit": 100000,
    "max_tokens": 128000,
    "parallel_tool_calls": true,
    "reasoning": {
      "reasoning_effort": "medium"
    }
  }
}
```

`max_tokens: 128000` is invalid against the llama.cpp server's 110,240-token context and does not include the configured 10,240-token thinking allowance; whether that allowance should also be the total generation cap needs confirmation.

## Earlier near-limit evidence

With a 95,000 Letta limit and 100,000 llama.cpp context, llama.cpp logged:

```text
prompt eval time = 954.86 ms / 124 tokens
eval time = 0.00 ms / 1 tokens
total time = 954.86 ms / 125 tokens
stop processing: n_tokens = 94763, truncated = 0
```

Letta returned `max_tokens_exceeded` and did not compact. Lowering the intended Letta limit to 90,000 did not restore automatic compaction. That 90k turn's exact effective runtime limit and `/context` snapshot were not captured, so this report does not claim a specific cause for that individual test.

## Reproduction outline

1. Run a Local agent using a llama.cpp OpenAI-compatible endpoint.
2. Set a Letta conversation context limit below llama.cpp's context (for example, Letta 100,000 and llama.cpp 110,240).
3. Continue a conversation until Letta `/context` approaches the configured limit.
4. Submit another turn.
5. Observe that automatic compaction does not run before dispatch.
6. Observe a `max_tokens_exceeded` error when the normalized provider stop reason is `length`.
7. Inspect the Local transcript under:

```text
~/.letta/lc-local-backend/conversations/<encoded-conversation>/messages.jsonl
```

8. Observe all-zero normalized usage on llama.cpp assistant rows.
9. Inspect `conversation.json` and observe the stale/invalid `max_tokens` value if present.

### Streaming usage is disabled for llama.cpp

`src/backend/dev/pi-llama-cpp-provider.ts` sets:

```ts
supportsUsageInStreaming: false
```

In pi-ai's OpenAI-completions adapter, this causes omission of:

```json
{"stream_options":{"include_usage":true}}
```

When llama.cpp does not send an unsolicited usage chunk, normalized usage remains zero.

### Automatic compaction uses a strict post-turn threshold

`src/backend/local/local-backend.ts` (`compactAfterContextUsage`) compacts only when:

```ts
contextTokens > contextWindow
```

When usage is zero, it falls back to `estimateProviderContextTokens(input)`, which estimates the pre-response input rather than using llama.cpp's resident token count.

### `length` is shown as `max_tokens_exceeded`

`src/backend/dev/provider-turn-executor.ts` maps normalized provider `length` to:

```text
max_tokens_exceeded
```

That label is ambiguous and is not proof that the provider context window was exceeded.

### Persisted `max_tokens` overrides discovered model metadata

`src/backend/dev/pi-model-factory.ts` applies persisted `model_settings.max_tokens` as a model max-output override. In this case, 128,000 exceeds the provider's 110,240 total context.

## Requested fixes

1. Capture accurate streamed usage from current llama.cpp versions, or provide a reliable provider-specific context accounting fallback.
2. Add preflight compaction based on estimated prompt size plus required output headroom, rather than waiting for strict post-turn `contextTokens > contextWindow`.
3. Consider compact-and-retry for a near-limit `length` finish when safe.
4. Prevent stale/impossible `max_tokens` values from surviving model/provider/context changes.
5. Improve the error text so `length` is not presented as definitive context/max-token exhaustion without usage evidence.

## Workaround

Manual `/compact` is the only workaround demonstrated in this investigation. Lowering the configured limit to 90,000 was tested and did not reliably restore automatic compaction.

## Known / Unknown

**Known**
- Current llama.cpp transcript rows persisted all-zero usage.
- The failed normalized stop reason was `length`.
- Letta displayed `max_tokens_exceeded`.
- `/context` estimated 96k while llama.cpp reported under 60k.
- The failed thinking content was 570 characters.
- The conversation explicitly stored `max_tokens: 128000` against a 110,240 llama.cpp context.

**Unknown**
- The raw OpenAI-compatible SSE payload for the failed response.
- Why llama.cpp emitted `finish_reason: "length"` after only 570 persisted thinking characters.
- The exact commit that introduced the regression; the user reports it began sometime in the 0.28.x series.
- Whether enabling streamed usage against this exact llama.cpp build returns correct token counts.

cpacker:
## Summary

- compact local conversations before provider dispatch once estimated context enters a reserved output-headroom band
- apply the same threshold after a turn, preferring exact provider usage and falling back to an estimate that includes the completed response
- preserve normal length stops for intentionally small configured output limits

Fixes #3508.

## Why no overflow was thrown

The missing compaction was a harness-policy gap, not primarily a llama.cpp usage bug.

The local adapter previously compacted after usage exceeded the full context window or after the provider threw a context overflow. But pi-ai makes a near-full request valid before dispatch by clamping the available output budget to:

    contextWindow - estimatedContext - 4096

with a one-token floor. For the reported 100K window and roughly 96K request, that becomes one output token. The provider therefore accepts the request and returns a legitimate length stop after one token. No overflow reaches Letta's retry path, so waiting for overflow cannot trigger compaction reliably.

This change treats context pressure as its own signal. It uses a 16,384-token reserve, matching the upstream coding harness, capped at 20% for small local windows. A 100K context therefore compacts above 83,616 tokens, before pi-ai has to collapse generation headroom.

## Relationship to #3314 and #3355

This does not restore the policy removed by #3355.

#3314 treated a reduced effective output budget as evidence of context overflow. That conflated two cases: automatic output clamping caused by a full context, and an intentionally small configured max output. #3355 correctly restored the rule that an ordinary length stop is not itself an overflow.

The new policy keeps that rule. It never classifies max output or a length stop as context pressure. It checks context usage independently, before dispatch and after completion. A regression test pins an explicit one-token max output as a normal length completion with no compaction.

## Ownership boundary

The provider-published pi-ai Model remains the source of truth for contextWindow. Letta's local harness owns only the policy for when to compact around that model metadata. This is separate from the endpoint-default fixes in #3512 and #3513: those fix which context window a discovered model publishes; this fixes what the harness does as any correctly-published window fills.

The provider-overflow recovery path remains in place as a bounded fallback if compaction cannot make a request fit.

## References

- pi-ai output clamp: https://github.com/earendil-works/pi/blob/cee5ff7520d8828bed9955ef00419e995d1f91e0/packages/ai/src/api/simple-options.ts#L12-L19
- upstream reserve policy: https://github.com/earendil-works/pi/blob/cee5ff7520d8828bed9955ef00419e995d1f91e0/packages/coding-agent/src/core/compaction/compaction.ts#L128-L137
- upstream threshold check: https://github.com/earendil-works/pi/blob/cee5ff7520d8828bed9955ef00419e995d1f91e0/packages/coding-agent/src/core/compaction/compaction.ts#L235-L238

## Verification

- 33 focused local-provider and compaction tests pass
- all 12 repository checks pass, including Biome and TypeScript

### DIFF
--- a/scripts/source-file-size-baseline.json
+++ b/scripts/source-file-size-baseline.json
@@ -2,7 +2,7 @@
   "src/agent/client-skills.test.ts": 1196,
   "src/agent/memory-git.ts": 2129,
   "src/backend/local-backend.test.ts": 2532,
-  "src/backend/local/local-backend.ts": 1030,
+  "src/backend/local/local-backend.ts": 1014,
   "src/backend/local/local-store.ts": 3594,
   "src/backend/pi-stream-adapter.test.ts": 1304,
   "src/cli/app/AppCoordinator.tsx": 5208,
--- a/src/backend/dev/pi-stream-adapter.ts
+++ b/src/backend/dev/pi-stream-adapter.ts
@@ -7,7 +7,6 @@ import type {
   SimpleStreamOptions,
   Tool,
   TSchema,
-  Usage,
 } from "@earendil-works/pi-ai";
 import { isContextOverflow, Type } from "@earendil-works/pi-ai";
 
@@ -48,9 +47,12 @@ import type {
   ProviderTurnInput,
 } from "./provider-turn-executor";
 import {
+  contextTokensFromUsage,
+  estimateProviderContextTokens,
   providerLettaChunk,
   providerLocalMessage,
   providerStreamPart,
+  shouldCompactForContextPressure,
 } from "./provider-turn-executor";
 
 const LOCAL_PROVIDER_MAX_RETRIES = 3;
@@ -65,6 +67,19 @@ export type PiStreamFunction = (
 ) => AsyncIterable<AssistantMessageEvent> & {
   result(): Promise<AssistantMessage>;
 };
+
+export interface LocalContextPressure {
+  contextTokens: number;
+  contextWindow: number;
+  phase: "preflight" | "post_turn";
+  source: "estimate" | "usage";
+}
+
+interface LocalCompactionResult {
+  uiMessages: LocalMessage[];
+  summary: string;
+  stats?: LocalCompactionStats;
+}
 
 export interface PiStreamAdapterOptions {
   stream?: PiStreamFunction;
@@ -85,14 +100,10 @@ export interface PiStreamAdapterOptions {
     summary: string;
     stats?: LocalCompactionStats;
   } | null>;
-  onContextUsage?: (
+  onContextPressure?: (
     input: ProviderTurnInput,
-    usage: Usage,
-  ) => Promise<{
-    uiMessages: LocalMessage[];
-    summary: string;
-    stats?: LocalCompactionStats;
-  } | null>;
+    pressure: LocalContextPressure,
+  ) => Promise<LocalCompactionResult | null>;
   onLlmStart?: (info: LlmStartInfo) => void | Promise<void>;
   onLlmEnd?: (info: LlmEndInfo) => void | Promise<void>;
 }
@@ -471,7 +482,7 @@ export class PiStreamAdapter implements ProviderStreamAdapter {
   private readonly localProviderAuthStorageDir?: string;
   private readonly modelsRuntime: LocalPiModelsRuntime;
   private readonly onContextWindowOverflow?: PiStreamAdapterOptions["onContextWindowOverflow"];
-  private readonly onContextUsage?: PiStreamAdapterOptions["onContextUsage"];
+  private readonly onContextPressure?: PiStreamAdapterOptions["onContextPressure"];
   private readonly onLlmStart?: PiStreamAdapterOptions["onLlmStart"];
   private readonly onLlmEnd?: PiStreamAdapterOptions["onLlmEnd"];
 
@@ -492,7 +503,7 @@ export class PiStreamAdapter implements ProviderStreamAdapter {
     this.abortSignal = options.abortSignal;
     this.localProviderAuthStorageDir = options.localProviderAuthStorageDir;
     this.onContextWindowOverflow = options.onContextWindowOverflow;
-    this.onContextUsage = options.onContextUsage;
+    this.onContextPressure = options.onContextPressure;
     this.onLlmStart = options.onLlmStart;
     this.onLlmEnd = options.onLlmEnd;
   }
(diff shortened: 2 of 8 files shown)


edt-xx:
Looks like this is working.  Have not had to do a manual compact today, after updating to 29.4.  
Thank You!
BTW.  It seems to compact earlier than it needs to.  I have the context limit at 100000, with a llama.cpp limit of 110240 and a thinking limit of 10240.  The compacts have been triggering around 85000.  Not really a problem but a bit more aggressive that I would like.  

Think
LOCAL_CONTEXT_COMPACTION_RESERVE_TOKENS = 16_384
could easily be 1/2 to 1/3 of this without problems here.  Its easy to work around this though.

PR #3522, opened 2026-07-27
fix(listener): preserve in-flight turns across relay reconnects

overlord-letta:
## Problem

A transient listener relay disconnect currently tears down the entire conversation runtime. That aborts an otherwise live Core response stream, clears queued input and approval ownership, and can leave a later client-tool request persisted on the server with no local owner.

## Change

- Treat ordinary relay loss as transport churn: pause delivery, preserve active turns/queues/resolvers, and rebind captured listener transports when replacement sockets open.
- Make queue draining and client-tool execution wait for an open replacement transport, so a tool cannot execute invisibly while disconnected.
- Keep intentional shutdown, policy close `1008`, and explicit `Replaced by new connection` supersession terminal on either split socket.
- Add real-socket regressions for close-after-streaming/reconnect/tool continuation, queued input, interactive approvals, split-socket replacement, and exact-once execution.

## Before / after

Before: a short relay move can cancel local ownership while the server-side run continues, leaving a pending client-tool boundary latent until a later turn cleans it up.

After: the same-process Core reader and turn lease survive ordinary relay churn; later client tools wait for the replacement socket, execute once, and continue normally.

## Scope

This is the narrow same-process relay-reconnect layer. It does not add durable run replay after the Core response reader dies, process-replacement recovery, or durable client-tool claims. Those require a separate server-derived resume protocol.

## Risk

This changes listener lifecycle semantics. Coverage specifically checks terminal versus transient close reasons, split control/stream ordering, resolver and queue ownership, stale socket guards, and no execution before transport recovery.

## Validation

- `bun test --isolate src/websocket/listener/auth-lifecycle-approval-reconnect.test.ts src/websocket/listener/auth-lifecycle.test.ts src/websocket/listener-reconnect-reminders.test.ts` (13 pass)
- `bun test --isolate src/websocket/listener` (336 pass)
- independent adversarial review: no material findings


cpacker:
Live Desktop evidence now confirms this PR's failure mode in LC 0.29.9:

- three active conversations visibly SAD'd/paused at the exact same instant
- listener PID stayed alive
- `~/.letta/logs/remote/...` records one shared reconnect at 19:22:33
- live runtime inspection afterward found incident turns still active, ruling out independent backend/model crashes

So #3522 is the correct runtime-ownership repair for a real transient relay recycle.

I also found why this particular recycle occurred: Electron's 60-second monitor callback was delayed 117 seconds under the renderer/listener load, past the 90-second pong timeout. The current wall-clock watchdog then terminates the shared socket before offering it a fresh ping. #3547 is stacked here to count actual unanswered probes instead. The Desktop-side 1 Hz full-state sync amplifier is fixed in letta-ai/letta-cloud#13437.

Timestamped chain and regression attribution are recorded on LET-9601: https://linear.app/letta/issue/LET-9601/crazy-fps-dropping-happening-on-latest-desktop-build#comment-5d5079ac

PR #3526, opened 2026-07-27
fix(app-server): return pagination metadata from conversation_messages_list

cpacker:
## Summary
- `conversation_messages_list_response` carried only the raw message page — no cursor, no more-history signal — so external clients building history backfill had no reliable way to walk older pages (reported by a mobile SDK consumer).
- Add `next_before` (oldest message ID in the page) and `has_more` to the response, computed by a shared `listConversationMessagePage()` helper: it trims overfilled backend pages, and probes with `limit: 1` only when a page is exactly full, so the common partial-page case costs no extra request.
- Normalize cursor semantics at the `APIBackend` boundary: the `Backend` contract (and the local backend) treat `before`/`after` chronologically (`before` = older, always), while the REST API interprets them relative to sort order. Descending requests now swap cursor keys before reaching the generated client, so all backends present the same chronological contract.
- Tests: page-walk regression (multiple descending pages, no duplicates, correct `has_more` progression), overfilled/exact/partial page trimming, ascending oldest-ID selection, and cursor-normalization assertions for both orders.

### DIFF
--- a/scripts/source-file-size-baseline.json
+++ b/scripts/source-file-size-baseline.json
@@ -40,7 +40,7 @@
   "src/tools/impl/message-channel.ts": 1187,
   "src/tools/manager.ts": 3281,
   "src/tools/tool-execution-context.test.ts": 1422,
-  "src/types/protocol_v2.ts": 2961,
+  "src/types/protocol_v2.ts": 2963,
   "src/websocket/listen-client-concurrency.test.ts": 3017,
   "src/websocket/listen-client-protocol.test.ts": 6723,
   "src/websocket/listener/commands/channels.ts": 1390,
--- a/src/backend/api-backend.test.ts
+++ b/src/backend/api-backend.test.ts
@@ -320,6 +320,20 @@ describe("APIBackend", () => {
     );
 
     await backend.listConversationMessages("conv-1", {
+      before: "message-default-order-page",
+      limit: 10,
+    });
+    expect(listConversationMessagesMock).toHaveBeenLastCalledWith(
+      "conv-1",
+      {
+        after: "message-default-order-page",
+        before: undefined,
+        limit: 10,
+      },
+      undefined,
+    );
+
+    await backend.listConversationMessages("conv-1", {
       before: "message-older-page",
       order: "asc",
       limit: 10,
--- a/src/backend/api-backend.test.ts
+++ b/src/backend/api-backend.test.ts
@@ -296,4 +296,58 @@ describe("APIBackend", () => {
       agentId: "agent-1",
     });
   });
+
+  test("normalizes descending message cursors to chronological before and after", async () => {
+    const backend = new APIBackend({
+      getClient: getClientMock as unknown as () => Promise<APIClient>,
+      forkConversation: forkConversationMock,
+    });
+
+    await backend.listConversationMessages("conv-1", {
+      before: "message-older-page",
+      order: "desc",
+      limit: 10,
+    });
+    expect(listConversationMessagesMock).toHaveBeenLastCalledWith(
+      "conv-1",
+      {
+        after: "message-older-page",
+        before: undefined,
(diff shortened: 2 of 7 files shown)


cpacker:
Reviewed from the mobile-client side (I filed the original report — thanks for the fast turnaround, and for tracing the `Date` leak properly; my prompt only had the receipt, not the cause).

Two notes: one likely bug for `order`-omitting clients, one compatibility confirmation.

### 1. The cursor swap requires an explicit `order: "desc"`, but the page metadata defaults to desc

`toApiConversationMessageListBody` only swaps when `order` is *explicitly* `"desc"`:

```ts
if (!body || body.order !== "desc" || (!body.before && !body.after)) {
  return body;
}
```

…while `listConversationMessagePage` treats a **missing** `order` as desc:

```ts
const order = normalizedQuery.order === "asc" ? "asc" : "desc";
const oldestMessage = order === "asc" ? messages[0] : messages[messages.length - 1];
const nextBefore = messageId(oldestMessage);
```

That matters because the SDK omits `order` unless the caller passes it — `AppServerSession.listMessages()` builds its query with `if (options.order !== undefined) query.order = options.order`. So the common call `session.listMessages({ limit: 50, before })` arrives with `before` set and **no** `order`.

For such a request against the API backend:

- `next_before` is computed as the last element, i.e. correct for a descending page;
- but the chronological→order-relative cursor swap is skipped, so that same cursor is then interpreted order-relative on the way out.

The client is handed a cursor that pages the wrong direction when it sends it back. Our mobile client is exactly this caller (it never sets `order`), so it would still fail to reach older history after this fix.

Suggested: default the order the same way in both places, e.g.

```ts
const effectiveOrder = body.order ?? "desc";     // or share one helper with listConversationMessagePage
if (!body || effectiveOrder !== "desc" || (!body.before && !body.after)) return body;
```

and add a normalization case for `{ before, limit }` with no `order` — the descending-walk test currently always passes `order: "desc"` explicitly, so this path isn't covered.

Worth double-checking what the REST API's own default order actually is, too: if it defaults to desc, the two defaults agree and only the swap guard needs fixing; if it defaults to asc, the metadata default is the one that's wrong.

### 2. Field naming is compatible — confirmed

`next_before` / `has_more` snake_case is correct for SDK consumers. The SDK reads either casing:

```js
const nextBefore = typeof response.nextBefore === "string" || response.nextBefore === null
  ? response.nextBefore
  : typeof response.next_before === "string" || response.next_before === null
    ? response.next_before : undefined;
const hasMore = typeof response.hasMore === "boolean" ? response.hasMore
  : typeof response.has_more === "boolean" ? response.has_more : undefined;
```

So `session.listMessages()` will surface both fields as `nextBefore` / `hasMore` with no SDK change needed.

Also 👍 on the `limit: 1` probe only for exactly-full pages — that resolves the `items.length >= limit` ambiguity on a final page without costing anything on partial pages.

Happy to re-verify end to end against a local app-server once this lands; our client has a workaround (deeper first page on remote, stop offering "load older" after a cursor rejection) that I'll delete at that point.

cpacker:
Good catch — this is valid. I checked every default before changing it: the SDK documents omitted `order` as descending, the Python REST route defaults to `desc`, TS core-runtime defaults to `desc`, and the local backend defaults to `desc`. The only inconsistent site was my API cursor-normalization guard, which required the value to be explicit.

Fixed in a7fb7900:
- added one shared `DEFAULT_CONVERSATION_MESSAGE_ORDER = "desc"` used by both page-metadata calculation and API cursor normalization;
- added a direct `{ before, limit }` normalization assertion with no order;
- changed the full multi-page descending walk to omit order entirely, matching the ordinary SDK call path you identified.

The focused tests pass (8/8), `bun run check` passes all 12 checks, and commit hooks/typecheck passed. Thanks for tracing this from the actual consumer path.

PR #3528, opened 2026-07-27
fix(channels): route shared reload command

overlord-letta:
## Problem

Channel users could not invoke the listener reload path through the shared slash-command router. The existing implementation only accepted Slack mention-scoped reload commands, leaving other routed channel chats to receive unsupported-command guidance.

## Change

- Add `/reload` to the shared channel command registry and help text.
- Route it through the existing per-conversation reload handler.
- Preserve Slack mention and legacy `!reload` behavior.
- Keep reload in the mutating command tier when channel command gating is enabled.
- Clarify that reload refreshes settings, local mods, and agent secrets.

## Scope

This exposes the existing reload behavior. It does not add a broader channel adapter hot-reload state machine or a raw command-injection escape hatch.

## Validation

- `bun test src/channels/commands.test.ts src/channels/access-control.test.ts src/channels/registry-reload-routing.test.ts` — 52 pass
- `bun run check` — 12/12 pass

Supersedes #3194, #3111, and #3448.


just-cameron (approved):
Happy to merge when tests pass.

PR #3533, opened 2026-07-28
feat(permissions): add strict mode that forces all tools through approval

cpacker:
## Summary

- Adds `strict` permission mode that disables all auto-allow paths
- All tools default to "ask" so the SDK `canUseTool` callback must explicitly approve every call
- Working directory auto-allow for Read/Glob/Grep/ListDir → skipped
- Read-only shell command auto-allow → skipped
- Skill tool auto-allow → skipped
- Default "allow" for read tools → changed to "ask"

Deny/alwaysAsk rules, explicit allow rules, and CLI `--allowedTools` still work as expected.

**Use case**: Unattended sessions where reads should not auto-allow, e.g. preventing credential exfiltration from `~/.letta/settings.json` even when Read tool is available.


.husky/_
```diff
@@ -0,0 +1 @@
+/Users/loaner/dev/letta-code-prod/.husky/_
```
amelia-letta: This is a symlink to `/Users/loaner/dev/letta-code-prod/.husky/_` — a local machine path that won't exist on other machines. Should not be committed, will break husky setup elsewhere.

francisjuma (commented):
Few error still in code run

PR #3543, opened 2026-07-29
feat: add cloud conversation handoff command

sarahwooders:
Adds `/cloud` and `/cloud <instruction>` for hosted Letta agents.

The command uses the same canonical `POST /v1/sandboxes` path as the Desktop environment selector, routes subsequent turns through the returned sandbox, displays the final remote response locally, and restores a persisted target when the server supports it.

The command is backward-compatible with the currently deployed Cloud API. New servers advertise `repositoryHandoffVersion: 1`; when that marker is present, dirty and unpushed repository state is transferred exactly through an isolated `commit-tree` snapshot and ephemeral `refs/letta/handoffs/...` ref. On older deployments, `/cloud` still starts and routes through a normal sandbox, suppresses snapshot provenance, and explicitly reports that exact local repository state was not transferred.

Snapshot creation never touches the user’s branch, real index, or working tree. Ignored files are excluded, likely-secret and oversized files are rejected, and temporary refs are deleted after provisioning.

The full snapshot and Desktop synchronization behavior depends on letta-cloud PR #13459; basic `/cloud` execution does not.

Validated with all 12 `bun run check` gates, focused fallback and snapshot invariance tests, and a real private-ref push/fetch/exact-commit/deletion cycle against GitHub.


src/cli/app/submit-cloud-command.ts
```diff
@@ -0,0 +1,335 @@
...
+  const key = cloudKey(params.agentId, params.conversationId);
+  let cloudPrepared = false;
+  try {
+    const repositoryHandoffRequested = Boolean(
+      getGithubRepositoryForDirectory(params.projectDirectory),
+    );
+    let cloud = await prepareCloudConversation({
+      agentId: params.agentId,
+      conversationId: params.conversationId,
+    });
```
amelia-letta, line 134: First `prepareCloudConversation` call omits `setAsRuntimeTarget: true`, so when the server doesn't support repository handoffs, the sandbox won't persist across CLI restarts. Previous version always set it.

```suggestion
    let cloud = await prepareCloudConversation({
      agentId: params.agentId,
      conversationId: params.conversationId,
      setAsRuntimeTarget: true,
    });
```

sarahwooders:
Closing after correcting the ownership boundary: remote-environment messaging and execution-target selection are Desktop responsibilities. The handoff will be implemented through Desktop’s existing Cloud environment flow in letta-cloud PR #13459.

PR #3589, opened 2026-07-31
fix(telemetry): preserve fatal process semantics

### LINKED ISSUE #3579: Telemetry fatal-error handlers allow the process to continue after uncaught errors
## Summary

The telemetry service installs `uncaughtException` and `unhandledRejection` listeners to record and drain telemetry, but neither handler exits, rethrows, or restores the runtime's fatal behavior.

Once these listeners are installed, an uncaught error can be recorded and the letta-code process can continue running in an undefined or partially corrupted state.

## Relevant code

`src/telemetry/index.ts:399-435`:

```ts
process.on("uncaughtException", (error) => {
  void (async () => {
    try {
      // track error
      await this.drain();
    } catch {
      // Silently ignore - don't prevent process from exiting
    }
  })();
});

process.on("unhandledRejection", (reason) => {
  void (async () => {
    try {
      // track error
      await this.drain();
    } catch {
      // Silently ignore - don't prevent process from exiting
    }
  })();
});
```

The comments say not to prevent process exit, but registering these listeners changes the default fatal behavior. There is no explicit exit after the asynchronous drain.

The EPIPE/EIO/EBADF and rate-limit early returns also return from the handler without terminating.

## Minimal reproduction of the runtime behavior

Using the repository's Bun runtime:

```sh
bun -e 'process.on("uncaughtException", () => console.log("caught")); setTimeout(() => console.log("continued"), 10); throw new Error("boom")'
```

Output:

```text
caught
continued
```

A long-running listener with active sockets/timers can therefore remain alive indefinitely after the same class of failure.

## Why this matters

Continuing after an uncaught exception is unsafe because invariants may already be broken: locks may be held, state may be partially mutated, adapters may be half-started, or an exact-once lifecycle may be incomplete. It can turn a visible crash into silent state corruption or repeated incorrect behavior.

The current telemetry tests cover batching/draining, but not fatal exit semantics in a child process.

## Expected behavior

Telemetry should make a best effort to record fatal errors without changing the application's fatal outcome. After a bounded flush attempt, the process should terminate with a non-zero exit code.

Signals used for normal shutdown should retain their existing intended exit codes.

## Suggested direction

- Use a one-shot fatal handler with a re-entry guard.
- Set a non-zero `process.exitCode` immediately.
- Race telemetry drain against a short hard timeout.
- Explicitly terminate after the drain/timeout, or remove the listener and rethrow in a way that preserves fatal semantics.
- Avoid recursively processing errors thrown by telemetry itself.
- Consider removing listeners during telemetry cleanup/reinitialization to avoid duplicate handlers.

## Tests to add

Run a small fixture in a child Bun process and assert:

- `uncaughtException` exits non-zero after attempting a drain.
- An unhandled rejection exits non-zero.
- A hanging telemetry request cannot keep the process alive indefinitely.
- Only one fatal event is emitted during recursive failures.

Observed on `main` at `ed3fec21`.

GautamSharma99:
## Summary

Fixes #3579.

Telemetry previously registered `uncaughtException` and `unhandledRejection` listeners that recorded the failure and drained events, but never restored the runtime fatal outcome. In Bun, the presence of those listeners suppresses default termination, so a long-running letta-code process could continue after an uncaught failure with partially mutated state or broken invariants.

This PR makes fatal telemetry best-effort without making fatal errors recoverable.

## What changed

- Extracted process-fatal handling into a focused `installFatalErrorHandlers` helper.
- Sets `process.exitCode = 1` synchronously as soon as either fatal event arrives.
- Uses a shared one-shot guard across uncaught exceptions and unhandled rejections so recursive failures do not enqueue duplicate fatal events or start competing drains.
- Attempts to record the first actionable fatal event, then races telemetry draining against a hard 3-second timeout.
- Calls `process.exit(1)` after the drain settles, fails, or times out, ensuring active sockets and timers cannot keep the process alive.
- Keeps existing noise filtering for broken-pipe errors and rate-limit rejections, but separates reporting policy from termination policy: filtered uncaught failures still terminate non-zero.
- Stores and removes SIGINT and fatal listener references during `telemetry.cleanup()`, preventing duplicate process listeners across cleanup/reinitialization cycles.
- Updated the stale crash-telemetry comment to reflect the new bounded-flush behavior and the remaining persistence limitation.

## Why explicit termination

Setting only `process.exitCode` is insufficient for listener/headless processes because active handles can keep the event loop alive indefinitely. Removing the listener and rethrowing also creates recursive/error-ordering hazards while an asynchronous flush is running. A bounded flush followed by explicit non-zero termination gives telemetry a short delivery window while preserving deterministic fatal semantics.

The exit code is set before tracking or draining so failures inside telemetry itself cannot accidentally produce a successful exit. Tracking and drain errors are deliberately contained because neither should interfere with termination.

## Tests

Added real child-Bun-process coverage rather than mocking `process.exit`:

- uncaught exceptions attempt a drain and exit with code 1;
- unhandled rejections attempt a drain and exit with code 1;
- a drain promise that never settles is cut off by the hard timeout;
- a recursive fatal event records only the first failure;
- filtered fatal reasons are not reported but still exit non-zero;
- cleanup removes both installed fatal listeners.

Validation completed:

- `bun test src/telemetry/fatal-error-handler.test.ts` — 6 passed
- focused telemetry suites — 20 passed before the final additional filtered-error case
- `bun run check` — all 12 repository checks passed

## Scope

Normal SIGINT shutdown retains its existing successful exit behavior. This change only affects uncaught exceptions and unhandled rejections, which now reliably remain fatal after telemetry is initialized.

### DIFF
--- a/src/telemetry/fatal-error-handler.test.ts
+++ b/src/telemetry/fatal-error-handler.test.ts
@@ -1,0 +1,158 @@
+import { describe, expect, test } from "bun:test";
+import { pathToFileURL } from "node:url";
+import { installFatalErrorHandlers } from "./fatal-error-handler";
+
+type FatalScenario =
+  | "filtered-rejection"
+  | "hanging-drain"
+  | "recursive-failure"
+  | "uncaught-exception"
+  | "unhandled-rejection";
+
+interface ScenarioResult {
+  exitCode: number;
+  output: string;
+  runtimeMs: number;
+}
+
+const handlerModuleUrl = pathToFileURL(
+  `${import.meta.dir}/fatal-error-handler.ts`,
+).href;
+
+async function runFatalScenario(
+  scenario: FatalScenario,
+): Promise<ScenarioResult> {
+  const script = `
+    import { installFatalErrorHandlers } from ${JSON.stringify(handlerModuleUrl)};
+
+    const scenario = process.env.FATAL_TEST_SCENARIO;
+    installFatalErrorHandlers({
+      timeoutMs: 50,
+      trackError(errorType, message) {
+        console.log(\`tracked:\${errorType}:\${message}\`);
+      },
+      drain() {
+        console.log("drain-attempted");
+        if (scenario === "recursive-failure") {
+          process.emit("uncaughtException", new Error("recursive"));
+        }
+        if (scenario === "hanging-drain") {
+          return new Promise(() => {});
+        }
+        return Promise.resolve();
+      },
+    });
+
+    setTimeout(() => console.log("process-continued"), 1_000);
+    if (
+      scenario === "unhandled-rejection" ||
+      scenario === "filtered-rejection"
+    ) {
+      const message =
+        scenario === "filtered-rejection"
+          ? "429 rate limit fixture"
+          : "rejection-fixture";
+      void Promise.reject(new Error(message));
+    } else {
+      queueMicrotask(() => {
+        throw new Error("exception-fixture");
+      });
+    }
+  `;
+
+  const startedAt = performance.now();
+  const child = Bun.spawn([process.execPath, "-e", script], {
+    env: {
+      ...process.env,
+      FATAL_TEST_SCENARIO: scenario,
+    },
+    stderr: "pipe",
+    stdout: "pipe",
+  });
+  const [exitCode, stdout, stderr] = await Promise.all([
+    child.exited,
+    new Response(child.stdout).text(),
+    new Response(child.stderr).text(),
+  ]);
+
+  return {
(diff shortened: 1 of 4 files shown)


devanshrj:
Thanks @GautamSharma99 — the core fix and child-process coverage here were solid. Before merging, I pushed a small follow-up to handle rejection reasons whose string conversion itself throws, which could otherwise bypass the forced exit. I also reused the shared error-normalization helper and removed the wall-clock timing assertion from the timeout test.

Merged now. Thanks again for tracking this down and putting together the main fix!

PR #3592, opened 2026-07-31
fix(headless): honor can_use_tool deny interrupts

### LINKED ISSUE #3582: Headless can_use_tool deny responses ignore the interrupt flag
## Summary

The bidirectional headless protocol accepts `interrupt: true` on a denied `can_use_tool` response, but the headless handler discards the flag. A client can explicitly request that denial interrupt the turn, yet letta-code continues with ordinary tool-denial behavior.

The source contains TODOs documenting that the field is exposed but not wired up.

## Relevant code

`src/types/protocol.ts:751-755` defines:

```ts
export interface CanUseToolResponseDeny {
  behavior: "deny";
  message: string;
  /** TODO: Not wired up yet - infrastructure exists in TUI */
  interrupt?: boolean;
}
```

`src/headless.ts:4160-4170` parses the response but drops `interrupt`:

```ts
if (response.behavior === "allow") {
  // ...
} else {
  result = {
    decision: "deny",
    reason: response.message,
    // TODO: handle interrupt flag
  };
}
```

The result type at `src/headless.ts:4131-4135` has nowhere to carry interruption state.

## Reproduction

1. Start headless mode using the bidirectional JSON protocol.
2. Send a request that results in a `can_use_tool` control request.
3. Reply with the matching request ID and:

   ```json
   {
     "behavior": "deny",
     "message": "Stop this turn",
     "interrupt": true
   }
   ```

4. Observe that the response becomes a normal denial reason; no turn cancellation is triggered.

This differs from the semantic meaning exposed by the protocol and from the comment that the infrastructure exists in the TUI.

## Why this matters

Clients use `interrupt` to distinguish “deny this tool but let the model recover” from “deny and stop the current turn.” Ignoring that distinction can let the provider continue generating or attempt additional tools after the user requested a stop.

It also makes the public TypeScript protocol misleading: the field is accepted without warning but has no effect.

The repository has broad headless interrupt tests, but no test combining a bidirectional `can_use_tool` deny response with `interrupt: true`.

## Expected behavior

- `interrupt: false` or an omitted flag should preserve ordinary deny behavior.
- `interrupt: true` should cancel the active turn through the same lifecycle-safe interruption path used elsewhere.
- The turn should emit exactly one terminal outcome and should not continue tool/provider work afterward.

## Suggested direction

- Carry the flag through the local permission-decision result type.
- Invoke the existing turn abort/cancellation mechanism rather than implementing a separate shortcut.
- Ensure deferred input lines are restored correctly when interruption occurs while waiting for a control response.
- Define whether the denied tool result is emitted before the interrupted terminal event, and keep that ordering consistent with TUI behavior.

## Tests to add

- Bidirectional deny without `interrupt` continues as today.
- Deny with `interrupt: true` aborts the active turn.
- No provider/tool activity occurs after the interrupted terminal event.
- The terminal event is emitted exactly once.
- A response for a stale request ID cannot interrupt a newer turn.

Observed on `main` at `ed3fec21`.

GautamSharma99:
## Summary

Fixes #3582.

Bidirectional headless clients can already send `interrupt: true` with a denied `can_use_tool` response, but the runtime previously discarded that field and continued the approval flow. This change carries the interrupt intent into the active turn lifecycle and terminates the turn through its existing `AbortController`.

## What changed

- Extracted the permission-response wait loop into a focused `headless-permission` owner module.
- Added an explicit interrupted state to the local permission result.
- Abort the active turn synchronously when the matching deny response has `interrupt: true`.
- Break out of the approval loop before emitting local tool events, executing any decision batch, or starting another provider round.
- Continue to treat omitted or false `interrupt` values as ordinary denials.
- Preserve unrelated and malformed stdin lines in their original order for the main loop.
- Correlate responses to the active permission request so a stale request ID cannot interrupt the current turn.
- Updated the public protocol comment now that the field is wired up.
- Updated the deterministic Bash fixture with its required `description` argument so the end-to-end approval path is exercised rather than auto-denied during validation.
- Reduced `src/headless.ts` by 51 lines and lowered its source-size baseline from 5,243 to 5,192.

## Lifecycle and ordering

For an interrupting denial, the runtime does not synthesize or emit a denied tool return. It aborts the current controller and exits the approval loop immediately. The turn emits one terminal `result` with subtype `interrupted`, and no tool execution or provider continuation occurs for that response. If the backend still has a dangling approval, the existing post-interrupt recovery path resolves it before the next user turn.

This ordering prevents an auto-approved sibling decision in the same approval batch from executing after the client requested a stop.

## Tests

Added focused coverage for:

- deny without `interrupt`, which retains the current continuation behavior;
- deny with `interrupt: true`, including abort-before-return ordering;
- stale interrupting responses, which are deferred and cannot abort a newer request;
- preservation and ordering of other deferred stdin lines;
- a two-turn child-process protocol scenario proving:
  - ordinary deny emits a tool return and continues to a provider result;
  - interrupting deny emits exactly one interrupted result;
  - no tool return or assistant/provider message is emitted for the interrupted approval.

Validation completed:

- `bun test src/headless-permission.test.ts src/headless-interrupt-latch.test.ts src/backend/fake-headless-backend.test.ts` — 16 passed, 0 failed
- `bun run check` — all 12 checks passed


devanshrj (approved):

devanshrj:
Thanks @GautamSharma99 for the fix! Before merging, we updated the branch with the latest `main`, improved the end-to-end test diagnostics, and fixed Windows test isolation. Appreciate the solid implementation and thorough test coverage.

PR #3601, opened 2026-07-31
feat(whatsapp): support message reactions

just-cameron:
## Problem
WhatsApp message reactions were not handled at the Baileys reaction event boundary. In Baileys 6.7.21, real add/remove reactions arrive on `sock.ev.on("messages.reaction", entries)`; `reactionMessage` content seen through `messages.upsert` is only the synthetic wrapper and should not be routed as an inbound chat message.

## Before / after
Before, reactions could be missed, misattributed across PN/LID identities, or routed into pairing/unbound prompts. After, reactions parse into structured inbound messages with canonical sender/target metadata, dedupe by canonical chat, and leave ordinary text/media upserts unchanged.

## Controls and behavior
- Access and routing: denied, unpaired, or unrouted reactions are dropped silently. Only reactions to messages owned by the linked account are surfaced; unknown or inbound-owned targets fail closed.
- LID/PN: Baileys `senderPn`, `senderLid`, `participantPn`, and `participantLid` are preserved and used for first-reaction identity resolution.
- Removal/groups: empty reaction text removes a reaction and does not require an emoji. Group outbound reactions reuse the original stored Baileys target key, including participant, for add and removal.
- Restart limit: the target-key cache is process-local and bounded. After restart or expiry, outbound group reactions require the target message to be observed again.

## Risk
Channel adapter/parser change limited to WhatsApp. Main risk is Baileys event shape variance or process-local cache expiry; both fail closed instead of sending malformed group reaction keys or surfacing untrusted targets.

## Tests
- 34 focused WhatsApp reaction tests
- 130 WhatsApp tests
- `bun run check`

Depends on public letta-code PR #3599: https://github.com/letta-ai/letta-code/pull/3599
Replaces the reaction portion of public letta-code PR #3400 while preserving useful source material: https://github.com/letta-ai/letta-code/pull/3400

### DIFF
--- a/src/channels/registry-inbound.ts
+++ b/src/channels/registry-inbound.ts
@@ -360,6 +360,9 @@ export function createChannelInboundRouter(deps: {
     // exact topic route exists; a disabled exact route remains authoritative.
     const route = loadRouteForInboundMessage(msg, accountId);
     if (!route) {
+      if (msg.reaction) {
+        return;
+      }
       await adapter.sendDirectReply(
         msg.chatId,
         buildUnboundRouteInstructions(msg.channel, msg.chatId),
--- a/src/channels/whatsapp-message-channel.test.ts
+++ b/src/channels/whatsapp-message-channel.test.ts
@@ -111,4 +111,21 @@ describe("WhatsApp MessageChannel actions", () => {
       }),
     );
   });
+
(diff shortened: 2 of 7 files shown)


just-cameron:

PR #3602, opened 2026-07-31
feat(whatsapp): add outbound attachment policy

just-cameron:
## Problem
WhatsApp outbound uploads currently follow the requested `mediaPath` once a message action emits an upload, so account config has no default-off way to restrict attachment recipients, file types, or source paths. Canonical WhatsApp identity resolution and symlinked paths make naive checks easy to get wrong.

## Behavior
Before:
- WhatsApp had no attachment policy config surface.
- Outbound media payloads used the requested path/name directly.
- Non-Ogg audio uploads were rejected as invalid voice memo uploads.

After:
- Attachment filtering is default-off and only runs when `attachment_filter: true`.
- The send target is resolved to the canonical WhatsApp JID first, including stored LID to phone mappings.
- Policy denial happens before presence updates or network sends.
- Recipient allowlists support normalized direct phone identities and exact group JIDs.
- Source files and allowed directories are realpath resolved, with direct-child containment by default and optional recursive containment.
- Extension-derived MIME, payload path, and document basename come from the canonical resolved file target.
- Ogg/Opus remains push-to-talk audio; MP3/M4A/WAV and other files send as regular document attachments.

## Config surface
Adds WhatsApp account/plugin fields:
- `attachment_filter`
- `attachment_mime_types`
- `attachment_allowed_recipients`
- `attachment_allowed_paths`
- `attachment_path_recursive`

These are persisted, normalized on legacy load, exposed in account/config snapshots, and accepted through protocol/plugin config validation.

## Risk and limits
Default-off preserves existing behavior unless explicitly enabled. The high-risk path is outbound media selection, especially canonical recipient resolution and symlink containment before any send side effects.

This is path-selection and symlink protection. It is not content MIME verification, and it is not hostile-local-writer TOCTOU safety. MIME is inferred from the resolved canonical file extension.

## Tests
- 45 focused WhatsApp attachment/media/service/protocol tests
- 118 broader WhatsApp tests
- `bun run check`

## Stacking and credit
Depends on #3599.

Credits #3397 for the original attachment-policy direction. This PR replaces only its WhatsApp attachment-policy portion with a focused stacked implementation.

### DIFF
--- a/src/channels/accounts.ts
+++ b/src/channels/accounts.ts
@@ -58,6 +58,11 @@ const SNAKE_TO_CAMEL: Record<string, string> = {
   inbound_debounce_ms: "inboundDebounceMs",
   listen_mode: "listenMode",
   media_max_bytes: "mediaMaxBytes",
+  attachment_filter: "attachmentFilter",
+  attachment_mime_types: "attachmentMimeTypes",
+  attachment_allowed_recipients: "attachmentAllowedRecipients",
+  attachment_allowed_paths: "attachmentAllowedPaths",
+  attachment_path_recursive: "attachmentPathRecursive",
   mention_patterns: "mentionPatterns",
   recipient_aliases: "recipientAliases",
   remove_stale_routes: "removeStaleRoutes",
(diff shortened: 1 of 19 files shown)


just-cameron:

PR #3603, opened 2026-07-31
feat(whatsapp): debounce inbound messages

just-cameron:
## User problem
WhatsApp users can send several short messages in a burst. Without a debounce window, each fragment becomes a separate inbound turn, while read receipts must still avoid acknowledging messages that were dropped, canceled, stale, or failed during handoff.

## What changed
- Adds optional WhatsApp inbound text debouncing keyed by canonical account/chat/sender identity.
- Keeps the default behavior at zero milliseconds, so existing accounts continue immediate delivery unless configured otherwise.
- Delivers debounced text on the trailing edge and combines delivered text with newline separators.
- Marks WhatsApp messages read only after successful inbound handoff, preserving best-effort non-blocking read failure handling.
- Ties pending delivery and read receipts to socket generation so stop, reconnect, and stale socket traffic do not leak delivery or read state.
- Keeps dropped group traffic, unauthorized traffic, reactions, empty bodies, and attachment paths from being accidentally read or merged.

## Before / after
Before: rapid WhatsApp text fragments were dispatched separately, and an early read-receipt implementation could mark messages read before delivery was guaranteed.

After: configured accounts batch same sender/chat text bursts, default accounts behave as before, and only successfully handed-off entries are acknowledged as read.

## Risks and limits
- Debounce applies only to plain inbound text with no attachments; attachment and non-text paths continue immediate handling.
- Read receipt failures are logged but intentionally do not block delivery after a successful handoff.
- This is stacked on #3599 and targets `maintainer/whatsapp-lid-identity-clean`.
- This cleanly replaces the debounce/read-receipt portions of public #3396 and #3400.

## Tests
- 37 focused WhatsApp/debounce tests
- 40 sibling debounce/ingress tests
- `bun run check` passed

### DIFF
--- a/src/channels/accounts.ts
+++ b/src/channels/accounts.ts
@@ -207,6 +207,13 @@ function prepareAccountForStorage(account: ChannelAccount): ChannelAccount {
   return cloned;
 }
 
+function normalizeInboundDebounceMs(value: unknown): number | undefined {
+  if (typeof value !== "number" || !Number.isFinite(value) || value < 0) {
+    return undefined;
+  }
+  return Math.trunc(Math.min(value, 10000));
+}
+
 function cloneAccount<T extends ChannelAccount>(account: T): T {
   const cloned = {
     ...account,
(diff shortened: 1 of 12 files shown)


just-cameron:

PR #3398, opened 2026-07-17
feat(whatsapp): reconnect guardrail for session-conflict loops

signo:
> [!CAUTION]
> **SUPERSEDED HEAD — DO NOT REVIEW OR MERGE YET.**
>
> The current remote head (`34522ba3`) has been replaced locally by reviewed checkpoint `7597ca4`, based directly on canonical identity PR #3395 (`2c801028`). The rebuild deduplicates close events per socket generation, trips only after six distinct unstable generations in 60 seconds, resets history/backoff only after stable uptime, and owns/cancels reconnect plus stability tasks deterministically.
>
> Focused/complete validation and 12/12 checks pass. Combined checkpoint `d7c60a59` passes 193/193 WhatsApp tests and the mandatory Arrakis gate. Nothing has been pushed. This PR remains blocked until #3395 merges and publication is explicitly authorized.
>
> Process-level stale session-lease recovery after abrupt container death is separate and tracked privately; it must not be "fixed" with PID-match-only ownership because container PID values can collide.

## Historical description of the superseded head

If the WhatsApp socket disconnects repeatedly within a short window, stop reconnecting and surface account error state rather than looping forever.

👾 Generated with [Letta Code](https://letta.com)

PR #3604, opened 2026-07-31
fix(whatsapp): stop unstable reconnect loops

just-cameron:
## Summary
WhatsApp can get stuck in a tight reconnect loop when Baileys repeatedly closes a linked-device socket, usually from a competing session. This PR adds a generation-based reconnect state machine and terminal-state ownership so late or duplicate events cannot keep restarting the channel or overwrite the useful error.

Depends on #3599.
Credits and replaces #3398.

## What changed
- Tracks socket generations and counts only one close per generation, so duplicate close events from the same socket do not inflate the unstable disconnect threshold.
- Stops the adapter after 6 distinct unstable disconnect generations inside 60s, with exponential backoff before that and a 60s stable-open reset. Reconnect and stable-reset timers are unrefed.
- Lets the adapter claim terminal connection state from the session layer, so conflict and circuit-breaker errors are not overwritten by generic `disconnected` or `logged_out` fallback state.
- Guards reconnect timer callbacks and rejected reconnect attempts by generation plus running/stopping state, so stop, restart, and account reconfigure paths do not let stale async failures mutate current state or schedule more reconnects.

## Before / after
Before: repeated close events or a stale reconnect failure could leave the channel retrying indefinitely, or replace a conflict/circuit-breaker error with generic session close state.

After: unstable reconnects stop with explicit user-facing guidance; stable connections reset history; late closes and rejections from old generations are ignored.

## Explicitly not included
This does not change the PID/container session lease behavior. Existing lease locking remains as-is.

No live WhatsApp network session was exercised or claimed; validation is unit/integration coverage around adapter/session ordering and lifecycle edges.

## Tests
- `bun test src/channels/whatsapp-adapter.test.ts src/channels/whatsapp-session.test.ts src/channels/registry-lifecycle.test.ts src/channels/lifecycle-error.test.ts` (`40 pass`)
- `bun run check` (`12 checks passed`)

### DIFF
--- a/src/channels/whatsapp-adapter.test.ts
+++ b/src/channels/whatsapp-adapter.test.ts
@@ -1,4 +1,5 @@
 import { afterEach, describe, expect, test } from "bun:test";
+import { rmSync } from "node:fs";
 import {
   createWhatsAppAdapter,
   isWhatsAppConflictDisconnect,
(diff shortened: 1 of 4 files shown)


just-cameron:

PR #3605, opened 2026-07-31
feat(whatsapp): add typing indicators

just-cameron:
## Summary
WhatsApp users currently get no active waiting signal while Letta is processing a turn, so a long-running reply can look idle. This adds an opt-in waiting behavior (`waiting_behavior: "typing_indicator"`) that sends WhatsApp composing presence while the turn is processing and pauses it when the turn finishes or is otherwise cleared.

## Behavior
- Default remains off. Existing configs/accounts normalize missing `waiting_behavior` to `off`; account config accepts only `off` and `typing_indicator`.
- Lifecycle/socket/batch ownership: `queued` is inert, `processing` starts managed typing per batch/source, and terminal lifecycle events, including error outcomes, stop only matching batch/source entries before any error reply handling. Multiple sources in the same batch/chat are reference-counted; a newer batch supersedes older typing without letting the older finish pause the new entry.
- Socket/chat ownership: typing entries keep the socket that started them, refresh composing presence, cap lifetime, canonicalize WhatsApp LID JIDs through the existing LID resolver, and clear the correct owner on reconnect/session conflict/stop.
- Cancellation and sends: outbound messages, direct replies, and reactions clear managed typing before sending. When typing was not lifecycle-managed, ordinary message sends keep the existing one-shot composing presence. Presence errors are best-effort and contained.
- Ordinary negative cases: off mode does nothing for lifecycle events; non-WhatsApp/mismatched-account sources, invalid waiting behavior values, queued events, stale finishes, and absent sockets are inert.

## Risk
Channel runtime change for WhatsApp presence only. The feature is default-off, the managed state is in-memory per adapter/socket, and failure to send presence is swallowed so message delivery is not blocked. The main risk is stale presence state across reconnect/stop paths, covered by focused lifecycle tests.

## Tests
- `bun test src/channels/whatsapp/typing-controller.test.ts src/channels/whatsapp/typing-adapter.test.ts src/channels/whatsapp/waiting-behavior-config.test.ts` (17 pass)
- `bun run check` (12 checks passed)

## Stacking / related work
Depends on #3599.

This cleanly replaces the typing-indicator portion of #3396 without changing that PR.

### DIFF
--- a/src/channels/accounts.ts
+++ b/src/channels/accounts.ts
@@ -66,6 +66,7 @@ const SNAKE_TO_CAMEL: Record<string, string> = {
   thread_policy_by_channel: "threadPolicyByChannel",
   transcribe_voice: "transcribeVoice",
   download_media: "downloadMedia",
+  waiting_behavior: "waitingBehavior",
 };
 
 let warnedAboutDualKeys = false;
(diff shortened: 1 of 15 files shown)


just-cameron:

PR #3606, opened 2026-07-31
feat(whatsapp): add outbound message prefix

just-cameron:
## Summary
- Adds a default-off WhatsApp `message_prefix` setting and carries it through account creation, updates, legacy YAML loading, live config changes, and snapshots.
- Applies the prefix at the final WhatsApp payload boundary so ordinary outbound `text` and media `caption` payloads follow the same rule without mutating the caller message.
- Keeps control/system paths unprefixed: approval prompts, lifecycle error replies, reactions, reaction removals, and captionless voice memo payloads are excluded unless a direct reply explicitly opts in.

## Before / after
Before this, WhatsApp ordinary outbound messages had no account-level prefix, and the earlier prefix implementation operated before media caption selection, so title-derived captions could be missed and system replies could be decorated accidentally.

After this, configuring `message_prefix: "[bot] "` sends ordinary replies like `[bot] hello`. For media, WhatsApp uses `text` when present, otherwise `title`; the selected non-empty image/video/document caption gets the prefix exactly once. Empty text/title fields still omit captions, and voice memo payloads remain captionless.

## Config persistence
The setting is accepted as `message_prefix`, normalized to `messagePrefix`, preserved on account merge, migrated from legacy YAML config, exposed in account/config snapshots, and passed through live config updates.

## Risks / limits
- Default behavior is unchanged when `message_prefix` is absent or empty.
- The prefix is a literal prepend, not a template system.
- Direct replies stay unprefixed by default so system/control messages remain clean; only explicit ordinary-direct-reply callers can opt in.
- This PR only changes WhatsApp prefix behavior. It does not change inbound routing, LID identity handling, or other channel adapters.

## Tests
- 25 focused WhatsApp prefix/media/control/lifecycle/config tests and adversarial payload probes for text-only, title-only, text+title, empty fields, document/image/video captions, voice memos, reactions/removals, repeat send exact-once behavior, and control/lifecycle exclusions.
- `bun run typecheck`
- `bun run check` passed all 12 checks.

## Stacking / related PRs
Depends on public #3599 (`maintainer/whatsapp-lid-identity-clean`).

This cleanly replaces the outbound-prefix portion of public #3396.

### DIFF
--- a/src/channels/accounts.ts
+++ b/src/channels/accounts.ts
@@ -57,6 +57,7 @@ const SNAKE_TO_CAMEL: Record<string, string> = {
   group_mode: "groupMode",
   inbound_debounce_ms: "inboundDebounceMs",
   listen_mode: "listenMode",
+  message_prefix: "messagePrefix",
   media_max_bytes: "mediaMaxBytes",
   mention_patterns: "mentionPatterns",
   recipient_aliases: "recipientAliases",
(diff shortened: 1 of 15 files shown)


just-cameron:

PR #3607, opened 2026-07-31
fix(cli): omit unavailable memory block counts

overlord-letta:
## Summary
- only show an agent memory-block count when it is positive
- preserve singular/plural formatting while avoiding a misleading `0 memory blocks` label
- cover unavailable, empty, singular, and plural metadata with focused tests

## Why
Agent list responses can leave the optional blocks relationship unpopulated. The selector previously collapsed both unavailable metadata and a true empty relationship to zero, then presented that fallback as fact. Omitting non-positive counts keeps the compact selector truthful without adding an N+1 fetch.

## Risk
Display-only. Positive counts, model metadata, and local/shared selector behavior remain unchanged.

## Test plan
- [x] `bun run check`
- [x] `bun test src/cli/components/agent-selector-utils.test.ts`

### DIFF
--- a/src/cli/components/AgentSelector.tsx
+++ b/src/cli/components/AgentSelector.tsx
@@ -23,6 +23,7 @@ import {
   AGENT_SELECTOR_TAB_EMPTY_STATES,
   type AgentSelectorListAgent,
   type AgentSelectorTabId,
+  formatAgentMemoryBlockCount,
   formatAgentModel,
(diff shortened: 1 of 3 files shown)


kl2806 (approved):

frezbo:
Follow-up issue #3845 tracks the same misleading memory-count problem in the separate startup profile selector, including MemFS-aware rendering.

PR #3616, opened 2026-08-01
feat(memory): restore read-only file controls

sarahwooders:
## Summary

- restore pre-commit protection for memory files marked `read_only: true`, including edits, deletion, and renames
- add `letta memory read-only <path> <true|false>` as the narrow authorized path for toggling the protected field
- force every agent invocation of that command through an unpersistable `alwaysAsk` permission decision
- preserve existing frontmatter/body formatting and reject unsafe or out-of-scope paths

## Validation

- `bun test src/cli/subcommands/memory-read-only.spec.ts src/agent/memory-git.precommit.*.ts src/permissions/permissions-checker.*.ts src/permissions/permissions-mode.*.ts`
- `bun test src/tools/memory-tool.*.ts src/tools/memory-apply-patch.*.ts`
- `bun run check`

Generated with [Letta Code](https://letta.com)

### DIFF
--- a/src/permissions/mandatory-approval.ts
+++ b/src/permissions/mandatory-approval.ts
@@ -1,0 +1,37 @@
+import type { PermissionCheckResult } from "./types";
+
+type ToolArgs = Record<string, unknown>;
+
+export function envFlagEnabled(name: string): boolean {
+  const value = process.env[name];
+  if (!value) return false;
+  return value === "1" || value.toLowerCase() === "true";
+}
+
+function extractShellCommand(toolArgs: ToolArgs): string | null {
+  const command =
+    typeof toolArgs.cmd === "string" ? toolArgs.cmd : toolArgs.command;
+  if (typeof command === "string") return command;
+  return Array.isArray(command) ? command.join(" ") : null;
+}
+
+export function getMandatoryApproval(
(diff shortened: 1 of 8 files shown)


src/permissions/mandatory-approval.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function getMandatoryApproval(
+  canonicalTool: string,
+  toolArgs: ToolArgs,
+): PermissionCheckResult | null {
+  if (canonicalTool !== "Bash") return null;
+  const command = extractShellCommand(toolArgs);
+  if (
+    !command ||
+    /(?:^|[\s;&|'"])(?:[^\s;&|]*\/)?letta(?:\.js)?['"]?\s+(?:memory|memfs)\s+read-only(?:\s|$)/.exec(
```
amelia-letta, line 26: The `alwaysAsk` gate only matches `letta memory read-only` in the command string, but the pre-commit hook bypass is the env var `LETTA_APPROVED_READ_ONLY_CHANGE=1`. An agent in unrestricted mode can edit a read-only file then run `LETTA_APPROVED_READ_ONLY_CHANGE=1 git commit -m "..."` to bypass both this permission check and the pre-commit hook.

PR #3617, opened 2026-08-01
fix(listener): restore custom router registration

### LINKED ISSUE #3539: self hosted remotes, i.e. custom LETTA_BASE_URL broken due to #3460
#3460 adds a check `startupMode.kind === "remote"` which breaks custom LETTA_BASE_URL, since that mode uses `"unsupported-self-hosted"` rather than 'remote', which then does not go through the registration.

Documentation of self-hosted remotes is here: https://docs.letta.com/agent-sdk/remote-client/self-hosted

cpacker:
## Summary

- restore custom environment-server registration when the existing self-hosted listener override is explicitly enabled
- preserve the standalone listener ownership guard added in #3460
- add a regression test that verifies the listener reaches a custom `/v1/environments/register` endpoint
- pair with [letta-docs#136](https://github.com/letta-ai/letta-docs/pull/136), which restores the BYOR architecture and setup guide
- fixes #3539

### DIFF
--- a/src/cli/subcommands/listen-single-instance.test.ts
+++ b/src/cli/subcommands/listen-single-instance.test.ts
@@ -1,5 +1,7 @@
 import { afterEach, beforeEach, describe, expect, mock, test } from "bun:test";
 import { mkdtemp, rm } from "node:fs/promises";
+import { createServer } from "node:http";
+import type { AddressInfo } from "node:net";
 import { tmpdir } from "node:os";
 import path from "node:path";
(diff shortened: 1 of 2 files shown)


cpacker:
Closing after tracing the feature provenance. The generic self-hosted override originated as an internal Desktop escape hatch, not a supported custom-router product surface; restoring it would turn accidental behavior into a public commitment. The supported paths are App Server for self-hosted clients and BYOM for hosted remote environments. Follow-up tracked in LET-10476.

PR #3653, opened 2026-08-03
feat(cli): add letta shared-memory subcommand and make the shared memory skill git-first

cpacker:
## Summary

Fixes LET-10545 (management-plane half). Agents were managing shared memory through raw curl calls because the `managing-shared-memory` skill documented only the REST API — and that path dead-ends in environments where `LETTA_API_KEY` is not exported in the shell (OAuth keychain auth). The git plumbing already exists end-to-end: shared repositories are served over the same smart-HTTP transport as MemFS and mount at `$MEMORY_DIR/../<name>` as normal git checkouts. Agents just were never told, and mounts did not materialize on mid-session attaches.

This makes the data plane plain git (matching MemFS) and adds a management-plane subcommand:

- **`letta shared-memory list | create | attach | detach | sync | history`** — uses harness auth (works with keychain OAuth, no env key needed) and inherits the agent id from `AGENT_ID`, following the existing self-management subcommand pattern (`letta messages search`, `letta memory pull`).
- **`attach`** performs the full sequence agents previously fumbled: API attach → poll until the async attachment is visible → clone the local mount (`syncAttachedAgentRepositories`) → recompile the system prompt so the projection matches disk. Accepts a repository name or id. Exits non-zero if the mount did not materialize, and surfaces recompile failures as `recompile_failed` instead of swallowing them.
- **`sync`** self-heals missing mounts — the "system prompt references `$MEMORY_DIR/../<name>` but it is not on disk" state that occurs when a repository is attached from another surface while a session/sandbox is already running.
- **SKILL.md rewritten git-first**: leads with the mount + git workflow (edit files, `git commit`, `git push`, `git pull --rebase`), uses the subcommand for management, and includes a troubleshooting section for the failure modes observed in the wild. The curl recipes are reduced to an SDK/API footnote.

The subcommand is guarded against the local backend (shared memory requires hosted agents, consistent with the existing skill gating).

**Split out:** the credential-helper reset in `memory-git.ts` (the fix for the macOS keychain preempting the repo-local helper — the HTTP 500s on plain `git pull`) now lives on `fix/memfs-credential-helper-reset` and will land as its own PR, so the auth approach can be reviewed and reverted independently. Its two tests were also the Windows CI failures on this branch. Until that lands, plain-git push/pull in mounts has the same auth behavior as MemFS today.

Verified: 20 unit tests on the subcommand (attach polling, name/id resolution, env inheritance, mount/recompile failure reporting), `bun run check` green, and live read-only smoke tests (`list`, `history`) against the production API using keychain auth with no `LETTA_API_KEY` in the environment.

Follow-ups (separate PRs): the credential-helper reset above, and extending post-turn auto-push + dirty reminders to attached repository mounts if forgotten pushes show up in practice.


cpacker:
Pushed de4bd1ce addressing the live-test findings from the retest:

1. **Inherited credential helper preemption (the plain-git 500s)** — git accumulates credential helpers across config scopes, and a system/global helper (macOS osxkeychain) holding a stale credential for `api.letta.com` answered before the repo-local helper, so plain `git pull` authenticated as the wrong identity. Harness-driven git never hit this because it passes the auth header inline with helpers disabled — which is why MemFS looked fine while shared mounts (the first surface where agents run plain git themselves) failed. Fix: the repo-local config now writes an **empty helper entry before the inline helper, scoped to the Letta remote URL only** — the empty entry resets git's accumulated helper list for that URL, so inherited helpers can never answer for the Letta host, while user helpers for all other hosts (e.g. `/memory-repository` GitHub mirrors) are untouched. Applies to both MemFS and shared mounts (shared code path). Verified with a `git credential fill` test against a marker-writing fake global helper: the fake helper is never invoked and the fresh token wins.

2. **Stale non-git directory at the mount path** — `sync` now moves it aside to `<name>.pre-mount-<timestamp>` (preserving possibly un-synced work) and clones the real mount, reporting the moved path in the sync summary instead of erroring.

Also added to the skill troubleshooting: both cases, plus an explicit "never hand-clone the repo to /tmp as a workaround" rule.

Not addressed here (follow-ups): a dynamic `letta credential` helper that resolves harness auth per git operation instead of persisting a token snapshot (same tradeoff MemFS already makes today — token refreshed at every sync point), and the server returning 500 instead of 403/404 for wrong-identity git requests (letta-cloud fix).

cpacker:
Reverted the mount move-aside auto-heal (59845237) per review — silently renaming a directory the harness does not understand is implicit mutation of anomalous state, the opposite of the explicit-error model. The collision now stays a hard error, extended with the resolution path ("Move or delete that directory, then re-run the sync"), the directory is left untouched for the agent to inspect, and the skill troubleshooting documents the same. The credential-helper reset fix from de4bd1ce is unchanged.

Issue #3647, opened 2026-08-03
App Server SSE stream stalls during silent tool execution — no heartbeat causes VS Code and Open WebUI timeouts


KShepherd-Kustoms:
### AI Disclosure

- [ ] This issue was written entirely by a human
- [x] This issue was written with AI assistance and **reviewed and edited by a human**
- [x] I have read the [AI Policy](https://github.com/letta-ai/letta-code/blob/main/AI_POLICY.md) and agree to its terms

### AI Tool(s) Used

Letta Code

### Human Verification

I have personally reproduced or verified this issue, reviewed its contents, and take responsibility for its accuracy.

### Describe the bug

When using the Letta App Server's OpenAI-compatible API (`/v1/chat/completions` and `/v1/responses`) from VS Code Copilot Chat or Open WebUI, **tool-heavy turns hang** because the App Server's SSE stream does not emit any data during long silent tool/subagent execution intervals.

**What I expected:** The SSE stream would emit periodic heartbeat bytes or events while the agent is executing tools, keeping the HTTP connection alive and the client informed that work is progressing.

**What happened:** 
- **VS Code Copilot Chat:** Terminates after ~5 minutes with `BodyTimeoutError` (Undici HTTP body inactivity timeout). The socket is closed because no data arrived during the silent tool execution.
- **Open WebUI:** Hangs indefinitely until manually cancelled. Open WebUI's SSE client has no body timeout, so it waits forever for data that never comes.

Both clients time out due to the **same root cause**: the App Server stops emitting SSE events while tools are running, and the clients interpret this silence as a dropped connection.

### Letta Code version

0.30.2

### Where did this happen?

App Server / SDK

### Operating system

Ubuntu

### Steps to reproduce

1. Start the App Server with `letta server --openai-api` (or via Docker compose with `--openai-api` flag)
2. In VS Code: `Chat: Manage Language Models` → add OpenAI-compatible provider pointing to App Server (`http://192.168.1.174:4500/v1/chat/completions`)
3. In Open WebUI: Add App Server as a custom model provider
4. Send a prompt that triggers tool execution (e.g., file editing, Bash commands, subagent calls)
5. Wait for the agent to execute tools (especially if tools take >5 minutes)
6. **Observe:** VS Code terminates with `BodyTimeoutError`. Open WebUI hangs indefinitely.

### Relevant logs, screenshots, or recordings

[...]

PR #3654, opened 2026-08-03
fix(app-server): execute tools without relay connection

just-cameron:

## Problem

The OpenAI-compatible App Server uses a process runtime transport. With no WebSocket client attached, that transport intentionally reports closed. The listener interpreted that as a relay disconnect and waited for a connection before executing every auto-approved tool, so even an instant shell command emitted a function call and then stalled forever.

This reproduces with direct `curl` against both `/v1/responses` and `/v1/chat/completions`; OpenWebUI is not involved.

## Fix

- mark HTTP bridge messages as explicitly process-owned
- let only those turns execute local tools without an attached listener connection
- preserve the existing reconnect wait for relay-originated turns
- add a lifecycle regression that fails at the blocking transport wait on `main`

The bypass is internal listener metadata. Remote protocol input is still mapped through the explicit message router and cannot set it.

## Validation

- [x] regression fails without the source fix and passes with it
- [x] 45 App Server, process-transport, and relay-reconnect tests pass
- [x] `bun run check` — all 12 checks pass
- [x] direct Responses smoke: `function_call` → `function_call_output` → `response.completed`
- [x] direct Chat Completions smoke: tool output reported → `[DONE]`
- [x] text-only Responses control still completes
- [ ] complete `bun test`: 5,828 passed / 27 skipped / 21 environment failures; failures were outside this diff and predominantly exhausted Cloud credits, plus unrelated global-settings test interference

## Scope

This fixes the hard stall where tool execution never begins. It does not add SSE heartbeats for genuinely long, quiet tools, so the separate client-inactivity concern in #3647 remains relevant.

### DIFF
--- a/src/websocket/app-server-openai-turn.ts
+++ b/src/websocket/app-server-openai-turn.ts
@@ -370,9 +370,9 @@ async function runTurnViaListenerRuntime(
       // would strand the second observer (only the first OTID survives a
       // merged batch) and answer the first with both prompts.
       noCoalesce: true,
-      // This request is owned by the in-process HTTP bridge. Local tools must
-      // not wait for a relay WebSocket connection that may never exist.
-      allowToolExecutionWithoutListenerConnection: true,
+      // This request's output is owned by the in-process HTTP bridge, so the
+      // turn must not block waiting for a relay WebSocket client to attach.
+      processOwnedTurn: true,
       // HTTP clients cannot surface Letta Code's interactive overlays. Keep
       // those tools out so the agent asks in ordinary assistant text.
       excludeInteractiveTools: true,
--- a/src/websocket/app-server-openai-turn.ts
+++ b/src/websocket/app-server-openai-turn.ts
@@ -370,6 +370,9 @@ async function runTurnViaListenerRuntime(
       // would strand the second observer (only the first OTID survives a
       // merged batch) and answer the first with both prompts.
       noCoalesce: true,
+      // This request is owned by the in-process HTTP bridge. Local tools must
+      // not wait for a relay WebSocket connection that may never exist.
+      allowToolExecutionWithoutListenerConnection: true,
       // HTTP clients cannot surface Letta Code's interactive overlays. Keep
       // those tools out so the agent asks in ordinary assistant text.
       excludeInteractiveTools: true,
--- a/src/websocket/listener/turn-approval.ts
+++ b/src/websocket/listener/turn-approval.ts
@@ -157,8 +157,8 @@ export async function handleApprovalStop(params: {
   pendingNormalizationInterruptedToolCallIds: string[];
   turnToolContextId: string | null;
   turnLease: TurnLease;
-  /** Process-owned turns can execute local tools without a relay connection. */
-  allowToolExecutionWithoutListenerConnection?: boolean;
+  /** This turn's output is owned by an in-process caller, not a relay client. */
+  processOwnedTurn?: boolean;
   buildSendOptions: () => Parameters<
     typeof sendApprovalContinuationWithRetry
   >[2];
@@ -185,7 +185,7 @@ export async function handleApprovalStop(params: {
     turnInput,
     turnToolContextId,
     turnLease,
-    allowToolExecutionWithoutListenerConnection = false,
+    processOwnedTurn = false,
     buildSendOptions,
     providerFallback,
     dependencies,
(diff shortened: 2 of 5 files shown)


cpacker (commented):
## Summary

I traced this end to end. **The diagnosis is correct and the fix is not redundant** — the concern that prompted this review was whether an existing mechanism already solved it, and I checked three candidates that don't.

I've pushed one follow-up commit (3d10d88) renaming the field. No behavior change.

## Root cause (confirmed)

`letta server --listen --openai-api` calls `setActiveRuntime(runtime)` at boot with a runtime that has **zero listener connections** (`app-server.ts:175,394`) — connections are only registered in `handleWebSocketConnection`. So:

- `ensureListenerRuntime()` (`app-server-openai-turn.ts:132`) sees a non-closed active runtime and returns early, never reaching `startLocalChannelListener`
- `getOrCreateProcessTransport()` wraps a connection-less runtime → `ProcessRuntimeTransport.isOpen()` iterates zero connections → `false`
- the #3522 reconnect gate in `turn-approval.ts:415` then waits forever for a WS client that will never exist

Worth noting for anyone reading later: process services and the cron scheduler bootstrap from `attachOpenListenerSocket`, so in the no-WS-client deployment they never start either. That bounds the blast radius to bridge-submitted turns.

## Why an existing mechanism doesn't cover this

- **`LocalListenerTransport`** (`transport.ts:26`) is the closest — it exists precisely for "no remote peer, still run the turn," `isOpen()` hardcoded `true`, `send()` discards. `startLocalChannelListener` registers it as an *initialized* connection (`lifecycle.ts:713,727` → `lifecycle.ts:399`), which is why the local-channel path never hit this bug. **But** it's runtime-scoped, and the app-server runtime also serves real WS clients — a permanently-open local connection would make `isOpen()` always true and silently defeat #3522 for those clients, dropping tool output into a discarding transport mid-reconnect.
- **`getListenerTransportKind(socket) === "websocket"`**, the idiom already used at 4 sites, would be *wrong here*: relay queued turns also run on the process transport (`lifecycle.ts:443-450` `scheduleQueuePump`), whose `kind` is `"runtime"`. Gating on kind regresses #3522 the same way.
- **`TurnOrigin`** is only `"message" | "approval_recovery"` — no submitter identity.

The decisive point: within one app-server runtime, output ownership varies **per turn** (HTTP request → in-process observers; WS client → that socket). So per-turn threading is the right altitude, and there was no existing per-turn carrier.

## What I changed

**Renamed the field to `processOwnedTurn`.** `allowToolExecutionWithoutListenerConnection` described the remedy rather than the condition, which reads as a safety bypass and invites misuse at future call sites. Naming the fact and letting `turn-approval.ts` derive the skip means the next transport-liveness gate can reuse the same fact instead of adding a second boolean.

**Added the inverse regression.** The original test only covered the bypass, so inverting the gate or flipping its default to `true` would not have failed anything. The new test asserts a relay-owned turn with a closed transport still waits for reconnect before executing tools.

## Propagation: I was wrong, it's already complete

I initially flagged two synthetic `IncomingMessage` sites as dropping the flag. Both turned out not to be gaps, so I did **not** make those changes:

- `recovery.ts:838` — reachable only from `control-inputs.ts:266`, i.e. a remote client sending `input.kind=approval_response`. A WS client is attached by construction there, and the bridge never sends control frames. Propagating `processOwnedTurn` would have been actively wrong.
- `queue.ts:162` — the notification-only branch only handles batches with no user message (cron/channel). Bridge messages always carry content, so they take the primary-item path at `queue.ts:206`, which preserves the flag via `{...template}`.

Within a single bridge turn the flag also survives multiple tool rounds, since `turn.ts:786` passes the same `msg` on each iteration.

## One gap I did not close

There's still no test covering the `app-server → turn.ts → handleApprovalStop` wiring end to end. Delete `turn.ts:801` or the `app-server-openai-turn.ts:375` assignment and the suite stays green. A real guard needs the OpenAI-compat endpoint driven against a live listener runtime with a stubbed executor — the existing `app-server-openai-*.test.ts` files stub `runTurn` via `__testSetRunTurnImpl` and so bypass the listener path entirely. I didn't build that harness here; flagging it rather than leaving it implied.

## What was already solid

- **Security claim verified.** `message-router.ts:392` builds `IncomingMessage` field-by-field with no spread, so remote protocol input genuinely cannot set this.
- **Default `false`** preserves relay behavior; #3522's `auth-lifecycle-approval-reconnect.test.ts` passes (17 tests across the approval/reconnect files).
- Single production call site; `as never` test casts match the file's existing convention.
- `bun run check` — all 12 checks pass on the pushed commit.

PR #3662, opened 2026-08-03
fix(cron): preserve the invoking computer by default

just-cameron:

## Summary

- Restore invocation locality for durable Cloud schedules by resolving and verifying the computer that runs `letta cron add`, then persisting it as `target_device_id`.
- Keep the explicit precedence: `--runner local`, `--computer <id>`, `--runner cloud`, then the verified active computer.
- Fail without creating a schedule when the active computer or durable Cloud schedule route cannot be verified. The error names the three explicit choices.
- Update bundled scheduling guidance to match the runtime behavior.

## Behavior

Before, an untargeted Cloud-agent schedule ran in the managed Cloud sandbox even when it was created from a connected computer. After this change, the default durable schedule targets the verified invoking computer. Explicit `--runner cloud` remains the opt-in for an untargeted managed-sandbox schedule.

Existing targetless schedules keep their current behavior. This PR does not change server dispatch, offline fallback, or listener acknowledgement semantics.

## Test plan

- [x] `runCronSubcommand()` tests through a real local HTTP boundary for registered default, runtime override, explicit Cloud, explicit computer, local runner, missing schedule routes, unregistered computer, verification failure, and Desktop-local-only identity
- [x] runner and payload unit tests
- [x] prompt and bundled-skill contract tests
- [x] `bun run check`
- [x] independent adversarial review of the committed diff

### DIFF
--- a/scripts/isolated-unit-tests.json
+++ b/scripts/isolated-unit-tests.json
@@ -26,6 +26,11 @@
       "reason": "Uses a top-level Bun module mock for backend search."
     },
     {
+      "path": "src/cli/subcommands/cron.test.ts",
+      "timeoutMs": 15000,
+      "reason": "Runs a real HTTP server while mutating the settings singleton, console methods, process environment, and local cron storage."
+    },
+    {
       "path": "src/hooks/prompt-executor.test.ts",
       "timeoutMs": 15000,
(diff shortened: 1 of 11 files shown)


just-cameron:

Superseded by #3665, which is the canonical implementation PR.

PR #3665, opened 2026-08-04
fix(cron): preserve active listener execution by default

overlord-letta:
## Summary
- Restore the original execution locality for implicit Cloud-agent schedules: keep the timer durable in Cloud, but persist the active registered listener as the preferred target.
- Carry the listener device identity through tool runtime context so a child `letta cron` command cannot select a stale installation ID.
- Keep explicit choices authoritative: `--computer` selects a device, `--runner cloud` selects the Cloud sandbox, and `--runner local` stays process-local.
- Reject implicit targets that are offline, unregistered, synthetic, Desktop-local, or managed sandboxes instead of silently moving the work to Cloud.

## Before / after
Before, `letta cron add` without runner flags omitted `target_device_id`, so work created on a connected listener ran in the Cloud sandbox.

After, the CLI resolves the listener that owns the current turn, verifies that it is a live external environment, and stores it in the existing `target_device_id` field. Existing targetless schedules keep their historical Cloud-sandbox behavior.

## Limits
Managed sandboxes and Desktop local-proxy connections are not accepted by the current Cloud schedule target API. The CLI reports that limit and gives explicit alternatives instead of claiming local execution.

## Test plan
- [x] `bun test src/cli/subcommands/cron.test.ts src/cli/subcommands/cron-runner.test.ts src/backend/api/client.test.ts src/tools/shell-env.test.ts` (64 pass)
- [x] `bun run check` (12/12 checks pass)
- [x] Copied the HTTP-boundary regression onto the parent commit; the three default-target cases fail there and pass with this change.
- [x] Built the branch CLI, launched an isolated registered listener, created a real one-shot schedule with no runner or computer flag, and waited for it to fire. `letta cron runs` reported `status: success`, `delivery: device`, and the exact listener connection. The temporary schedule and listener were removed.
- [x] GitHub CI is fully green across all platform, headless, integration, build, lint, and review jobs.

### DIFF
--- a/src/agent/prompts/letta.md
+++ b/src/agent/prompts/letta.md
@@ -126,7 +126,7 @@ Creating crons:
 - Recurring monitoring/heartbeat: `letta cron add --name <short-name> --description <description> --prompt <future-message> --every "2h"` or `--cron "0 9 * * *"`
 Always include `--name`, `--description`, and `--prompt`. `$AGENT_ID` is automatically injected into the shell environment, and `letta cron` uses it by default, so you do not need to specify which agent to invoke unless overriding the current agent intentionally.
 
-Where crons run: for cloud agents, schedules default to durable Cloud schedules that fire from the cloud and execute in your cloud sandbox — they survive local shutdown, so this is the right default. If the scheduled work must run on a specific computer (e.g. it needs that computer's filesystem or local services), add `--computer <deviceId>` (from `letta environments list`) to keep the durable Cloud schedule but execute on that computer, with sandbox fallback if it is offline. Use `--runner local` only when that fallback is unacceptable; local schedules only fire while a Letta session is running on that computer.
+Where crons run: for cloud agents, the default is a durable Cloud schedule targeted to the registered external listener running the current turn. `--computer <deviceId>` overrides that target. Explicit `--runner cloud` runs in your Cloud sandbox instead. Managed sandboxes and Desktop-local proxy connections cannot currently be targeted. Use `--runner local` for process-local storage; it only fires while a Letta session runs on that computer.
 
 # Harness Architecture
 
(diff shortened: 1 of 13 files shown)


just-cameron:
https://linear.app/letta/issue/LET-10543/schedules-invoked-from-an-available-listener-execute-in-cloud-instead

cpacker:
Pushed 4c1ad53a: managed sandboxes no longer error on default `letta cron add` — they fall through to the untargeted schedule, which already fires in the agent's Cloud sandbox. That was the pre-existing behavior and is the locality-preserving answer for that runtime, so requiring `--runner cloud` there just broke the most common no-flags path.

Behavior matrix for default `letta cron add` (cloud agent) after this commit:

| Runtime | Result |
|---|---|
| Registered external listener (VPS, `letta server`) | Cloud schedule targeted to that listener |
| Managed cloud sandbox (`sandbox-*` runtime id) | untargeted Cloud schedule → fires in the sandbox (unchanged from main) |
| Desktop-local / unregistered / offline runtime | error with `--runner cloud` / `--computer` / `--runner local` guidance |

Implementation: `validateInferredTargetDevice` became `resolveInferredTargetDevice`, returning `device` / `cloud-sandbox` / `error`. The `sandbox-*` check runs before the environments-registry lookup (and skips it entirely) because sandbox rows can be registered and online in the registry, but individual sandboxes get retired and recreated, so pinning one as a device target would be wrong. Prompts, `--help`, and the scheduling skill updated to match; HTTP-boundary test added asserting the sandbox path POSTs without `target_device_id` and never hits `/v1/environments/`.

Known remaining gap (follow-up, not this PR): desktop-local sessions still error on the default. They exist only in the desktop's local proxy, not the Cloud environments registry, so Cloud dispatch has no wire to deliver on — making them targetable requires registering the desktop listener in the environments registry first.

PR #3675, opened 2026-08-04
docs(skills): clarify cron clock and execution targets

just-cameron:
## Summary
- Separate cron clock location from scheduled-message execution targeting.
- Document invoking-computer defaults, explicit computer selection, Cloud fallback, and the current CLI management commands.
- Correct recurring UTC and interval guidance while keeping the existing skill structure.

## Limits
- This documents current behavior. It does not change runtime routing, make Cloud fallback optional, or add in-place schedule editing.

## Test plan
- [x] `bun run check`

---
<a href="https://app.blacksmith.sh/letta-ai/codesmith/letta-code/pr/3675"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788476823&installation_model_id=430896&pr_number=3675&repository=letta-ai%2Fletta-code&return_to=https%3A%2F%2Fgithub.com%2Fletta-ai%2Fletta-code%2Fpull%2F3675&signature=fcd51888857686c789a040fea9014b57303e7de797a1b9c3a039e0f3b5954ee7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/src/skills/builtin/scheduling-tasks/SKILL.md
+++ b/src/skills/builtin/scheduling-tasks/SKILL.md
@@ -1,11 +1,11 @@
 ---
 name: scheduling-tasks
-description: Schedules reminders and recurring tasks via the letta cron CLI. Use when the user asks to be reminded of something, wants periodic messages, or needs to manage scheduled tasks.
+description: Schedules reminders and recurring tasks via the letta cron CLI. Use when the user asks to be reminded of something, wants periodic work or check-ins, or needs to list, inspect, replace, or cancel scheduled tasks.
 ---
 
 # Scheduling Tasks
 
-This skill lets you create, list, and manage scheduled tasks using the `letta cron` CLI. Scheduled tasks send a prompt to the agent on a timer — useful for reminders, periodic check-ins, and deferred follow-ups.
+This skill lets you create, list, and manage scheduled messages using the `letta cron` CLI. Scheduled messages send a prompt to the agent on a timer — useful for reminders, periodic check-ins, and deferred follow-ups.
 
 ## When to Use This Skill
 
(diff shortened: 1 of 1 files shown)


cpacker:
Superseding this with #3682 — thanks for the careful CLI-reference work here; the accurate corrections (`--once`, id-or-name `get`/`delete` + `remove` alias, `runs` history, `--every` compiling to a UTC cron expression, numeric-only cron fields, no in-place editing) are folded into that PR with co-author credit.

Why supersede rather than merge: after discussing the scheduling design with Charles, the direction is for agents to **omit** `--runner`/`--computer` and let the CLI choose placement, rather than teaching agents to reason about clock location vs execution target themselves. #3682 makes that guidance actually workable by changing the code: bare `letta cron add` on a desktop-local/unregistered runtime now falls back to a local schedule with a warning instead of erroring, so the "omit the flags" default works in every runtime. Two pieces of this doc's guidance also conflicted with that design:

- "Do not use `--runner local` only to target the current computer" — before the fallback lands, `--runner local` was the *only* way to keep a follow-up on the same execution environment as an active desktop conversation (avoiding cross-environment turn-queue races). After #3682 the default handles it, so the prohibition becomes moot rather than wrong.
- Rewriting example user requests to "9am UTC" — real users say local times; the skill now instructs converting local → UTC when writing Cloud `--cron` expressions (including the day-field shift), instead of changing what users ask for.

Closing once #3682 merges.

cpacker:
Closing — superseded by #3682, which has now merged (stacked on #3683).

What happened to this PR's content: the verified CLI-reference corrections (`--once`, id-or-name for `get`/`delete` + `remove` alias, `runs` history with local-only `--run-id`, `--every` compiling to a UTC cron expression on Cloud schedules, numeric-only cron fields, create-verify-delete replacement since there's no in-place edit) are all in the merged skill, with co-author credit on the commit. Thanks for the careful verification work — those gaps were real.

The guidance sections went a different direction after a design discussion with Charles: instead of documenting the clock-location vs execution-target split for agents to reason about, #3682 changed the CLI so bare `letta cron add` always has a working placement (including a local fallback for computers the cloud scheduler can't reach), and the skill now leads with "omit the flags." #3683 moved cron mechanics out of the system prompts entirely so the skill is the single source of truth.

PR #3712, opened 2026-08-06
feat(skills): add browser-use builtin skill

sarahwooders:
Adds a bundled `browser-use` skill that instructs agents on driving a real browser over the Chrome DevTools Protocol (navigate, click, type, wait, screenshots, video), including fallback behavior when no Chromium browser is installed.


cpacker:
Added in sandbox instead

PR #3716, opened 2026-08-07
fix(connect): validate API keys against the supplied base URL

### LINKED ISSUE #3381: letta connect openai-compatible: --base-url is not passed to checkProviderApiKey, always validates against api.openai.com
## Bug

`letta connect openai-compatible --base-url <URL> --api-key <KEY>` fails even with valid keys because the `checkProviderApiKey` call on line 449 of `src/cli/subcommands/connect.ts` does not forward the `baseURL` from `connectionOptions`.

The API key is validated against the default `https://api.openai.com/v1/models` instead of the user-supplied `--base-url`, so any non-OpenAI third-party key produces a `401`.

## To reproduce

```bash
# Using any OpenAI-compatible endpoint that requires a non-OpenAI key
export KEY="<any-valid-key-for-your-provider>"
export URL="http://localhost:8080/v1"          # e.g., llama.cpp or LM Studio

letta connect openai-compatible --base-url "$URL" --api-key "$KEY"
# → Failed to connect OpenAI-compatible API: API error (401):
#   "UNAUTHENTICATED: Failed to authenticate with OpenAI: Client error 401
#    Unauthorized for url https://api.openai.com/v1/models"
```

<details>
<summary>Full error</summary>

```
Failed to connect OpenAI-compatible API: API error (401): {"error":{"type":"llm_authentication","message":"Authentication failed with the LLM model provider.","detail":"UNAUTHENTICATED: Failed to authenticate with OpenAI: Client error '401 Unauthorized' for url 'https://api.openai.com/v1/models'\nFor more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401"}}
```

</details>

The provider and models ARE saved successfully despite the error (at least on the local backend), but the misleading error message leaves users in a confused state.

## Root cause

In `src/cli/subcommands/connect.ts`, the `ConnectSubcommandDeps.checkProviderApiKey` signature (lines 58-64) accepts `providerType`, `apiKey`, `accessKey`, `region`, `profile` but has no `baseURL` parameter:

```typescript
checkProviderApiKey: (
  providerType: string,
  apiKey: string,
  accessKey?: string,
  region?: string,
  profile?: string,
) => Promise<void>;
```

On line 449 the call is made without forwarding `connectionOptions`:

```typescript
// Bug: missing baseURL
await io.checkProviderApiKey(provider.byokProvider.providerType, apiKey, ...);

// connectionOptions IS collected from --base-url/--timeout (lines 275, 337, 411-417)
// AND IS correctly forwarded to createOrUpdateProvider on line 452-460,
// but skipped for checkProviderApiKey.
```

Compare with the **Bedrock** handler directly above (line 368) which does forward region and profile. And compare with the **TUI connect flow** in `connect-provider-service.ts` (line 308) which wraps via `checkProviderApiKey2` and passes `{ connection: options }` including `baseURL`.

## Suggested fix

Two changes needed:

1. Add an optional `baseURL` parameter to `ConnectSubcommandDeps.checkProviderApiKey`:
   ```typescript
   checkProviderApiKey: (
     providerType: string,
     apiKey: string,
     accessKey?: string,
     region?: string,
     profile?: string,
     baseURL?: string,       // ← add this
   ) => Promise<void>;
   ```

2. Forward `connectionOptions.baseURL` at the call site (line 449):
   ```typescript
   await io.checkProviderApiKey(
     provider.byokProvider.providerType,
     apiKey,
     undefined,
     undefined,
     undefined,
     connectionOptions.baseURL,  // ← add this
   );
   ```

The corresponding `api-provider.ts` implementation (`checkProviderApiKey` → `POST /v1/providers/check` with `base_url`) already accepts the parameter; it just isnt wired to the CLI subcommand.

## Impact

- Anyone using `letta connect openai-compatible --base-url <...>` on a non-OpenAI endpoint sees a red-herring 401 / "Failed to connect"
- Affects LM Studio, llama.cpp, Ollama Cloud, and any third party OpenAI-compatible endpoint that requires a different API key than the user&#39;s OpenAI key
- The error is especially misleading because the same flow through the **TUI/Desktop panel** (which uses `checkProviderApiKey2`) correctly handles baseURL

## Workaround

Use the TUI Desktop panel instead of the CLI subcommand. The provider and models are still saved to the local backend even after the error -- the issue is only in the validation step.

## Environment

- `letta` CLI 0.28.4 and 0.28.6 (affected)
- `openai-compatible` provider type, `--base-url` flag

SihyeonJeon:
## Summary

Fixes #3381.

`letta connect openai-compatible --base-url <URL> --api-key <KEY>` returns 401 from
`api.openai.com` even with a valid key. `ConnectSubcommandDeps.checkProviderApiKey` declares five
parameters where the real function takes six, so the call site dropped the connection options
without a type error and the base URL never reached the check. Everything below the CLI already
handles it.

Now forwards `connectionOptions`, branching on `hasConnectionOptions(...)` like the
`createOrUpdateProvider` call below it. Three existing tests asserted the two-argument call, so
they are updated; one new test covers the reported case.

`operationOptions.target` is left unset on purpose. `useLocalProviderStore()` falls back to
`defaultProviderStorageTarget()`, so passing it would change which store is used.

## Verification

```
$ bun run check
✓ 12 checks passed

$ bun test src/cli/subcommands/connect.test.ts
 22 pass  0 fail
```

With only the production file reverted to `origin/main`, four tests fail. The new one fails on the
missing options:

```
    "third-party-key",
-   undefined,
-   undefined,
-   undefined,
-   { "connection": { "baseURL": "http://localhost:8080/v1" } },
```

## AI Disclosure

## AI Tool(s) Used

Claude Code (Opus 5), used to locate the dropped parameter, write the fix and the test, and capture
the output above.

## Human Verification

### DIFF
--- a/src/cli/subcommands/connect.test.ts
+++ b/src/cli/subcommands/connect.test.ts
@@ -271,6 +271,15 @@ describe("connect subcommand", () => {
     expect(deps.checkProviderApiKey).toHaveBeenCalledWith(
       "lmstudio_openai",
       "not-needed",
+      undefined,
+      undefined,
+      undefined,
+      {
+        connection: {
+          baseURL: "http://127.0.0.1:1234/v1",
+          timeout: 600_000,
+        },
+      },
     );
     expect(deps.createOrUpdateProvider).toHaveBeenCalledWith(
       "lmstudio_openai",
@@ -283,6 +292,35 @@ describe("connect subcommand", () => {
         baseURL: "http://127.0.0.1:1234/v1",
         timeout: 600_000,
       },
+    );
+  });
+
+  // Regression for #3381: the API key was validated against the provider's
+  // default endpoint because --base-url never reached checkProviderApiKey, so
+  // any third-party key failed with a 401 from api.openai.com.
+  test("validates the API key against the supplied base URL", async () => {
+    const { deps } = createIoDeps();
+    setProviderTarget("api");
+
+    const exitCode = await runConnectSubcommand(
+      [
+        "openai-compatible",
+        "--base-url",
+        "http://localhost:8080/v1",
+        "--api-key",
+        "third-party-key",
+      ],
+      deps,
+    );
+
+    expect(exitCode).toBe(0);
+    expect(deps.checkProviderApiKey).toHaveBeenCalledWith(
+      "openai",
+      "third-party-key",
+      undefined,
+      undefined,
+      undefined,
+      { connection: { baseURL: "http://localhost:8080/v1" } },
     );
   });
 
@@ -313,6 +351,10 @@ describe("connect subcommand", () => {
     expect(deps.checkProviderApiKey).toHaveBeenCalledWith(
       "openai-compatible",
       "not-needed",
+      undefined,
+      undefined,
+      undefined,
+      { connection: { baseURL: "http://127.0.0.1:8000/v1/" } },
     );
     expect(deps.createOrUpdateProvider).toHaveBeenCalledWith(
       "openai-compatible",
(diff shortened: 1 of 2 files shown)


cpacker (approved):

cpacker:
Thanks!

PR #3757, opened 2026-08-09
feat(backend): hot-reload agent records from disk on mtime change

sich97:
## Summary

Conversation records in the local backend already hot-reload from disk when their on-disk mtime changes (`refreshConversationRecordFromStorage`). Agent records do not — they are loaded once at startup into an in-memory `Map` (`src/backend/local/local-store.ts:1143`) and every read returns the cached copy.

This makes it impossible to patch an agent's `model_settings` (e.g. `context_window_limit`, `max_tokens`) from an external script without restarting the entire server, which drops in-flight Slack/channel turns.

This PR mirrors the existing conversation hot-reload pattern for agent records:

- Add `agentRecordMtimeMsById` field to track each agent file's last seen mtime.
- Add `agentRecordFileMtimeMs(agentId)` helper to stat the agent JSON file on disk.
- Add `refreshAgentRecordFromStorage(agentId)` which returns the cached record on a fast-path mtime match, and otherwise re-reads and re-validates the file via `normalizeAgentRecord`.
- Hook `refreshAgentRecordFromStorage` into `retrieveAgentRecord` so every agent read picks up external edits transparently.
- Update `loadFromStorage` and `persistAgent` to record the mtime after initial load / Letta-internal writes, so self-writes don't trigger spurious re-reads on the next access.

The behavior is identical to the conversation hot-reload: when nothing on disk has changed, the call is a single `statSync` (no JSON parse). When something has changed, the file is re-read, normalized via the existing `normalizeAgentRecord` function, and replaces the cached copy.

**Failure modes / tradeoffs:**

- **Stat overhead:** every `retrieveAgentRecord` call now performs one `statSync`. On local NVMe this is sub-millisecond and well below the noise floor of an LLM turn, but it is technically a new syscall on a previously cache-only path.
- **Corrupt-file tolerance:** if the on-disk file becomes unparseable JSON, the cached in-memory record is preserved (the agent keeps working). The bad file is left in place — no auto-repair. This matches how `refreshConversationRecordFromStorage` handles parse failures.
- **Mtime resolution:** on filesystems with 1-second mtime resolution, two writes within the same second will look identical and the second will be ignored. The conversation refresh has the same property. For the realistic case (a human or watchdog script running `jq` to patch a file), this is a non-issue.
- **External tool responsibility:** scripts that patch agent files should write atomically (temp file + rename) so the refresh never observes a half-written file. `jq -i` and `python -c 'json.dump(...)'` followed by `os.replace()` both do this by default.

## Verification

- `git am 0001-local-store-agent-hot-reload.patch` applies cleanly against `letta-ai/letta-code@main` (commit `bddb818`) with no offsets or rejects.
- `tsc --noEmit --strict false --skipLibCheck --noResolve src/backend/local/local-store.ts` against the patched file reports no new errors in any of the lines I added (the unresolved-module errors are all pre-existing imports of `@letta-ai/letta-client` and the `@/*` aliases, which require the project's full dependency graph to resolve).
- New tests added in `src/backend/local/local-store-agent-hot-reload.test.ts` mirroring the existing `local-store-appledouble.test.ts` pattern.

## AI Disclosure

## AI Tool(s) Used

- Anthropic Claude (Sonnet 4.5) — used to read the Letta Code source on GitHub, identify the pattern used for conversation hot-reload, draft the implementation, draft the test file by mirroring `local-store-appledouble.test.ts`, draft this description, and verify the patch applies cleanly via `git am` against `main`. The submitter (human) reviewed the diff and the AI disclosure.

## Human Verification

### DIFF
--- a/0001-local-store-agent-hot-reload.patch
+++ b/0001-local-store-agent-hot-reload.patch
@@ -1,0 +1,129 @@
+From 078924812cd74d8f7ce0fb2a555d31f31f3ade2c Mon Sep 17 00:00:00 2001
+From: Simon <sich97@users.noreply.github.com>
+Date: Sun, 9 Aug 2026 10:43:03 +0200
+Subject: [PATCH] local: hot-reload agent records from disk on mtime change
+MIME-Version: 1.0
+Content-Type: text/plain; charset=UTF-8
+Content-Transfer-Encoding: 8bit
+
+Conversation records in the local backend already hot-reload from disk
+when their on-disk mtime changes (see refreshConversationRecordFromStorage).
+Agent records do not — they are loaded once at startup into an in-memory
+Map (local-store.ts) and every read returns the cached copy.
+
+This makes it impossible to patch an agent's model_settings (e.g.
+context_window_limit, max_tokens) from an external script without
+restarting the entire server, which drops in-flight Slack/channel turns.
+
+This commit mirrors the conversation hot-reload pattern for agent records:
+
+  - Add agentRecordMtimeMsById field to track each agent file's last
+    seen mtime.
+  - Add agentRecordFileMtimeMs(agentId) helper to stat the agent JSON
+    file on disk.
+  - Add refreshAgentRecordFromStorage(agentId) which returns the cached
+    record on a fast-path mtime match, and otherwise re-reads and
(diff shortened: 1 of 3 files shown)


cpacker:
Thanks for the PR!

PR #3779, opened 2026-08-10
feat: propagate super run correlation through listener turns

4shub:
## Summary

- Propagate optional `super_run_id` from cloud input frames through listener turns and Node core request metadata.
- Keep correlation owned by the turn lifecycle so status, approval, recovery, and terminal frames retain the correct super run without leaking into replacement turns.
- Keep different super runs separate during queue steering and outbound status coalescing while preserving legacy clients that omit the field.

### DIFF
--- a/src/websocket/listener/recovery.ts
+++ b/src/websocket/listener/recovery.ts
@@ -407,6 +407,7 @@ export async function resolveRecoveredApprovalResponse(
       connectionId: string,
     ) => void;
     connectionId?: string;
+    superRunId?: string;
     dependencies?: {
       applySuggestedPermissions?: typeof applySuggestedPermissionsForApproval;
       classifyApprovals?: typeof classifyApprovalsWithSuggestions;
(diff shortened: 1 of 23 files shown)


src/websocket/listener/recovery.ts
```diff
@@ -807,20 +813,24 @@ export async function resolveRecoveredApprovalResponse(
...
+        error:
+          stopReason === "error"
+            ? getTranscriptLoopErrorMessage({
+                error,
+                message: error instanceof Error ? error.message : String(error),
+              })
+            : undefined,
+      });
+    } finally {
+      runtime.turnLifecycle.releaseSuperRunId(recoveryLease);
```
amelia-letta, line 832: Error path releases here, but the success path returns without calling `releaseSuperRunId` — superRunId leaks after recovery finishes idle, and the next queued turn's `emitDequeuedUserMessage` would carry the stale correlation.

PR #3837, opened 2026-08-15
feat(mods): refresh conversation titles in the active TUI

just-cameron:

## Problem

A mod can update a conversation summary through the API, but the active TUI keeps the old summary in React state. The terminal tab therefore stays stale until the TUI reloads the conversation.

## Change

- Add `ctx.conversation.updateTitle(title)` to persist a title through the scoped mod conversation handle.
- Publish the title change after the backend update succeeds.
- Update the active TUI only when the changed conversation is open.
- Document the mod helper and test persistence, event delivery, active-conversation filtering, and cleanup.
- Isolate the listener reconnect suite because it mutates process-wide listener, settings, `HOME`, and environment state.

## Visual proof

I asked the agent to change the title, note that the change appears in the top bar:

image

## Validation

- `bun test src/mods/conversation-handle.test.ts src/cli/app/conversation-title-sync.test.ts` — 8 passed
- `bun test src/websocket/listener/auth-lifecycle-approval-reconnect.test.ts` — 2 passed


frezbo:
Follow-up issue #3843 tracks the complementary resume-path bug: persisted conversation titles are still not hydrated into the active TUI when an existing conversation is resumed or switched to.

PR #3860, opened 2026-08-16
fix(tools): emit tool_end for text-only structured tool results

### LINKED ISSUE #3828: `tool_end` is skipped for text-only structured tool results such as Bash
### AI Disclosure

### AI Tool(s) Used

Letta Code (Ezra support bot and user operated Letta Agent)

### Human Verification

### Describe the bug

A mod subscribed to both tool_start and tool_end lifecycle events gets `tool_start` on a successful local Bash call, but `tool_end` is never emitted — even though the result is returned normally and contains only text. The failure is silent: no error is raised, and the tool result is unaffected. 

### Expected behavior: 
`tool_end` fires after a successful local Bash call, consistent with the documented lifecycle contract for subscribed mods, and with only multimodal/image results bypassing string result rewriting.

### Actual behavior:
` tool_start` fires and the Bash result is returned, but `tool_end` is skipped.

### Letta Code version

0.29.12 

### Where did this happen?

Letta Code Desktop

### Operating system

Windows 11

### Reproduction Steps:
1. Add a valid global mod under `~/.letta/mods/` that writes separate marker files from the `tool_start` and `tool_end `events.
2. Run `/reload` and confirm the activation report shows `events.tools: true`.
3. Select the Desktop Local environment.
4. Run a simple foreground Bash command such as `echo tool-end-repro`.
5. Observe that the `tool_start` marker file is written but the `tool_end` marker file is not.
6. As a control, repeat with a string-returning tool such as `Read` — its `tool_end` is expected to fire.

_No response_

### Additional context

Potentially relevant source locations identified during AI-assisted investigation are `src/tools/impl/bash.ts` and the `tool_end` dispatch path in `src/tools/manager.ts`. These are included as leads only; the report does not assert the implementation root cause.

During AI-assisted investigation, Ezra reported reproducing the same event split with a temporary unit-level test against public main at commit 77ddc4eb (package version 0.30.19), though I have not personally reproduced this on a 0.30.19 Desktop build.

feiiiiii5:
## Summary

Fixes #3828: `tool_end` was silently skipped for text-only structured tool results.

**Root cause:** `executeTool` in `src/tools/manager.ts` gated the mod `tool_end` event on `typeof res.toolReturn === "string"`. Tools like Bash can return structured content parts (an array of `TextContent | ImageContent`), so even purely textual results (e.g. Bash output) never emitted `tool_end`, breaking mod lifecycle observers.

**Fix:** extract plain text via a new `toolReturnText()` helper (`src/tools/tool-return-text.ts`): plain strings pass through verbatim, text-only part arrays are concatenated, and returns containing any non-text (e.g. image) part still skip the event. `executeTool` emits the extracted text as the event output.

## Verification

- New regression tests in `src/tools/tool-end-event.test.ts`: Bash structured-text result emits `tool_end`; plain-string result (Read) still emits `tool_end`.
- Pre-fix: the Bash test fails (1 fail / 1 pass). Post-fix: 2/2 pass.
- `bun test $(find src/tools -name "*.test.ts" | grep -v integration-tests)`: 754 pass / 0 fail across 84 files.
- `bun run check`: 12/12 checks pass (biome, typescript, cycles, boundaries, source-file-size ratchet — `manager.ts` stays at its 2,994-line baseline).

## AI Disclosure

## AI Tool(s) Used

DeepSeek (AI-assisted development) operating on behalf of the submitting human. Extent: proposed the fix approach and test cases; the implementation was written and verified step-by-step in the human's environment (`bun test`, `bun run check`) and reviewed line-by-line by the submitter.

## Human Verification

### DIFF
--- a/src/tools/manager.ts
+++ b/src/tools/manager.ts
@@ -79,6 +79,7 @@ import {
 } from "./secret-substitution";
 import { TOOL_DEFINITIONS, type ToolName } from "./tool-definitions";
 import { TOOL_PERMISSIONS } from "./tool-permissions";
+import { toolReturnText } from "./tool-return-text";
 
 export const TOOL_NAMES = Object.keys(TOOL_DEFINITIONS) as ToolName[];
 
@@ -2861,9 +2862,8 @@ export async function executeTool(
     ? getExecutionContextById(options.toolContextId)
     : undefined;
   const modEvents = context?.modEvents;
-  if (!modEvents || typeof res.toolReturn !== "string") {
-    return res;
-  }
+  const textOutput = toolReturnText(res.toolReturn);
+  if (!modEvents || textOutput === null) return res;
 
   const executionScope = context?.runtimeContext
     ? buildExecutionRuntimeContextSnapshot({
@@ -2890,7 +2890,7 @@ export async function executeTool(
     toolCallId: options?.toolCallId,
     toolName: name,
     status: res.status,
-    output: res.toolReturn,
+    output: textOutput,
   });
 
   return override
--- a/src/tools/tool-end-event.test.ts
+++ b/src/tools/tool-end-event.test.ts
@@ -1,0 +1,62 @@
+import { afterAll, describe, expect, test } from "bun:test";
+import {
+  clearTools,
+  executeTool,
+  loadSpecificTools,
+  prepareCurrentToolExecutionContext,
+} from "@/tools/manager";
+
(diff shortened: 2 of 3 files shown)


feiiiiii5:
Closing as part of a contribution-portfolio consolidation: my effort is now focused on a single project (failroute, a static analyzer for silent-failure anti-patterns) and its upstream follow-through. Happy to revisit if there is interest.

Issue #3802, opened 2026-08-12
Cloud BYOK OpenAI-compatible providers ignore model token-limit metadata


whyyagswhy:
### AI Disclosure

- [ ] This issue was written entirely by a human
- [x] This issue was written with AI assistance and **reviewed and edited by a human**
- [x] I have read the [AI Policy](https://github.com/letta-ai/letta-code/blob/main/AI_POLICY.md) and agree to its terms

### AI Tool(s) Used

Letta Code

### Human Verification

I have personally reproduced or verified this issue, reviewed its contents, and take responsibility for its accuracy.

### Describe the bug

Letta Cloud does not preserve token-limit metadata advertised by OpenAI-compatible BYOK providers when synchronizing their model inventories. Models that are not recognized by Letta's static registry receive fallback or heuristic limits even when the provider's OpenAI-compatible `GET /v1/models` response includes `max_input_tokens` and `max_output_tokens`.

This is reproducible with a LiteLLM 1.90+ gateway. LiteLLM exposes these optional fields directly from `GET /v1/models`; this report does not require or propose a generic request to LiteLLM's non-standard `/model/info` endpoint.

Live Letta Cloud evidence captured on 2026-08-10 from `GET https://api.letta.com/v1/models?limit=1000` after refreshing the BYOK provider:

| Handle | Letta max_context_window | Letta max_tokens | Provider-advertised |
|---|---|---|---|
| `LiteLLM/claude-opus-5` | 30000 | 16384 | 1,000,000 / 128,000 |
| `LiteLLM/claude-sonnet-5` | 30000 | 16384 | 1,000,000 / 128,000 |
| `LiteLLM/gpt-5.6-luna` | 1050000 | 128000 | 272,000 / 128,000 |
| `LiteLLM/gpt-5.6-sol` | 1050000 | 128000 | 272,000 / 128,000 |
| `LiteLLM/gpt-5.6-terra` | 1050000 | 128000 | 272,000 / 128,000 |

So Letta returns both undersized fallback values (30,000/16,384 for unknown model names) and oversized heuristic values (1,050,000 for GPT-5.6 aliases whose actual input limit is 272,000).

Expected behavior: for OpenAI-compatible model discovery, valid optional provider metadata should be authoritative, with this precedence:

1. Positive integer `max_input_tokens` / `max_output_tokens` from the provider's standard OpenAI-compatible `GET /v1/models` response.
[...]

PR #3863, opened 2026-08-16
fix(providers): honor optional token-limit metadata from OpenAI-compatible model lists

feiiiiii5:
## Summary

The dynamic OpenAI-compatible provider (LETTA_LOCAL_BACKEND_EXPERIMENTAL / pi local endpoint) discarded the optional per-model limit metadata from GET /v1/models, so models behind OpenAI-compatible gateways (LiteLLM 1.90+, LM Studio, etc.) always got the harness fallback 128k/32k — even when the provider advertised accurate max_input_tokens / max_output_tokens. This is the local-backend analog of the bug class in #3802 (whose Cloud server half lives in the archived letta-ai/letta repo).

**Fix:** parse each /v1/models entry with defensive per-field validation (positive integers only; booleans, strings, zero, negatives, and NaN/Infinity are ignored), feed both limits to the shared model builder, and fall back per-field — last-known model value first, then the conservative harness defaults. Oversized windows remain clamped by the existing harness cap; provider refresh now updates both fields.

## Verification

- New tests in src/backend/dev/pi-openai-compatible-provider.test.ts (4 cases): advertised limits are published (30k/16,384 and 100k/40k accurate; 272k/128k clamped); malformed fields ignored (fallback 128k/32k); per-field fallback when only one limit is advertised; refresh updates limits when the provider changes them.
- Pre-fix: 3 of the 4 new tests fail against the previous implementation (1 pass / 3 fail); post-fix 4/4 pass.
- bun run check: 12/12 checks pass; sibling dev-provider tests 53/53 pass; local-backend + local-model-config tests 40/40 pass.

## AI Disclosure

## AI Tool(s) Used

DeepSeek (AI-assisted development) operating on behalf of the submitting human. Extent: proposed the fix approach and regression tests; the implementation was written and verified step-by-step in the human's environment (bun test, bun run check) and reviewed line-by-line by the submitter.

## Human Verification

### DIFF
--- a/src/backend/dev/pi-local-endpoint-provider.ts
+++ b/src/backend/dev/pi-local-endpoint-provider.ts
@@ -82,17 +82,53 @@ export function localEndpointOpenAIBaseURL(baseURL: string): string {
   return trimmed.endsWith("/v1") ? trimmed : `${trimmed}/v1`;
 }
 
-export function modelIdsFromOpenAICompatibleList(data: unknown): string[] {
+export interface OpenAICompatibleListEntry {
+  id: string;
+  /** Provider-advertised input limit from GET /v1/models, when valid. */
+  maxInputTokens?: number;
+  /** Provider-advertised output limit from GET /v1/models, when valid. */
+  maxOutputTokens?: number;
+}
+
+function isPositiveInteger(value: unknown): value is number {
+  return typeof value === "number" && Number.isInteger(value) && value > 0;
+}
+
+/**
+ * Parse an OpenAI-compatible GET /v1/models list, keeping each entry's
+ * optional limit metadata. Positive-integer limits are kept per field;
+ * malformed values (booleans, strings, zero, negatives, NaN/Infinity) are
+ * ignored so callers fall back per-field to last-known or harness defaults.
+ */
+export function openAICompatibleListEntries(
+  data: unknown,
+): OpenAICompatibleListEntry[] {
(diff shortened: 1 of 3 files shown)


feiiiiii5:
Implemented in 567ccad — sidecar it is:

- **No `Model` type changes.** Advertised-vs-effective limits now live in a module-scoped sidecar keyed by provider+model (`openAICompatibleTokenLimitRecord(providerId, modelId)`), with each successful refresh replacing the provider's whole slice so records for delisted models disappear.
- **Clamp stays explicit policy.** The conservative ceiling remains a `buildModel` decision; the sidecar reads effective values back *off the built Model*, so the record can never drift from what turn requests actually use, and no clamp arithmetic is duplicated.
- **Both values exposed.** A record pairs `advertisedContextLength`/`advertisedMaxOutputTokens` with `effectiveContextWindow`/`effectiveMaxTokens`, so a caller can tell "endpoint reported 272k" from "harness capped at 128k" (regression-tested exactly that case, plus unadvertised-fields-stay-undefined and stale-record eviction on refresh).

Sidecar regression tests: 7/7 passing; full `bun run check` 12/12 including typescript and the file-size ratchet.

feiiiiii5:
Closing as part of a contribution-portfolio consolidation: my effort is now focused on a single project (failroute, a static analyzer for silent-failure anti-patterns) and its upstream follow-through. Happy to revisit if there is interest.

PR #3908, opened 2026-08-19
fix(channels): isolate restored startup failures

kianjones9:
## Summary
- authenticate Slack before constructing Bolt and seed the verified bot identity, preventing Bolt’s eager duplicate `auth.test` promise from rejecting unobserved on `account_inactive`
- keep automatic Desktop channel restore failures isolated to the channel gateway instead of terminating the parent listener and disabling unrelated message/model APIs
- preserve fatal exit behavior for explicit `--channels` sessions

Closes [LET-11154](https://linear.app/letta/issue/LET-11154/prevent-channel-startup-failures-from-crashing-code-desktop-runtime).

## Test plan
- [x] `bun test src/channels/slack/adapter.test.ts src/channels/registry-lifecycle.test.ts src/channels/gateway-supervisor.test.ts`
- [x] `bun run check`
- [x] pre-commit hook stack
- [x] reproduced with the inactive local Slack account: startup returned `account_inactive` while the process remained alive

### DIFF
--- a/src/channels/slack/adapter.ts
+++ b/src/channels/slack/adapter.ts
@@ -62,6 +62,12 @@ export interface SlackChannelAdapter extends ChannelAdapter {
   }): Promise<ChannelMessageAttachment>;
 }
 
+type SlackAuthClient = {
+  auth: {
+    test(): Promise<{ team?: string; user_id?: string; bot_id?: string }>;
+  };
+};
+
 export function createSlackAdapter(
   config: SlackChannelAccount,
 ): SlackChannelAdapter {
@@ -71,6 +77,7 @@ export function createSlackAdapter(
   let running = false;
   let botUserId: string | null = null;
   let botId: string | null = null;
+  let workspaceName: string | null = null;
   let adapter: SlackChannelAdapter;
 
   const agentThreadTracker: AgentThreadTracker = createAgentThreadTracker();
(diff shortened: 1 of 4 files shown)


src/channels/slack/adapter.ts
```diff
@@ -100,12 +107,24 @@ export function createSlackAdapter(
 
   async function ensureApp(): Promise<SlackApp> {
     if (app) return app;
+    const auth = await (
```
4shub, line 110: does this actually bound the app or does it just crash the app earlier?

src/channels/slack/adapter.ts
```diff
@@ -388,16 +407,10 @@ export function createSlackAdapter(
     async start(): Promise<void> {
       if (running) return;
       const slackApp = await ensureApp();
-      const auth = await slackApp.client.auth.test();
```
4shub, line 391: shouldnt we just trycatch this?

kianjones9: disclaimer: agent response

`new App({ token })` calls Bolt’s singleAuthorization(), which immediately runs and caches:
```ts
cachedAuthTestResult = client.auth.test({ token: authorization.botToken });
```
Then our `start()` separately called `slackApp.client.auth.test()` meaning there are two requests. The explicit one was caught by `initializeChannels`; Bolt’s cached one had no consumer yet, so `account_inactive` became an unhandled rejection and Bun killed the gateway. Passing verified botId + botUserId makes Bolt use `Promise.resolve(...)` instead of launching that hidden request.

4shub (approved):

PR #4006, opened 2026-08-24
fix(hooks): match model-facing tool aliases

### LINKED ISSUE #3150: PreToolUse hooks do not fire for Agent (Task) tool calls
## Summary

PreToolUse hooks registered with matcher `"Agent"` are never invoked when the Agent tool is used to spawn subagents. The hook script is not executed at all — not even the first line. Other hook events (SubagentStop, Stop) fire correctly for the same subagent invocations.

## Version

Letta Code 0.27.11 (Desktop app, macOS, local backend)

## Reproduction

1. Register a PreToolUse hook with matcher `"Agent"`:
```json
{
  "hooks": {
    "PreToolUse": [
      {
        "matcher": "Agent",
        "hooks": [
          {
            "type": "command",
            "command": "/path/to/hook.sh",
            "timeout": 5000
          }
        ]
      }
    ]
  }
}
```

2. Create a minimal hook script that logs to a file on the **first line**:
```bash
#!/usr/bin/env bash
echo "[$(date -u +%Y-%m-%dT%H:%M:%SZ)] HOOK FIRED" >> /tmp/pretooluse-test.log
exit 0
```

3. Dispatch a subagent using the Agent tool.

4. Check `/tmp/pretooluse-test.log` — **file does not exist**. The hook was never invoked.

5. As a control: SubagentStop hooks fire correctly for the same subagent dispatch (confirmed via separate log).

## Evidence

- Hook is correctly registered (verified via `show_config.py` — appears under PreToolUse with matcher=Agent)
- Hook script is executable and works when invoked manually (`echo '...' | bash hook.sh`)
- SubagentStop hooks fire correctly for the same subagent dispatch (confirmed in english-enforce.log)
- The hook script's first line (which writes to a debug file) never executes — the file is never created
- This means the hook is not being invoked at all, not that it is failing silently

## Expected Behavior

PreToolUse hooks with matcher `"Agent"` should fire before the Agent tool spawns a subagent, per the documentation.

## Actual Behavior

The hook is never invoked. The Agent tool appears to use a subprocess creation path that bypasses the standard `executeTool()` → `runPreToolUseHooks()` pipeline.

## Impact

This prevents any PreToolUse-based enforcement on subagent dispatch — model enforcement, cost control, safety gates, audit logging. Users cannot gate which models subagents use.

## Related

- Issue #1419 documents a similar bypass for interactive tools (ExitPlanMode, AskUserQuestion)
- The subagent documentation states the Agent tool creates "a subprocess of Letta Code" — this may be the code path that bypasses PreToolUse

## Environment

- Letta Code 0.27.11 (Desktop app)
- macOS (Darwin)
- Local backend

jibramnobhill:
## Problem

When a user configures a PreToolUse hook for the `Agent` tool, it never fires. The hook system matches against the internal tool name (`Task`), but the model sees and calls it as `Agent`. The same mismatch affects PostToolUse and PostToolUseFailure payloads â€” they report `Task` instead of the name the user configured.

This means cost gates, safety hooks, and audit rules set up for `Agent` silently don't run.

## Fix

Match built-in tool hooks against both the model-facing name (`Agent`) and the internal name (`Task`). Report the model-facing name in hook payloads. Each matcher runs once even if multiple aliases match it.

## Files changed

- `src/hooks/loader.ts` â€” resolve tool aliases when matching hooks
- `src/tools/manager.ts` â€” pass model-facing name to hook payloads (3 call sites)
- `src/hooks/index.ts` â€” export alias resolution
- `src/hooks/loader.test.ts` â€” updated matcher tests
- `src/hooks/tool-alias-integration.test.ts` â€” new integration test (real `executeTool(Agent)` path, legacy `Task` compatibility, blocking, error/exception paths)

## Compatibility

- Existing hooks matching `Task` still work (legacy compatibility preserved)
- No changes to shell launchers, bash tool, exec-command, or shell-command
- No new dependencies

## Attribution

Refresh of #3236 by @abhay-codes07. Production approach is theirs; commit retains Abhay Singh as co-author. Rebased on current main â€” old hooks-reference path no longer exists.

Fixes #3150. Supersedes #3236.

## Validation

- 38 focused tests passed
- `bun run check` â€” 12/12 checks passed
- Mutation check: reverting alias call sites makes lifecycle tests fail
- Windows-portable integration test verified

## Scope note

This is a cross-platform fix (affects all platforms, not just Windows). Identified during Windows/Desktop QA work. Ready for review by whoever owns the hooks/tooling layer.

## AI disclosure

Developed with Letta Code assistance. Independently verified before publication.

### DIFF
--- a/src/hooks/index.ts
+++ b/src/hooks/index.ts
@@ -29,11 +29,20 @@ export * from "./types";
 // ============================================================================
 
 /**
+ * Tool names may include aliases for the same call. The first name is the
+ * canonical, model-facing name reported to hook commands; every name
+ * participates in matcher resolution.
+ */
+function canonicalToolName(toolName: string | readonly string[]): string {
+  return typeof toolName === "string" ? toolName : (toolName[0] ?? "");
+}
+
+/**
  * Run PreToolUse hooks before a tool is executed
  * Can block the tool call by returning blocked: true
  */
 export async function runPreToolUseHooks(
-  toolName: string,
+  toolName: string | readonly string[],
   toolInput: Record<string, unknown>,
   toolCallId?: string,
   workingDirectory: string = process.cwd(),
@@ -51,7 +60,7 @@ export async function runPreToolUseHooks(
   const input: PreToolUseHookInput = {
     event_type: "PreToolUse",
     working_directory: workingDirectory,
-    tool_name: toolName,
+    tool_name: canonicalToolName(toolName),
     tool_input: toolInput,
     tool_call_id: toolCallId,
     agent_id: agentId,
@@ -66,7 +75,7 @@ export async function runPreToolUseHooks(
  * These run in parallel since they cannot block
  */
 export async function runPostToolUseHooks(
-  toolName: string,
+  toolName: string | readonly string[],
   toolInput: Record<string, unknown>,
   toolResult: { status: "success" | "error"; output?: string },
   toolCallId?: string,
@@ -87,7 +96,7 @@ export async function runPostToolUseHooks(
   const input: PostToolUseHookInput = {
     event_type: "PostToolUse",
     working_directory: workingDirectory,
-    tool_name: toolName,
+    tool_name: canonicalToolName(toolName),
     tool_input: toolInput,
     tool_call_id: toolCallId,
     tool_result: toolResult,
@@ -106,7 +115,7 @@ export async function runPostToolUseHooks(
  * Stderr from hooks with exit code 2 is fed back to the agent
  */
 export async function runPostToolUseFailureHooks(
-  toolName: string,
+  toolName: string | readonly string[],
   toolInput: Record<string, unknown>,
   errorMessage: string,
   errorType?: string,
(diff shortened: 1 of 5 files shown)


jibramnobhill:
## Scope note

This PR fixes a cross-platform bug (Agent/Task tool name matching in hooks) that affects all platforms, not just Windows. It was identified during Windows/Desktop QA work.

The fix is complete, tested (12/12 checks green, mutation-tested), and retains the original author @abhay-codes07 as co-author. Ready for review by whoever owns the hooks/tooling layer (src/hooks/, src/tools/manager.ts).

jibramnobhill:
## Closing — flagging as lead for hooks/tooling owners

Closing this PR. The bug is real and confirmed on current main: PreToolUse hooks configured for Agent don't fire because the hook system matches against the internal name Task. But this is a cross-platform issue outside our team's scope (Windows/Desktop QA), and we're not the right owners to defend the approach.

**What we found:**
- Bug confirmed on current main (Aug 24, 2026)
- Task and Agent are the same tool — Task is internal, Agent is model-facing
- Hooks matching Agent silently don't run; hooks matching Task do
- PostToolUse/PostToolUseFailure payloads report Task instead of Agent

**What's preserved:**
- Branch ix/agent-tool-hook-matching with the fix and tests remains available
- Original work by @abhay-codes07 in #3236
- Issue #3150 is still open

If the hooks/tooling owner wants to pick this up, the branch and tests are here. Happy to hand off context.

PR #4007, opened 2026-08-24
fix(hooks): preserve native exit codes in PowerShell command hooks

### LINKED ISSUE #4009: Windows: PowerShell doesn't propagate child process exit codes in hooks
## Problem

On Windows, PowerShell doesn't propagate `$LASTEXITCODE` from child processes. When a hook command runs `node script.js` and the script exits with code 2 (block), `powershell.exe` returns exit code 1 (error) instead. The hook executor treats blocking hooks as errors, so Windows users' PreToolUse hooks silently fail to block.

## Reproduction

```powershell
PS> powershell.exe -Command "node -e `"process.exit(2)`""; $LASTEXITCODE
1
```

Expected: exit code 2 (block). Actual: exit code 1 (error).

## Impact

Any Windows user with PreToolUse hooks configured for blocking (exit 2) will have those hooks silently treated as errors instead of blocks. Cost gates, safety hooks, and audit rules don't fire as intended.

jibramnobhill:
## Problem

PowerShell does not propagate a final native application's exit code, so Windows hooks can lose BLOCK status. A stale native exit must not turn a later cmdlet, language failure, or PowerShell script into a false block.

## Fix

Parse the hook command before execution and inspect its final direct operation. Propagate `$LASTEXITCODE` only when that operation resolves to a native application; PowerShell cmdlets, language expressions, control structures, dynamic expressions, and `.ps1` scripts retain PowerShell ERROR semantics unless the hook explicitly exits.

## Validation

28 focused Windows tests pass on PowerShell 7.5.2 and Windows PowerShell 5.1; the full 12-check suite passes. Mutations prove the final-operation selection and application-versus-script boundary are enforced.

## Tracking

Fixes #4009

## Attribution

Developed with Letta Code assistance and independently verified before publication.


jibramnobhill:
Addressed the requested stale-`$LASTEXITCODE` case in `ace57518`.

The wrapper now snapshots the prior PowerShell error and returns ERROR when the hook command produces a new PowerShell error before considering a native exit code. The exact native-exit-2 followed by failing `Get-Item -ErrorAction SilentlyContinue` regression now returns 1 instead of BLOCK.

Validation: 16 focused Windows tests, 12/12 full checks, and mutation proof that removing the stale-state guard restores the failure.

jibramnobhill:
Added direct hook-level proof in `ba8441dd`: the exact command sequence is executed through `executeCommandHook`, then asserts `HookExitCode.ERROR` and explicitly rejects `HookExitCode.BLOCK`.

A production-code mutation removing the stale-state guard makes that test fail with **Expected 1, Received 2**. Restored production code passes 17/17 focused tests.

jibramnobhill:
Added `a42965c1` to preserve the regression-prevention rationale next to the wrapper:

> PowerShell cmdlets update `$?` and `$Error` but leave `$LASTEXITCODE` from the last native process untouched. Track new PowerShell errors so a stale native exit 2 cannot turn an ordinary hook failure into a false BLOCK.

The follow-up commit history now separates the correction, direct prevention proof, and maintenance rationale.

jibramnobhill:
Windows CI classification for the final head:

- All three #4007-specific Windows regressions passed, including the executor-level ERROR-not-BLOCK assertion.
- The job failed because the unrelated `Bash tool > times out long-running command` exceeded its 2,000 ms test deadline by 15 ms (`2015ms`). The test and its default shell path are untouched by this PR; hook exit preservation is opt-in.
- The remaining suite reported 1,693 passes. I classified the 2.015s timeout as a timing flake and retried the failed Windows job only.

jibramnobhill:
Final CI is green: **21 pass / 0 fail**. The first Windows attempt hit an unrelated 2.015-second timeout in an untouched Bash test; all #4007 regressions passed in that attempt. Retrying only the classified flaky Windows job succeeded.

The implementation and requested regression are complete. The PR remains GitHub-blocked only because the bot's stale `CHANGES_REQUESTED` review on `f680867b` still needs re-review/dismissal against the corrected head `a42965c1`.

jibramnobhill:
Addressed both follow-up review cases in `36923921`.

`$Error` is no longer used to decide whether the final failure was native. The wrapper installs a scoped `PostCommandLookupAction`, records the last runtime-resolved command kind (including alias resolution), restores the previous action, and only propagates `$LASTEXITCODE` when the final resolved command is an application or external script.

Verified matrix:
- native 2 ? BLOCK
- native 2 then failing cmdlet with `SilentlyContinue` or `Ignore` ? ERROR
- earlier failing cmdlet then final native 2 ? BLOCK
- `$PSNativeCommandUseErrorActionPreference = $true` then native 2 ? BLOCK
- aliased native 2 ? BLOCK
- successful final cmdlet ? ALLOW
- explicit exit 7 ? 7

22/22 focused Windows tests and 12/12 full checks pass. Forcing every final command to cmdlet catches lost blocks (expected 2, received 1); forcing every final command to native catches false blocks (expected 1, received 2).

jibramnobhill:
The new Windows failure was substantive and PowerShell-7-specific, not a retry candidate. Reproduced locally with portable pwsh 7.5.2 and fixed in `e78bf0a9`.

PowerShell 7 performs nested command lookups while rendering `NativeCommandExitException`; those formatting lookups replaced the recorded final native application, downgrading a real BLOCK to ERROR. The wrapper now excludes lookups nested under native error rendering while continuing to record subsequent user hook statements.

Added the inverse combined regression:
`PSNativeCommandUseErrorActionPreference=true; native 2; failing cmdlet -ErrorAction Ignore` must still return ERROR, not inherit BLOCK.

Validation:
- Windows PowerShell 5.1: 24/24 focused tests pass
- PowerShell 7.5.2: 24/24 focused tests pass
- mutation removing the rendering filter: expected 2, received 1
- targeted Biome/typecheck and commit hooks pass
- full check reaches 11/12; its only failure is an unrelated existing Biome diagnostic in `src/agent/prompt-assets.test.ts`

Fresh CI is running on `e78bf0a9`.

jibramnobhill:
Replaced the unsound lookup tracking entirely in `e6a6849b`.

The wrapper now parses the original hook text and classifies its final direct AST operation before execution. It propagates `$LASTEXITCODE` only if that operation resolves to `CommandTypes.Application`; it no longer treats command lookup as operation completion, and `ExternalScript` is not considered native.

Added real-Windows coverage for:
- native 2 followed by `$null.NoSuchMethod()` ? ERROR 1, not BLOCK;
- final `.ps1` with exit 2 ? ERROR 1 under the implicit-propagation policy;
- explicit Windows PowerShell 5.1 native/language matrix;
- existing pwsh 7 native-error-preference, `Ignore`, aliases, cmdlets, and explicit exits.

Validation: 28/28 focused tests on pwsh 7.5.2, 28/28 on Windows PowerShell 5.1, 12/12 full checks. Selecting the first rather than final AST statement fails six regressions; treating `ExternalScript` as the native case fails the `.ps1` boundary test.

Deliberate boundary: implicit propagation applies only to a direct final application. Compound/control-flow/dynamic commands and PowerShell scripts can use explicit `exit 2` in the hook command when they intend BLOCK.

PR #4043, opened 2026-08-25
feat(channels): add opt-in automatic reply relay

overlord-letta:
## Summary

Channel accounts can opt into automatic assistant reply delivery. Existing accounts continue to require an explicit `MessageChannel` call.

Relay mode:
- is selected per channel account and uses the shared local adapter path for all channel types;
- reconstructs streamed fragments by stable message identity;
- sends each assistant message separately through the scoped `MessageChannel` path as soon as that message is finalized;
- preserves message order and delivery state across tools, approvals, and gateway handoffs;
- does not attach `MessageChannel` to relay-only turns, and exposes only tool-mode accounts in mixed scopes;
- suppresses matching explicit `send` or `send-rich` delivery if an in-flight or stale tool remains during a mode transition.

## Safety

Relay requires one routed destination. A new message identity, tool or approval boundary, or normal stop reason finalizes the current assistant message. The gateway does not send partial current text on retry, error, or cancellation, and it does not relay subagent, ambiguous, or incomplete recovered output. A turn with more than one routed destination does not relay automatically, including a turn that mixes relay-mode and tool-mode accounts; the agent must target `MessageChannel` explicitly. Relay-only turns cannot use `MessageChannel` actions such as reactions, files, or proactive sends; use tool mode when those actions are required.

Remote gateway hosts must implement the new relay hook. The local channel gateway implements it in this pull request. A later Desktop change will expose the account setting after this version ships.

First-party unbound-chat help now shows the CLI route command beside the Channels UI path.

## Validation

- `bun test src/channels`: 1,289 passed, 1 skipped
- `bun run check`: 12 checks passed
- `bun run build`: passed
- Hermetic Telegram gateway integration: a relay turn reaches the model without `MessageChannel`, relays separate messages around a tool, then attaches `MessageChannel` after the account switches to tool mode
- Mixed Slack/Telegram account test: relay applies only to the source account, tool mode stays explicit-only, a mixed-source turn does not broadcast, and the tool schema exposes only tool-mode accounts
- Live Telegram message boundaries: assistant text → tool → assistant text produced two separate Telegram messages in order
- Live Telegram relay: one plain assistant response produced one Telegram message with zero tool calls
- Live Telegram tool-mode control: plain assistant text with zero tool calls produced no Telegram message


just-cameron:
Questions:

- Have we tested this with any other channel/is it supported with all channels?
- What is the behavior for MessageChannel if relay mode is on, particularly if there are some channels with relay mode and some without?
- Can we guarantee no change in current default behavior, and that this is strictly opt-in?

overlord-letta:

- **Channel support:** The setting is on the shared channel-account type, and the local gateway sends through the common adapter/`MessageChannel` path. The implementation therefore applies to every local channel adapter. We have live and hermetic end-to-end coverage for Telegram, shared-path unit coverage for Slack, and a new mixed Slack/Telegram regression in `f55fa9d4`. We have not run live acceptance on every adapter. Remote gateway hosts do not gain relay until they implement the new relay hook.
- **`MessageChannel`:** It remains available in both modes. On a relay account, matching explicit `send`/`send-rich` text is deduplicated in either order. Reactions, files, proactive sends, and intentionally different messages still send normally. Mode is resolved from the source account for each single-destination turn. If one active turn has multiple routed destinations, including mixed relay/tool accounts, automatic relay is disabled and the model must target `MessageChannel` explicitly.
- **Default behavior:** Missing or invalid `reply_mode` normalizes to `tool`; legacy account construction also writes `tool`; and the negative control proves that plain assistant text is not sent in tool mode. Existing accounts therefore keep the current explicit-`MessageChannel` behavior unless an operator stores `reply_mode: relay`.

I added the mixed-account regression, reran all 1,288 channel tests plus `bun run check`, and updated the PR description with this scope.

overlord-letta:

Follow-up from Slack: `d28df9de` now withholds `MessageChannel` from relay-only turns and proactive relay-only runtimes. If a routed scope mixes modes, the schema includes only tool-mode accounts.

The Telegram App Server integration now proves the model receives zero external tools in relay mode, still gets both automatic messages around a tool boundary, then receives `MessageChannel` after the same account switches to tool mode. Relay-only agents therefore stay on the plain assistant-message contract; reactions, files, and proactive `MessageChannel` sends require tool mode.

PR #4083, opened 2026-08-27
test(windows): verify headless shell calls and results

jibramnobhill:
## Summary

The Windows headless lane asked the model to run three PowerShell-compatible commands, but it previously passed when the final response contained `BANANA`. That did not prove the commands ran.

This PR validates the `stream-json` records instead. It requires three separate `Bash` calls for:

1. `echo 'Hello from Windows'`
2. `echo 'Line1'; echo 'Line2'`
3. `git --version`

Each call must have its own successful `tool_return_message`, matched by `tool_call_id`; its normalized executable command must exactly equal the requested command; and its return must contain that command's expected output. Heredoc and `&&` syntax remain rejected. The final result must also be successful and contain `BANANA`.

This closes both false-pass cases from review. Output from one successful call can no longer satisfy a different command, and text inside a larger command or PowerShell comment cannot impersonate a required command. For example, `echo 'git version 9'; # git --version` is rejected because it never executes `git --version`.

The branch was rebuilt from current `main` with only the two intended commits. GitHub now displays exactly two changed files: `src/test-utils/headless-windows.ts` and its test. The focused behavior passed 7/7 locally. Current-head CI is complete: 19 successful checks, one intentional review skip, and no failures. Jibram's human verification under the repository AI policy is still required before ready.

## Verification

`bun test src/test-utils/headless-windows.test.ts` passed 7/7 focused cases. The fixtures bind each command to its own `tool_return_message` by `tool_call_id` and require exact normalized command equality. Reverting to aggregate output makes the cross-call-substitution case fail; reverting exact equality to substring matching makes the PowerShell-comment spoof case fail. The rebuilt current head passed all 19 substantive CI checks, including Windows x64, with one intentional review skip.

## Breadcrumbs

- Origin: #1632 introduced the Windows headless scenario and treated the final `BANANA` marker as sufficient proof that all three requested commands ran. History found no later PR introducing this false-pass behavior.
- Related: #4307 isolates the unrelated channel-route state that made the previous platform CI jobs fail.
- Letta conversation: [`default`](https://chat.letta.com/chat/agent-03c143ce-b161-4a7b-93e8-6a0d570cf5f4?conversation=default)

## AI Disclosure

## AI Tool(s) Used

Letta Code was used for implementation, focused tests, mutation validation, and pull-request drafting.

## Human Verification

Pending Jibram's human verification of current head `29b61135` before this draft is marked ready. Current-head CI is all green.

### DIFF
--- a/src/test-utils/headless-windows.test.ts
+++ b/src/test-utils/headless-windows.test.ts
@@ -1,0 +1,109 @@
+import { describe, expect, test } from "bun:test";
+import { validateWindowsScenarioOutput } from "./headless-windows";
+
+function line(value: Record<string, unknown>): string {
+  return JSON.stringify(value);
+}
+
+function validOutput(): string {
+  return [
+    line({
+      type: "message",
+      message_type: "tool_call_message",
+      tool_call: {
+        tool_call_id: "call-echo",
+        name: "Bash",
+        arguments: JSON.stringify({
+          command: "echo 'Hello from Windows'",
+        }),
+      },
+    }),
+    line({
+      type: "message",
+      message_type: "tool_return_message",
+      tool_call_id: "call-echo",
+      status: "success",
+      tool_return: "Hello from Windows",
+    }),
+    line({
+      type: "message",
+      message_type: "tool_call_message",
+      tool_call: {
+        tool_call_id: "call-multiline",
+        name: "Bash",
+        arguments: JSON.stringify({
+          command: "echo 'Line1'; echo 'Line2'",
+        }),
+      },
+    }),
+    line({
+      type: "message",
+      message_type: "tool_return_message",
+      tool_call_id: "call-multiline",
+      status: "success",
+      tool_return: [{ type: "text", text: "Line1\nLine2" }],
+    }),
+    line({
+      type: "message",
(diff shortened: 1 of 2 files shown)


jibramnobhill:
Addressed Titan's blocker in `8c39c187`: the validator now parses each Bash arguments object, pairs each call with its own return by `tool_call_id`, identifies the echo, multiline, and Git executions separately, and validates each command's expected output only against that call's return. Added the requested cross-call substitution fixture. It fails under the previous aggregate-output behavior and passes with the fix. Focused tests pass 6/6 and all 12 repository checks pass.

jibramnobhill:
Addressed the remaining command-spoof blocker in `29b61135`. Required Bash commands are now normalized and matched by exact command equality rather than substring. Added the requested `echo 'git version 9'; # git --version` negative fixture; it fails when equality is mutated back to substring matching and passes with the fix. Focused suite: 7/7 passed. Current-head CI is running.

jibramnobhill:
Current-head CI is green at `29b61135`: 19 substantive checks passed, one review check intentionally skipped while draft, and zero checks failed. Please rereview this head.

PR #4112, opened 2026-08-29
fix(permissions): apply Read deny rules to Grep/Glob search paths

carenthomas:
## Summary

Grep and Glob search the directory their `path` argument resolves to. Previously, `Read(secrets/**)` deny rules did not block `Grep` or `Glob` from searching the `secrets` directory, because the permission matcher requires the tool name to match (`Grep` ≠ `Read`).

This mirrors a security fix in Claude Code 2.1.251 (permissions.md docs change): "Grep and Glob search the directory the `path` argument resolves to. Claude Code applies `Read` deny rules to that directory."

The fix builds `Read(path)` and `Read(path/**)` queries for Grep/Glob invocations and checks them against deny rules (both session/user deny rules and CLI `--disallowedTools`), so:
- `Read(secrets/**)` now blocks `Grep` searching `secrets`
- `Read(secrets/**)` now blocks `Glob` searching `secrets`
- `Read(secrets/**)` does not block `Grep` searching `src` (unaffected directory)
- Tool aliases (`grep_files`, `glob_gemini`) are canonicalized and also covered

## Test plan

- [x] `bun test src/permissions/permissions-checker.test.ts` — 56 pass, 0 fail
- [x] `bun test src/permissions/` — 529 pass, 0 fail (all permission tests)
- [x] `bun run typecheck` — clean
- [x] `bunx @biomejs/biome check` — clean

## Provenance

Claude-watch: `claude@2.1.251:docs:98d9b9f1e830b3e0b4eafbe5f459c46fcff19ca987559d239da825bc56f95fcc:runtime:20f0ec137fdd9e9217fd1e49221ffa37237a253ab468d092da253530f62902ca`

### DIFF
--- a/src/permissions/checker.ts
+++ b/src/permissions/checker.ts
@@ -293,6 +293,37 @@ function checkPermissionForEngine(
     permissionToolName === toolName
       ? query
       : buildPermissionQuery(originalQueryTool, toolArgs, engine);
+
+  // Grep and Glob search the directory their `path` argument resolves to.
+  // Read deny rules must also apply to that directory, so build Read queries
+  // with the same path to check against deny patterns.
+  // We check both Read(path) for exact matches and Read(path/**) for directory
+  // scope matches, so Read(secrets/**) denies Grep searching the secrets dir.
+  const searchPath =
+    canonicalTool === "Grep" || canonicalTool === "Glob"
+      ? extractFilePath(toolArgs)
+      : null;
+  const searchReadQueries = searchPath
+    ? [`Read(${searchPath})`, `Read(${searchPath}/**)`]
+    : null;
+
+  // Check if a Read deny rule matches a Grep/Glob search path
+  const checkReadDeny = (pattern: string, traceStage: string): boolean => {
+    if (!searchReadQueries) return false;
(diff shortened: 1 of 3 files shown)


src/permissions/checker.ts
```diff
@@ -293,6 +293,37 @@ function checkPermissionForEngine(
...
       : buildPermissionQuery(originalQueryTool, toolArgs, engine);
+
+  // Grep and Glob search the directory their `path` argument resolves to.
+  // Read deny rules must also apply to that directory, so build Read queries
+  // with the same path to check against deny patterns.
+  // We check both Read(path) for exact matches and Read(path/**) for directory
+  // scope matches, so Read(secrets/**) denies Grep searching the secrets dir.
+  const searchPath =
+    canonicalTool === "Grep" || canonicalTool === "Glob"
+      ? extractFilePath(toolArgs)
```
amelia-letta, line 304: When Grep/Glob is called without a `path` arg, `extractFilePath` returns null and the Read deny check is skipped entirely. An agent could bypass `Read(secrets/**)` by omitting `path`, since Grep defaults to searching the whole working directory.

PR #4136, opened 2026-08-31
feat(channels): add first-party Feishu / Lark channel

### LINKED ISSUE #4130: Feature Request: Feishu (Lark) Channel Support
### AI Disclosure

### AI Tool(s) Used

letta-code

### Human Verification

I have personally reviewed this request, verified that it addresses a real use case, and take responsibility for its contents.

### Problem statement

  Letta currently supports Slack, Telegram, and Discord as messaging channels. None of these are
  reliably accessible from mainland China without a VPN. Feishu (飞书/Lark) is the dominant workplace
  communication platform in China, used by millions of developers and companies. Without Feishu
  support, the entire Chinese-speaking user base is effectively locked out of Letta's channel feature.

### Proposed solution

  Add a native Feishu channel to Letta, following the same pattern as existing channels (Slack,
  Telegram, Discord). The Lark Open API provides everything needed — WebSocket-based long connections
  for real-time events, im.message.receive_v1 for inbound messages, and a bot messaging API for
  replies. All well-documented.

  Migration cost is low: OpenClaw already supports Feishu as a channel, so the integration pattern is
  proven and the API surface is mapped. The work from OpenClaw's implementation can serve as a direct
  reference.

  I've built a working bridge myself (feishu-bridge.py + letta server) that handles real-time messaging
  in production, which I'm happy to share as a reference. A native channel would eliminate the need for
  custom bridge code, manual service management (launchd/systemd), and trial-and-error event
  subscription debugging.

### Alternatives considered

_No response_

### Additional context

_No response_

himax12:
## Summary

- Add a messenger-only first-party `feishu` channel (display name **Feishu / Lark**) over the Open Platform persistent connection (WebSocket). No webhook, cards, Drive, or custom Desktop screens.
- Pairing and allowlists use Feishu `open_id` (`ou_*`). Groups default to **mention-only** (matches the default `@bot` event scope). `@all` is not a bot mention. Bot/app senders are dropped.
- One running listener per App ID (cluster mode). `domain` selects Feishu (`open.feishu.cn`) vs Lark (`open.larksuite.com`). Text send goes through the shared `MessageChannel` tool.
- `appId` / `appSecret` / `domain` land on the existing flattened `ChannelPluginAccountPatch` (same shared create/update path as Signal), not a new discriminated patch union.

Closes #4130.

## Why

Mainland Feishu and international Lark are the same Open Platform with different consoles. A first-party plugin is what makes `letta channels configure/install/pair` work without a user-plugin manifest. Persistent connection is the local analog of Slack Socket Mode.

## Limits / non-goals

- No published `./channels/feishu` package export (Slack/Telegram subpaths shipped because Cloud needed them). Pure `ingress.ts` / `outbound.ts` stay SDK-free for a later webhook host.
- No HTTP webhook, Card Kit, docs/drive/Bitable, store apps, or `allowBots`.
- Self-built apps only. Restart the listener after editing `accounts.json`.

## Verification

- `bun run check` — 12/12
- `bun test src/channels/feishu src/channels/registry-community-copy.test.ts` — 34 passed
- Lifecycle errors go through `formatChannelLifecycleErrorMessage` (adapter test covers sanitizing before post).
- Mention-only drop, `message_id` dedup, and second App ID start are covered by tests that fail without this change.

## Live validation

No live Feishu/Lark app probe was run. Harness-only: injected SDK client, ingress/outbound unit tests, and the App ID uniqueness guard.

## AI Disclosure

## AI Tool(s) Used

Cursor

## Human Verification

Made with [Cursor](https://cursor.com)

Linear: https://linear.app/letta/issue/LET-11920/add-first-party-feishulark-channel-support

### DIFF
--- a/src/channels/feishu/ingress.ts
+++ b/src/channels/feishu/ingress.ts
@@ -1,0 +1,263 @@
+import type { FeishuGroupMode, InboundChannelMessage } from "@/channels/types";
+import { isRecord } from "@/utils/type-guards";
+
+export interface FeishuSenderId {
+  open_id?: string;
+  user_id?: string;
+  union_id?: string;
+}
+
+export interface FeishuMention {
+  key?: string;
+  name?: string;
+  mentioned_type?: string;
+  id?: FeishuSenderId;
+}
+
+export interface FeishuReceiveMessage {
+  message_id?: string;
+  root_id?: string;
+  parent_id?: string;
+  create_time?: string;
+  chat_id?: string;
+  chat_type?: string;
+  message_type?: string;
+  content?: string;
+  thread_id?: string;
+  mentions?: FeishuMention[];
+}
+
+export interface FeishuReceiveSender {
+  sender_id?: FeishuSenderId;
+  sender_type?: string;
+  tenant_key?: string;
+}
+
+export interface FeishuReceiveEventBody {
+  sender?: FeishuReceiveSender;
+  message?: FeishuReceiveMessage;
+}
+
+export type FeishuIngressDropReason =
+  | "malformed"
+  | "bot_sender"
+  | "self_sender"
+  | "mention_required"
+  | "broadcast_all"
+  | "empty";
+
+export type FeishuIngressDecision =
+  | { action: "drop"; reason: FeishuIngressDropReason }
+  | { action: "deliver"; inbound: InboundChannelMessage };
+
+const MENTION_KEY_RE = /@_user_\d+/g;
+const BROADCAST_ALL_TEXT_RE = /@_all\b|@all\b/i;
+const MEDIA_PLACEHOLDERS: Record<string, string> = {
+  image: "[image]",
+  file: "[file]",
+  audio: "[audio]",
+  media: "[video]",
+  sticker: "[sticker]",
+  interactive: "[card]",
+  share_chat: "[shared chat]",
+  share_user: "[shared user]",
+  merge_forward: "[forwarded messages]",
+  system: "[system]",
+};
+
+export function unwrapFeishuReceiveEvent(
+  payload: unknown,
+): FeishuReceiveEventBody | null {
+  if (!isRecord(payload)) {
+    return null;
+  }
+  if (isRecord(payload.event)) {
+    return payload.event as FeishuReceiveEventBody;
+  }
+  if (isRecord(payload.message) || isRecord(payload.sender)) {
+    return payload as FeishuReceiveEventBody;
+  }
+  return null;
+}
+
+export function stripFeishuMentionKeys(text: string): string {
+  return text
+    .replace(MENTION_KEY_RE, " ")
+    .replace(/@_all/g, " ")
+    .replace(/\s+/g, " ")
+    .trim();
+}
+
+function isBroadcastAllMention(mention: FeishuMention): boolean {
+  const key = mention.key?.trim().toLowerCase();
+  return key === "@_all" || key === "@all";
+}
+
+export function isFeishuBroadcastAllOnly(
+  text: string,
+  mentions: FeishuMention[],
+): boolean {
+  const hasAllMention = mentions.some(isBroadcastAllMention);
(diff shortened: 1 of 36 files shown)


src/channels/feishu/ingress.ts
```diff
@@ -0,0 +1,263 @@
...
+    return false;
+  }
+  return !mentions.some((mention) => !isBroadcastAllMention(mention));
+}
+
+export function isFeishuBotMention(
+  mentions: FeishuMention[],
+  botOpenId?: string | null,
+): boolean {
+  if (mentions.some((mention) => mention.mentioned_type === "bot")) {
```
lipracer: match on a field the payload does not have. refer https://open.feishu.cn/document/server-docs/im-v1/message/events/receive

lipracer: Nice structure overall — follows the existing channel conventions and skipping
media downloads avoids a lot of trouble. One blocker though.

**P0: group @mentions are never detected, so group chat does nothing by default.**

`isFeishuBotMention()` has two ways to detect a mention and both are broken:

```ts
// A — field does not exist in the payload
if (mentions.some((m) => m.mentioned_type === "bot")) return true;
// B — never runs: botOpenId is always null
const trimmedBotOpenId = botOpenId?.trim();
if (!trimmedBotOpenId) return false;
return mentions.some((m) => m.id?.open_id === trimmedBotOpenId);

himax12: Agreed — thanks for catching this. `mentioned_type` isn't on the live `im.message.receive_v1` mention object, and we were also passing `botOpenId: null`, so both checks were dead and mention-only groups dropped everything.

Fixed in d6fcaa46: start calls `GET /open-apis/bot/v3/info`, then `isFeishuBotMention()` matches `mentions[].id.open_id` against that bot `open_id`. `mentioned_type` is ignored. Tests cover the official-shaped payload (no `mentioned_type`), a `mentioned_type: bot` red herring.

lipracer:
Re-reviewed after the update — the P0 is resolved, and the fix is more thorough
than what I suggested. Removing `mentioned_type` from the `FeishuMention`
interface (not just from the check), extracting `parseFeishuBotInfo` so
`account-display.ts` reuses it, and clearing `botOpenId` on all three teardown
paths all look right.

The regression test is the part I'd single out: keeping a fixture that still
carries `mentioned_type: "bot"` while pointing `open_id` at a different user,
and asserting the message is dropped, locks the bug out permanently rather than
just fixing it.

## One new point — and a correction to my earlier suggestion

I argued for failing `start()` when `/open-apis/bot/v3/info` fails. You chose to
warn and continue, and you're right that hard-failing would break DM-only setups
over a call that path doesn't need.

The current shape has no recovery path, though. A single transient failure at
startup leaves `botOpenId` null for the whole session, and because `isRunning()`
reports `true` the channel looks healthy while every group @mention is silently
dropped. The only clue is one startup warning that has long scrolled away.

A lazy retry covers both cases better than either fail-fast or warn-once: on the
first mention-only group event, if `botOpenId` is still null, resolve it again
with simple backoff — only on the group path, so DMs stay unaffected. Transient
failures then self-heal, and persistent ones keep producing log lines instead of
a single one that was easy to miss.

## Still open from the earlier pass

Both non-blocking as far as I'm concerned:

- `senderName` is resolved by searching `mentions` for the sender, but that array
  holds *mentioned* users, so in practice it falls back to the raw `ou_...` id.
- `start()` sets `running = true` before the socket settles and installs no
  close/error handler, so `isRunning()` stays `true` after a mid-session
  disconnect. Worth confirming whether `WSClient` reconnects internally.

## Standing

I'm not a code owner on this repo and can't approve or gate the merge — please
treat this as a drive-by review from someone who wants this channel to land, not
as a required sign-off.

<sub>Reviewed with Comate CLI (model: Opus 5). Findings were verified
against the PR diff and the Feishu <code>im.message.receive_v1</code>
documentation; no live tenant test was run on my side either.</sub>

PR #4161, opened 2026-09-01
fix(cron): run local schedules on the local listener [LET-10446]

just-cameron:
## Summary

Desktop runs separate cloud and local-backend listeners, but `crons.json` previously allowed only one listener to own all natural schedule fires. A local agent's natural schedule could therefore run through the cloud listener while its Telegram route and `MessageChannel` tool were registered on the local listener. Manual **Send now** was already routed to the local listener and continued to work.

This pull request lets cloud and local listeners hold separate scheduler leases and process only schedules for their own backend. Standalone listeners retain the existing all-agent lease. Legacy all-agent leases and scoped leases exclude one another so mixed-version processes cannot fire the same schedule twice.

This draft is blocked on the dependent Desktop wiring and a full natural-fire product-path test using both listener lanes.

### Before

```text
cloud listener: owns every natural schedule fire
local listener: owns the local agent's Telegram route and MessageChannel
result: local natural fire starts without MessageChannel
```

### After

```text
cloud listener: owns cloud-agent schedule fires
local listener: owns local-agent schedule fires and the local Telegram route
result: local natural fire starts where MessageChannel is registered
```

The red-first tests start from the persisted scheduler lease state and local versus cloud agent IDs. On unmodified main, scoped leases are not independent and schedules cannot be partitioned by backend. With this change, cloud and local leases coexist, an all-agent lease remains mutually exclusive with scoped leases, and each scheduler accepts only its backend's tasks.

## Verification

- `bun test src/cron/cron-file.test.ts src/cron/scheduler.test.ts src/websocket/listener/channel-runtime-tools.test.ts src/channels/gateway-core.test.ts src/channels/routed-runtime-registration.test.ts` (109 passed)
- `bun run typecheck`
- Pre-commit source-size, ownership, Biome, and TypeScript checks
- Red-first run against the exact unmodified base failed in the scoped lease and backend partition assertions
- Not yet verified through a packaged Desktop natural schedule fire. This blocks ready-for-review.

## Breadcrumbs

- Linear: LET-10446
- Letta conversation: [`conv-264249fb-e28a-4fa4-b5f5-c7b1ed805232`](https://chat.letta.com/chat/agent-ec05a4e0-f50c-47fd-8e7b-960f081ab42e?conversation=conv-264249fb-e28a-4fa4-b5f5-c7b1ed805232)
- Origin: no single introducing pull request was found. The earliest relevant scheduler ownership change assigned schedule execution to one dedicated listener; later Desktop local-backend channel restoration created a split ownership boundary.
- Related: follow-up to https://github.com/letta-ai/letta-code/pull/3687, which made `MessageChannel` gateway-owned and confirmed process-owned turns use routed runtime tools
- Replaces the approach in closed https://github.com/letta-ai/letta-code/pull/3661, which targeted scheduled-turn tool delegation rather than listener ownership

## AI Disclosure

## AI Tool(s) Used

Letta Code performed source tracing, implementation, test authoring, local validation, and drafting. Cursor Agent and Letta Code GLM subagents were used during investigation but returned no usable implementation or conclusions.

## Human Verification


src/cron/cron-file.ts
```diff
@@ -643,28 +658,55 @@ export function deleteAllTasks(agentId: string): number {
...
 
-    data.scheduler_owner = {
+    const owner = {
       pid: process.pid,
       token,
       started_at: new Date().toISOString(),
       ...captureProcessIdentity(process.pid),
     };
+    if (scope === "all") {
+      for (const existingOwner of Object.values(data.scheduler_owners)) {
```
amelia-letta: A legacy listener only checks `scheduler_owner`, then normalizes this file without `scheduler_owners`. If it starts after a scoped listener, it can claim all schedules and erase the scoped lease before the new listener’s next tick, allowing a duplicate fire, so mixed-version exclusion isn’t bidirectional.

src/cron/cron-file.ts
```diff
@@ -101,6 +103,9 @@ export interface CronTask {
 interface CronFileData {
   version: 1;
   scheduler_owner: SchedulerOwner | null;
+  scheduler_owners: Partial<
```
amelia-letta, line 113: `AppCoordinator`’s shadow TUI scheduler still checks only `scheduler_owner`, while scoped Desktop listeners leave that field null. With `tui_cron` enabled, the TUI and matching scoped listener can both fire the same task.

src/cron/scheduler.ts
```diff
@@ -663,7 +687,10 @@ export function startScheduler(
     pendingTimers: new Set(),
   };
 
-  // Initial tick
+  schedulerState = state;
+
+  // Initial tick after the process is recorded as running so zero-jitter
+  // fires are not dropped by the `if (!schedulerState) return` revalidation.
   tick(state, socket, opts, processQueuedTurn);
```
amelia-letta, line 750: If this initial tick detects that the lease was lost, it calls `stopScheduler()` and clears `schedulerState`, but execution continues and installs two intervals on the detached state. Return when `schedulerState !== state` before arming intervals, otherwise the failed scheduler wakes forever and can keep running GC.

src/cli/app/AppCoordinator.tsx
```diff
@@ -1470,14 +1470,12 @@ export function App({
...
+      // listeners own scheduler_owners.local|cloud and keep a scheduler_owner
+      // tombstone; treat any of those live rows as an active lease holder.
       const cronData = readCronFile();
-      if (cronData.scheduler_owner) {
-        const { pid } = cronData.scheduler_owner;
-        if (isProcessAlive(pid, cronData.scheduler_owner)) {
-          // Desktop app is running the scheduler — defer
-          return;
-        }
+      if (hasLiveSchedulerOwner(cronData)) {
```
amelia-letta: This defers the TUI when either scoped owner is live. If only the opposite lane is running, nobody processes this agent's schedules because that listener filters them out, so this check needs to match the active agent's backend while still honoring a true all-owner.

src/cron/cron-file.ts
```diff
@@ -673,36 +687,70 @@ export function claimSchedulerLease(): string {
...
-    if (
-      data.scheduler_owner &&
-      data.scheduler_owner.pid === process.pid &&
-      data.scheduler_owner.token === token
-    ) {
-      data.scheduler_owner = null;
+    const owner =
+      scope === "all" ? data.scheduler_owner : data.scheduler_owners[scope];
+    if (!ownerMatches(owner, token)) return false;
+    if (scope !== "all" && !isLiveOwner(data.scheduler_owner)) {
```
amelia-letta: If the process named by `scheduler_owner` crashes while the other scoped owner remains alive, there is still a window until this surviving scheduler’s next 60s tick where a legacy listener sees a dead owner, claims `all`, and strips `scheduler_owners`. The survivor can keep firing during that window, so the mixed-version duplicate-fire race this tombstone is meant to prevent remains.

src/cron/scheduler.ts
```diff
@@ -684,7 +724,14 @@ export function startScheduler(
     }
   }, GC_INTERVAL_MS);
 
-  schedulerState = state;
+  if (scope !== "all") {
+    state.tombstoneInterval = setInterval(() => {
+      if (!refreshSchedulerLease(state.token, state.scope)) {
```
amelia-letta: `refreshSchedulerLease()` can throw after waiting 5s for `crons.lock`. If another process crashes while holding the lock, this 1s callback hits the 30s stale-lock window and the uncaught timer error can crash the surviving listener, so catch and log heartbeat failures without treating a transient lock error as lease loss.

src/cron/scheduler.ts
```diff
@@ -684,7 +724,21 @@ export function startScheduler(
     }
   }, GC_INTERVAL_MS);
 
-  schedulerState = state;
+  if (scope !== "all") {
+    state.tombstoneInterval = setInterval(() => {
+      try {
+        if (!refreshSchedulerLease(state.token, state.scope)) {
```
amelia-letta: `refreshSchedulerLease` treats a missing scoped row as lease loss, but any older client that updates a task rewrites this file without `scheduler_owners` while preserving the live tombstone. Both scoped schedulers then stop, and the tombstone blocks them from reclaiming until its process exits.

src/cron/cron-file.ts
```diff
@@ -673,36 +687,93 @@ export function claimSchedulerLease(): string {
...
+    const owner = data.scheduler_owners[scope];
+    if (ownerMatches(owner, token)) {
+      if (!isLiveOwner(data.scheduler_owner)) {
+        data.scheduler_owner = owner;
+        writeCronFile(data);
+      }
+      return true;
+    }
+    if (isLiveOwner(owner)) return false;
+    const restored = createSchedulerOwner(token);
```
amelia-letta, line 751: After an old writer strips `scheduler_owners`, a legacy listener can claim `scheduler_owner` during the heartbeat window. This branch then restores the surviving scoped row while preserving that live all-owner and returns true, so both schedulers can fire the scoped lane indefinitely.

src/cron/cron-file.ts
```diff
@@ -673,36 +687,99 @@ export function claimSchedulerLease(): string {
...
     if (
-      data.scheduler_owner &&
-      data.scheduler_owner.pid === process.pid &&
-      data.scheduler_owner.token === token
+      isLiveOwner(data.scheduler_owner) &&
+      !ownerMatches(data.scheduler_owner, token)
     ) {
-      data.scheduler_owner = null;
-      writeCronFile(data);
+      return false;
```
amelia-letta, line 749: After a legacy writer strips `scheduler_owners`, this foreign `scheduler_owner` can still be the other scoped lane’s tombstone, not an all-owner. Returning false makes this lane call `stopScheduler()` with no retry path, so one backend stops firing schedules until the listener restarts.

src/cron/scheduler.ts
```diff
@@ -117,6 +144,32 @@ function logScheduler(opts: StartListenerOptions, message: string): void {
...
+    state.token = claimSchedulerLease(state.scope);
+    state.tombstoneToken = readCronFile().scheduler_owner?.token;
+    logScheduler(opts, "Reclaimed scoped scheduler lease.");
+    return true;
+  } catch (err) {
+    logScheduler(
+      opts,
+      `Scheduler lease lost; retrying: ${err instanceof Error ? err.message : String(err)}`,
+    );
+    return false;
```
amelia-letta, line 170: When a scoped lease is lost to a live all-owner, this retry path leaves `schedulerState` and its pending jitter timers active. An already-armed `doFire` only checks that global state is non-null, so it can still fire alongside the all-owner; clear those timers while unleased or revalidate the lease before firing.

src/cron/scheduler.ts
```diff
@@ -549,6 +609,15 @@ function tick(
         // have been lost, or the task may have been deleted/cancelled during
         // the jitter window.
         if (!schedulerState) return;
+        if (
+          !refreshSchedulerLease(
```
amelia-letta: `refreshSchedulerLease()` can throw after waiting 5s for `crons.lock`, and this new timer callback does not catch it. A stale or contended lock therefore turns a skipped jitter fire into an uncaught exception that terminates the listener, so catch and log it here as the heartbeat does.

PR #4179, opened 2026-09-01
feat(tools): Workflow orchestration tool with background execution and live monitoring

sarahwooders:
## Summary

Adds a `Workflow` tool for JavaScript scripts that coordinate parent-linked, agent-free ephemeral conversations. Every worker has `agent_id: null`, the invoking `parent_agent_id`, `is_subagent: true`, and a worker name. It runs in the background, reports progress through task notifications and `/workflows`, and journals results for resume. `pipeline()` advances each item independently; `parallel()` waits for all results.

Workflows run locally or on existing connected computers, with a shared concurrency limit and per-call model, tool, and computer overrides. Local structured results use `StructuredOutput`; remote results use validated JSON text so existing listeners do not need query-scoped external tools.

The branch now incorporates current main and fixes empty tool allowlists, default-dependent cache identity, cancellation while sleeping, truncated task-output results, object-key order in schema validation, and launch-time sender identity on completion notifications.

The latest review fixes move script execution into a terminable worker thread while SDK calls remain in the invoking runtime. An unawaited agent call cannot terminate the host, and TaskStop can stop a script stuck in a CPU loop after an await. Local workers and relative scripts use the conversation directory; model inheritance works from agent-free conversations. Metadata comments parse correctly, and the Letta toolset includes Workflow.

## Review notes

- Resume compares the parent agent, effective model, tool allowlist, computer, local working directory, and the call's position within pipeline/parallel helpers. Identical calls stay associated with their items when upstream cache hits change completion order. Older journals remain readable, but their calls rerun because the old keys omitted execution settings, parent resources, or helper positions. Cache replay does not detect changed files.
- Existing computers are borrowed, not terminated. Fresh sandbox provisioning, automatic repository attachment, and file transfer are not included.
- `budgetUsd` is advisory: completed reported costs gate new calls, but cannot cap in-flight or unreported spending. Unknown cost remains `null`.

## Verification

- Latest review fixes: reproduced host termination, an uninterruptible CPU loop, wrong-directory execution, and swapped replay results before repair. The Bun source and bundled Node runner now survive detached worker calls, stop CPU loops after an await, and run another workflow in the same host. Replay tests deliberately reverse upstream completion order and cover pipelines, parallel helpers, and nested helpers.
- A built Node tool-handler run against the real API returned two schema-valid worker results, verified their null agent ownership and parent/subagent fields, and replayed both calls from the journal. It used an existing accessible parent and deleted the disposable worker conversations afterward. The stock API fixture could not retrieve newly created test-parent agents with this sandbox credential; a direct API create/retrieve check reproduced that 404 independently of Workflow.
- Focused workflow, task lifecycle, display, and background-snapshot tests: 86 passed.
- `bun test src/integration-tests/workflow.integration.test.ts`: passed against the real API and this branch's built Node CLI using published Agent SDK 0.8.11, without a local SDK override. Two concurrent workers returned schema-valid results and their stored conversations had null agent ownership, the correct parent, subagent status, and names. Invocation from an existing ephemeral child resolved the same parent; replay returned two cache hits, zero new queries, and zero new token usage. This test runs in the existing credentialed API CI matrix, without provider mocks.
- Earlier merge validation also passed the listener interruption, external-tool, and Monitor suites (30 tests).
- In the real Node TUI, a three-worker workflow completed in 27 seconds, emitted its completion notification, and resumed the assistant automatically. The live footer updated while `/workflows` provided on-demand snapshots. TaskStop terminated a sleeping workflow, showed `Workflow stopped`, and did not emit a completion notification. Both light and dark themes were exercised.
- A separate Node handler run verified that an invalid model produces a journaled failed outcome and `null`, and that stopping a sleeping script promptly terminates its registry entry without a completion notification.
- Worker identities are retained in the journal, including structured-output retries. Older SDKs without query identity support are rejected before creating workers. The dependency now requires Agent SDK 0.8.11, containing the metadata forwarding fix.
- A fresh Node run using the published SDK also passed concurrent local and Cloud-routed workers on a disposable listener. Both persisted all four identity fields, returned the expected structured results, and replayed with zero new workers. Test conversations and their parent were deleted, and the listener was stopped. No Cloud API patch or deployment was required.

## Visual proof

Captured from this branch's real Node TUI against the API, using synthetic weather notes and disposable test resources. Click a poster for the continuous desktop recording.

[Video: background workflow execution](https://chat.letta.com/artifacts/d8c6bf45-e787-4354-b2f8-d37c648bc6c9.mp4)

[Video: stopping a background workflow](https://chat.letta.com/artifacts/630bb124-aae5-49dd-a7eb-ecfccffb1d15.mp4)

| Dark | Light |
| --- | --- |
| Running workflow, dark | Running workflow, light |
| Completed workflow, dark | Completed workflow, light |

## Breadcrumbs

- Review and repair: [conv-415241f7-76af-4edd-97b0-4e73801ee875](https://chat.letta.com/chat/agent-b86b0d3a-c425-4c88-86b9-44171f178713?conversation=conv-415241f7-76af-4edd-97b0-4e73801ee875); [review findings](https://github.com/letta-ai/letta-code/pull/4179#pullrequestreview-5230074378).
- Origin: continues the workflow implementation in this PR; the fixes above address defects in that implementation, not a separately merged regression. The earlier multi-computer implementation starts at [800081eb](https://github.com/letta-ai/letta-code/commit/800081eb973060deeda34f15a3245f6e5b59b606).
- SDK dependency: [letta-agent-sdk #317](https://github.com/letta-ai/letta-agent-sdk/pull/317), released in 0.8.11 and now required by this PR.
- Separately discovered parentless attribution bug: LET-12673; not required for the intended parent-linked workers.
- Earlier verification: [conv-6990ac46-57da-4d7b-a037-1d24fa3df97a](https://chat.letta.com/chat/agent-57231da8-42f3-4523-b190-66c3eda21057?conversation=conv-6990ac46-57da-4d7b-a037-1d24fa3df97a).
- Current verification: [conv-ca3e2972-7056-4eca-95f8-bb424bf7b22f](https://chat.letta.com/chat/agent-57231da8-42f3-4523-b190-66c3eda21057?conversation=conv-ca3e2972-7056-4eca-95f8-bb424bf7b22f).

## AI Disclosure

Written with AI assistance. Human review remains pending; no human verification is asserted by the agent.

## AI Tool(s) Used

Letta Code: implementation, review, regression tests, live verification, and PR documentation.

## Human Verification

- [ ] I have reviewed and understand every change in this pull request and take responsibility for its correctness.


amelia-letta (approved):
LGTM

amelia-letta (approved):
LGTM

amelia-letta (approved):
LGTM

src/tools/impl/workflow.ts
```diff
@@ -0,0 +1,497 @@
...
+  let parentAgentId: string | null | undefined = args.parentScope?.agentId;
+  const conversationId =
+    args.parentScope?.conversationId ?? getConversationId();
+  if (!parentAgentId?.startsWith("agent-")) {
+    if (conversationId && conversationId !== "default") {
+      const conversation = await apiRequest<{
+        agent_id: string | null;
+        parent_agent_id?: string | null;
+      }>("GET", `/v1/conversations/${encodeURIComponent(conversationId)}`);
+      parentAgentId =
```
amelia-letta, line 118: When Workflow is called from an agent-free conversation with no explicit `model`, this resolves the lineage parent but the later model lookup omits it; `getCurrentAgentId()` then throws for the null-agent runtime, leaving every `agent()` call to fail with “requires a model.” Pass `parentAgentId` into `getPrimaryAgentModelHandle` so the documented model inheritance works for this path.

amelia-letta:
@cpacker Should default model resolution use the captured `parentScope.conversationId` instead of ambient `getConversationId()`, so a listener-scoped Workflow cannot inherit another conversation’s model?

src/tools/toolset-defaults.ts
cpacker, line 1: This probably needs to be added to the `Letta` custom toolset we recently added

amelia-letta (approved):
LGTM

PR #4205, opened 2026-09-03
fix(agent): keep memoryScope on provider-not-supported subagent retry

wutongyuonce:
## Summary

When a subagent child exits because its model/provider is not supported, `executeSubagent` retries once with the parent agent's primary model. That retry correctly passes `existingAgentId` / `existingConversationId` as `undefined` so `buildSubagentArgs` creates a new agent and can pass `--model` (deploying an existing agent omits `--model` and would keep the unsupported provider).

The same retry was also passing `memoryScope` and `systemPromptOverride` as `undefined`. A reflection (or any memory-subagent) child then gets `MEMORY_DIR = inheritedPrimaryRoot` instead of `memoryScope.primaryRoot`, so the replacement process writes the parent memory repo rather than the worktree it was given.

Lost-stdout retries in the same function already forward those arguments. This change makes the provider-not-supported retry do the same for `memoryScope` and `systemPromptOverride`, and updates the source-snapshot test that previously asserted the drop.

## Verification

- `bun test src/agent/subagent-retry-parent-scope.test.ts src/agent/subagent-env-composition.test.ts src/agent/subagent-model-resolution.test.ts` — 93 pass
- `bun run check` — 12/12 pass

## AI Disclosure

## AI Tool(s) Used

Pi coding agent (grok 4.6) for diagnosis, the code change, tests, and this PR text. A human directed the change after confirming the retry call sites in `src/agent/subagents/manager.ts`.

## Human Verification

### DIFF
--- a/src/agent/subagent-retry-parent-scope.test.ts
+++ b/src/agent/subagent-retry-parent-scope.test.ts
@@ -8,9 +8,9 @@ const managerSource = readFileSync(
 );
 
 describe("executeSubagent provider fallback wiring", () => {
-  test("forwards parentAgentIdOverride through the provider retry call", () => {
+  test("retries with a new agent, the primary model, and the original memory scope", () => {
     const retryCallMatch = managerSource.match(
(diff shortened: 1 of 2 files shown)


wutongyuonce:
Rebased onto latest `main` to clear the merge conflict with #4334.

Resolution: the provider-not-supported retry still forwards `memoryScope` / `systemPromptOverride` (this PR), and keeps the new `actingUserIdOverride` argument from #4334. The source-snapshot test matches that combined call.

CODEOWNERS does not cover `src/agent/subagents/`, so tagging people who have actually owned this retry/memory path:

- @devanshrj reviewed/approved the adjacent retry wiring in #3303
- @kl2806 and @amysguan have been the recent human authors on `manager.ts` reflection/memory integration

Would one of you mind reviewing when you have a chance?

PR #4211, opened 2026-09-03
feat(telemetry): measure listener first-send latency [LET-11109]

kianjones9:
## Summary
- Adds telemetry-only aggregate completion events for listener `runtime_start` and first send preparation through the Core message stream request.
- Preserves explicit runtime-target `agent_id` on listener latency events even when `TelemetryManager` has a different current agent, including queued telemetry events enriched later.
- Uses one accepted pre-ACK timing boundary for both direct and queued inputs; no runtime behavior, queueing, send, payload, or Core request semantics changed.

## Exact telemetry fields
- `listener_runtime_start_complete`: `request_id_hash`, `connection_id_hash`, `agent_id`, `conversation_id`, `success`, `created_agent`, `created_conversation`, `wait_for_replay`, `duration_ms`, `validate_ms`, `resolve_agent_ms`, `resolve_conversation_ms`, `source_tags_ms`, `runtime_state_ms`, `subscribe_tools_ms`, `ack_ms`, optional `replay_before_ack_ms`, optional `error_type`, plus standard `version`, `platform`, `session_id`, `surface`, `backend`.
- `listener_input_send_complete`: `connection_id_hash`, `agent_id`, `conversation_id`, `client_message_id_hash`, `client_message_count`, `message_count`, `includes_approval`, `success`, `accepted_to_core_stream_ms`, `prepare_listener_turn_ms`, `skill_content_injection_ms`, `core_stream_request_ms`, optional `error_type`, plus standard `version`, `platform`, `session_id`, `surface`, `backend`.

## Privacy / cardinality
- `request_id`, `connection_id`, and `client_message_id` are emitted only as deterministic SHA-256 prefix hashes (`*_hash`). This supports exact filtering when an operator already has the source ID, but avoids exposing raw protocol/client identifiers and keeps these fields out of grouping dimensions.
- `client_message_id` is client supplied in the inbound message path, so hashing is the safe choice over raw emission. `request_id` is also caller supplied by protocol clients. `connection_id` is an existing listener identifier, but hashed too for consistent correlation hygiene.
- No message content, tool payloads, prompts, file paths, or user text are emitted.

## Timing semantics
- `accepted_to_core_stream_ms` starts at the listener acceptance boundary immediately before ACK/disposition emission for both direct and queued paths.
- For queued inputs, the metric intentionally includes queue wait after accepted ACK until the turn opens the Core message stream.

## Validation
- `bun install --frozen-lockfile`
- `bun test src/telemetry/flush-batching.test.ts src/websocket/listener/commands/runtime-start-skill-sources.test.ts`
- `bun run check`

## Overlap risks
- Overlaps with LET-11109 listener/startup performance branches by instrumenting the same first-send path, but this PR is telemetry-only.
- The only non-telemetry-looking extraction is the existing `turn-finalizer` helper from the first commit, kept solely to avoid growing oversized `turn.ts`; review fixes did not broaden that refactor.

### DIFF
--- a/src/telemetry/flush-batching.test.ts
+++ b/src/telemetry/flush-batching.test.ts
@@ -38,6 +38,8 @@ describe("telemetry flush batching", () => {
     telemetryState.sessionEndTracked = false;
     telemetryState.inflightFlush = null;
     deleteEnvVarCaseInsensitive("LETTA_API_KEY");
+    deleteEnvVarCaseInsensitive("LETTA_CODE_TELEM");
+    deleteEnvVarCaseInsensitive("DO_NOT_TRACK");
     deleteEnvVarCaseInsensitive("LETTA_TELEMETRY_DISABLED");
     deleteEnvVarCaseInsensitive("LETTA_BASE_URL");
     settingsManager.getSettings = mock(() => ({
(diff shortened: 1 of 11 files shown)


kianjones9:
gonna close for now bc this seems like slop tbh

PR #4233, opened 2026-09-05
docs(skills): update creating-mods events reference

amelia-letta:
## Summary

- Document `turn_end` event with `ModTurnEndEvent` shape and `{ continue: string }` result for follow-up turns
- Add `"reload"` to `conversation_open` and `conversation_close` reason unions
- Document `tool_start` result short-circuit: returning `{ result: { status, output } }` skips tool execution and uses `output` as the tool return value

Builtin-skill-watch: creating-mods@f61414acebd4-a5f21a2d26d7aceb


amelia-letta:
Superseded by #4118 which covers the same drift issues plus `alwaysAsk` permission.

PR #4290, opened 2026-09-08
fix(providers): send OpenCode Go session header

overlord-letta:
## tl;dr

OpenCode Go rejects inference requests without `x-opencode-session`. The local backend now derives that header from the stable Letta conversation ID at a shared pi-ai request boundary used by ordinary turns and compaction.

The same conversation keeps the same value across later turns, retries, tool-loop steps, and summary requests. Different conversations get different values. Other providers receive no OpenCode header, and existing configured headers are preserved.

## Testing

Live verification ran against OpenCode Go from an isolated Linux sandbox with a real credential. A direct request without `x-opencode-session` returned HTTP 400 `MissingSessionID`; the identical request with the header returned HTTP 200 and the expected response. The current PR source completed an ordinary local-backend turn through `opencode-go/glm-5.2` and returned the requested `LOCAL_GO_OK` response.

The compaction regression was also exercised against both code states with the same credential and input. Commit `a3a76451`, which covered ordinary turns but not compaction, failed with `MissingSessionID` and `Request is missing x-opencode-session`. Current head `a703c796` completed the real summary request successfully.

## Breadcrumbs

- Requested behavior: https://github.com/vercel/ai/issues/20271
- Linear: LET-12185
- Original implementation conversation: `conv-f5079c4e-d025-41fe-a9b3-7ce850d2fc2f`
- Live verification conversation: [`conv-73ab4ba5-9db3-4c87-89c9-780973bb0025`](https://chat.letta.com/chat/agent-ec05a4e0-f50c-47fd-8e7b-960f081ab42e?conversation=conv-73ab4ba5-9db3-4c87-89c9-780973bb0025)
- Origin: no Letta regression was found. The local backend began passing its conversation ID to pi-ai before OpenCode Go added this header requirement.
- Related: Cloud's OpenAI-compatible path is handled separately under LET-12185.

> Fix the boundary, not only the symptom.


src/backend/local/local-backend.ts
kianjones9, line 1: seems unrelated?

kianjones9 (approved):

PR #4299, opened 2026-09-08
fix(teleport): unblock source input after disconnect before handoff [LET-12212]

kianjones9:
## Summary

A teleport request must not leave the source rejecting later messages when its connection disappears before it can hand off. The source currently retains that request indefinitely when it finds a closed or missing connection, even though it has not sent `teleport_ready` or yielded its turn.

This rework removes only that unsent request. An active source turn keeps running, already-accepted input still drains, and saved tool results are untouched. Successful handoffs keep their existing input gate and recorded-turn handling.

The earlier version retried readiness when another copy of `teleport_request` arrived. The production caller does not resend that command on an idempotent API retry, so those replay helpers and their synthetic reconnect tests are removed. Kian Jones's original commits remain in the branch history.

## Before and after

```text
Before
teleport_request → source connection closes before readiness
source keeps pending request → later input rejected after reconnect

After
teleport_request → source connection closes before readiness
source removes unsent request without yielding its turn
accepted work finishes normally → later input can start after reconnect
```

## Scope

This only handles a missing or closed source connection detected before handoff. It does not repair write exceptions, delivery lost after an open-socket send, or disconnects after the source has yielded. It leaves recovery of those yielded turns unchanged.

It also does not complete or fail the Cloud teleport record. A record already waiting for source readiness can remain pending until expiry; this change removes the permanent local input block, not that separate Cloud status problem.

## Verification

Six deterministic boundary cases start at `handleTeleportRequest`, close or remove the source connection at idle, active-turn, and accepted-queue-drain boundaries, then submit another message through the production input dispatcher. Before the fix, all six rejected that later input. With the fix, it is admitted. They also check that cleanup preserves the active lease, accepted queue contents, and saved record bytes/results/request OTID. Existing successful-handoff coverage still rejects input after readiness.

An additional case runs that later message through production `handleIncomingMessage` using the existing deterministic local backend. It checks `turn_finished` with `end_turn`, reloads the backend from disk, and verifies the persisted assistant reply. The same test fails on pre-fix source because input is rejected. The transport and unrelated post-turn repository sync are mocked; turn setup, message processing, persistence, and finalization are real. This verifies a local source turn, not local teleport capability or a live Cloud teleport.

Cross-service disconnect behavior and ordering of saved-result recovery after a yielded turn remain unverified and outside this patch.

## AI Tool(s) Used

Letta Code.

## Breadcrumbs

- Requested in: LET-12212.
- Original contribution: Kian Jones's [readiness cleanup commit](https://github.com/letta-ai/letta-code/commit/d245360b7ef699c487d3e4275e5cc06194077843).
- Origin: [teleport implementation](https://github.com/letta-ai/letta-code/pull/3813). The missing-connection branch retained the request without completing a handoff; this is separate from the end-turn continuation fixed in [4477](https://github.com/letta-ai/letta-code/pull/4477).
- Rework decision: [review and scope](https://github.com/letta-ai/letta-code/pull/4299#issuecomment-5690082014).
- Generating conversation: [`conv-88bfcb93-58e6-46ca-9240-f89211fd3dbf`](https://chat.letta.com/chat/agent-b86b0d3a-c425-4c88-86b9-44171f178713?conversation=conv-88bfcb93-58e6-46ca-9240-f89211fd3dbf).

### DIFF
--- a/src/cli/subcommands/teleport.test.ts
+++ b/src/cli/subcommands/teleport.test.ts
@@ -57,6 +57,10 @@ function captureOutput(): {
   };
 }
 
+const NO_CHANNEL_ROUTES = {
+  listActiveChannelRouteNames: () => [],
+};
+
 describe("teleport subcommand", () => {
   test("prints help and returns 0", async () => {
     const out = captureOutput();
@@ -210,6 +214,7 @@ describe("teleport subcommand", () => {
     try {
       const exitCode = await withEnvironment(CLOUD_ENV, () =>
         runTeleportSubcommand(["cloud"], {
+          ...NO_CHANNEL_ROUTES,
           initializeSettings: async () => {},
           getLastSession: () => null,
           resolveAgentSandboxConnectionId: async (agentId, options) => {
@@ -264,6 +269,7 @@ describe("teleport subcommand", () => {
     try {
       const exitCode = await withEnvironment(CLOUD_ENV, () =>
         runTeleportSubcommand(["back"], {
+          ...NO_CHANNEL_ROUTES,
           initializeSettings: async () => {},
           getLastSession: () => null,
         }),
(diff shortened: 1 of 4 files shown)


cpacker:
Reworking this PR under LET-12212. The source-side pending-marker problem remains distinct from https://github.com/letta-ai/letta-code/pull/4477: failed readiness delivery must not leave later work blocked indefinitely.

The current tests manually resend teleport_request after reconnect, but the production caller does not resend it on idempotent API replay. They also switch the source connection ID without exercising the receiver’s recorded-source validation. They therefore do not establish successful reconnect recovery.

The rework will trace the current failure/timeout path, add deterministic coverage of subsequent input becoming usable after failed readiness delivery, and preserve the recorded-turn handling added since this branch was opened. It will not add an unsupported retry protocol or claim that replay alone completes a teleport.

PR #4353, opened 2026-09-10
refactor(memory): share frontmatter checks outside Git hooks [LET-12266]

cpacker:
## Summary

Memory validators need to apply the same frontmatter rules as the local Git hook. The existing package entrypoint exposes budgets and tree checks, but the frontmatter and legacy `read_only` checks still lived only in Bash.

Export `validateMemoryFileFrontmatter` through `@letta-ai/letta-code/memory-constraints` and make both local hook branches embed that same function. Callers supply the file contents, accepted contents, and trusted format. The hook keeps its existing path selection: the whole projected tree for v2, changed memory files for legacy. It validates frontmatter in one Node process, then runs the existing budget checks.

This preserves the current content rules. It does not add a read-only approval command, change deletion/rename handling, choose Git revisions, or enable hosted enforcement by itself.

## Verification

The parity tests exercise the public function and real Git commits for legacy and v2 files, including missing/unknown/duplicate fields, empty quoted values, and protected-field changes against an accepted base. The initial 28-case matrix passed against the pre-extraction Bash hook at `9ead5f04`; it passed again after extraction. No Git behavior is mocked.

The Node test bundles the public entrypoint and hook, imports them under Node, installs the resulting hook into a fresh repository, accepts valid v2 memory, and rejects a file missing its name.

The first extraction launched Node once per projected file. A real 100-file repository reproduced that regression with 100 launches at `ec489697`. The new test observes the real Node executable and requires one frontmatter-validation launch per commit, including an otherwise-empty commit and a rejected commit reporting invalid files at both ends of the batch. File contents and Git operations are not mocked.

- `bun test src/memory-frontmatter.test.ts src/agent/memory-git.precommit.test.ts src/agent/memory-git-v2-precommit.test.ts src/memory-constraints.test.ts src/agent/memory-constraints-audit.test.ts`: 84 passed.

## Breadcrumbs

- Requested in: LET-12266.
- Origin: existing content checks in [shared-memory validation](https://github.com/letta-ai/letta-code/pull/4102). This is an extraction, not a newly attributed regression.
- Follow-up to: [shared memory-constraints entrypoint](https://github.com/letta-ai/letta-code/pull/4318), which explicitly left frontmatter and read-only checks in the hook.
- Related, separate work: [read-only file controls](https://github.com/letta-ai/letta-code/pull/3616). This extraction does not adopt its approval command or deletion/rename behavior.
- Review follow-up: [batch Node startup across projected files](https://github.com/letta-ai/letta-code/pull/4353#discussion_r3978274709).

## AI Disclosure

This pull request was written by Letta Integration, an organization maintainer.

## AI Tool(s) Used

Letta Code, for implementation and verification.

### DIFF
--- a/src/agent/memory-git-hooks.ts
+++ b/src/agent/memory-git-hooks.ts
@@ -9,6 +9,7 @@
 
 import { chmodSync, existsSync, mkdirSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
+import { validateMemoryFileFrontmatter } from "@/memory-frontmatter";
 import { debugLog } from "@/utils/debug";
 import {
   MEMORY_CONSTRAINTS_CONFIG_PATH,
@@ -37,9 +38,6 @@ export const PRE_COMMIT_HOOK_SCRIPT = `#!/usr/bin/env bash
 # Validate frontmatter in staged memory .md files
 # Installed by Letta Code CLI
 
-AGENT_EDITABLE_KEYS="description"
-PROTECTED_KEYS="read_only"
-ALL_KNOWN_KEYS="description read_only limit"
 errors=""
 
 memory_layout_policy_file="$(git rev-parse --git-common-dir 2>/dev/null)/${MEMORY_LAYOUT_POLICY}"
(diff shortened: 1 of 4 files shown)


src/agent/memory-git-hooks.ts
```diff
@@ -154,7 +119,7 @@ if [ "$use_v2_validation" = "true" ]; then
           ;;
       esac
     fi
-    [ "$projected" = "true" ] && validate_v2_file "$file"
+    [ "$projected" = "true" ] && validate_memory_file "$file" "memfs-v2"
```
amelia-letta: this loop covers the whole tracked tree (git ls-files), not just staged files, and each projected file now spawns a fresh node process — node startup dominates (~30ms each, I measured ~3s for 100 files), so every commit in a large v2 repo gains seconds of pre-commit latency. worth batching all paths into a single node invocation.

PR #4372, opened 2026-09-11
test(tools): cover shell completion notifications on Windows

jibramnobhill:
## Summary

**Type: Test/CI.** Background shell commands should notify the owning agent and conversation when they finish. Both existing notification suites skipped Windows entirely, leaving their delivery, cancellation, and secret-redaction assertions without a Windows gate.

This enables the same 14 scenarios on Windows by replacing Unix-only fixtures with temporary JavaScript scripts launched with the running test executable and native PowerShell sleep commands. Each test owns its fixture directory. Explicit PowerShell exit propagation preserves the fixtures' chosen shell exit codes; it does not change the shell tools' default exit behavior. Windows completion deadlines allow for PowerShell startup.

Only two test files change. No production behavior changes or new live-provider tests.

## Review notes

- Start with the fixture commands and the removal of `describe.skipIf(isWindows)` in both notification suites. The existing delivery, routing, timeout, cancellation, duplicate-suppression, and output-scrubbing assertions remain.
- CI already discovers both files in the `tools` family and runs that family in its Windows, macOS, and Linux jobs. No workflow change is needed.
- Merge after current-head platform checks pass and the human-verification section is completed. Main risk: shell startup or fixture cleanup behaving differently under CI load.

## Verification

- `bun test src/tools/exec-command-notification.test.ts src/tools/bash-background-notification.test.ts --timeout 15000`: **14 passed, 72 assertions**, Windows 11 / Windows PowerShell 5.1 / Bun 1.3.14 (CI pin). The current revision also passes with Node absent from PATH.
- `bun run check`: **12/12 passed**, including TypeScript and Biome.
- Mutation checks: independently suppressing Windows Bash completion delivery and Windows exec completion delivery made the corresponding tests fail with missing-notification errors. Both production mutations were restored.
- The real CI impact selector includes both changed suites (`tools`, 90 selected tests).
- [Current-head CI](https://github.com/letta-ai/letta-code/actions/runs/34655314379) at `94732640`: **23 successful checks, none pending or failed**, including Windows x64, macOS arm64, Linux x64 and Linux arm64. Focused rereview requested from amelia-letta.
- Review correction: both fixtures use quoted `process.execPath`; each Windows failure case restricts PATH to its fixture directory. Reintroducing `node` makes both cases fail. Both inline review threads have commit/proof replies and are resolved.
- Scope ends at the real notification queue; no claim about a model consuming the notification or OS toast delivery. Follow-up review remains pending.

## Breadcrumbs

- Origin: the existing Bash and exec notification suites were introduced with Windows skips in [#3741](https://github.com/letta-ai/letta-code/pull/3741) and [#4219](https://github.com/letta-ai/letta-code/pull/4219). This closes a coverage gap, not a demonstrated production regression.
- Follow-up to: [#4220](https://github.com/letta-ai/letta-code/pull/4220), which added automatic Bash backgrounding scenarios to the existing suite.

## AI Disclosure

AI-assisted; the exact human-verification statement below remains pending.

## AI Tool(s) Used

Letta Code (Hadron): source/history investigation, test changes, local validation, mutation checks, and PR preparation.

## Human Verification

Pending human review and the repository's required verification statement.

### DIFF
--- a/src/tools/bash-background-notification.test.ts
+++ b/src/tools/bash-background-notification.test.ts
@@ -32,7 +32,7 @@ describe("Background bash completion notifications", () => {
     const script = join(fixtureDir, "command.cjs");
     writeFileSync(script, source);
     // Preserve the explicit shell exit status used by the old Unix fixtures.
-    return `node "${script}"${isWindows ? "; exit $LASTEXITCODE" : ""}`;
+    return `"${process.execPath}" "${script}"${isWindows ? "; exit $LASTEXITCODE" : ""}`;
   }
 
   const startBackground = async (args: {
@@ -148,6 +148,8 @@ describe("Background bash completion notifications", () => {
         "setTimeout(() => { console.error('automatic failure'); process.exitCode = 9; }, 100)",
       ),
       description: "Run failing automatic command",
+      // An otherwise empty PATH must not require a separate Node installation.
+      secretEnv: isWindows ? { PATH: fixtureDir } : undefined,
     });
 
     const notification = await waitForNotification(bashId);
--- a/src/tools/bash-background-notification.test.ts
+++ b/src/tools/bash-background-notification.test.ts
@@ -1,4 +1,7 @@
 import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { mkdtempSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
 import { bash } from "@/tools/impl/bash";
 import { kill_bash } from "@/tools/impl/kill-bash";
 import { backgroundProcesses } from "@/tools/impl/process_manager";
@@ -21,8 +24,16 @@ const isWindows = process.platform === "win32";
  * process registry and the queue bridge are module-level singletons, and a
  * shell torn down by a previous test can deliver its exit event here.
  */
-describe.skipIf(isWindows)("Background bash completion notifications", () => {
+describe("Background bash completion notifications", () => {
   let queued: QueuedMessage[] = [];
+  let fixtureDir: string;
+
+  function command(source: string): string {
+    const script = join(fixtureDir, "command.cjs");
+    writeFileSync(script, source);
+    // Preserve the explicit shell exit status used by the old Unix fixtures.
+    return `node "${script}"${isWindows ? "; exit $LASTEXITCODE" : ""}`;
+  }
 
   const startBackground = async (args: {
     command: string;
@@ -71,6 +82,7 @@ describe.skipIf(isWindows)("Background bash completion notifications", () => {
   };
 
   beforeEach(() => {
+    fixtureDir = mkdtempSync(join(tmpdir(), "bash notification-"));
     queued = [];
     clearPendingMessages();
     setMessageQueueAdder((message) => {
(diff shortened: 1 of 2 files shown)


src/tools/bash-background-notification.test.ts
```diff
@@ -21,8 +24,16 @@ const isWindows = process.platform === "win32";
...
-describe.skipIf(isWindows)("Background bash completion notifications", () => {
+describe("Background bash completion notifications", () => {
   let queued: QueuedMessage[] = [];
+  let fixtureDir: string;
+
+  function command(source: string): string {
+    const script = join(fixtureDir, "command.cjs");
+    writeFileSync(script, source);
+    // Preserve the explicit shell exit status used by the old Unix fixtures.
+    return `node "${script}"${isWindows ? "; exit $LASTEXITCODE" : ""}`;
```
amelia-letta: hardcoded `node` breaks on Bun-only Windows machines — no `node.exe` on PATH, so the PowerShell-spawned child fails to start and exits 1 (same trap as #4145, which is why shell-runner.test.ts uses `process.execPath`). passes on your machine and CI only because both have Node installed.

```suggestion
    return `"${process.execPath}" "${script}"${isWindows ? "; exit $LASTEXITCODE" : ""}`;
```

src/tools/exec-command-notification.test.ts
```diff
@@ -14,8 +16,16 @@ import {
...
-describe.skipIf(isWindows)("Exec command completion notifications", () => {
+describe("Exec command completion notifications", () => {
   let queued: QueuedMessage[] = [];
+  let fixtureDir: string;
+
+  function command(source: string): string {
+    const script = join(fixtureDir, "command.cjs");
+    fs.writeFileSync(script, source);
+    // Test notification routing, not PowerShell's default native-exit mapping.
+    return `node "${script}"${isWindows ? "; exit $LASTEXITCODE" : ""}`;
```
amelia-letta: same `node` hardcode as the bash suite — fails on Bun-only Windows where no `node.exe` is on PATH (#4145 pattern).

```suggestion
    return `"${process.execPath}" "${script}"${isWindows ? "; exit $LASTEXITCODE" : ""}`;
```

src/tools/bash-background-notification.test.ts
jibramnobhill: Addressed in 9473264083135287b13587dcaf742bf3d160ff46. Both fixtures now invoke the quoted `process.execPath`. The existing Windows failure case in each suite runs with PATH restricted to its temporary fixture directory, so Node installed elsewhere on the host cannot hide this dependency.

Verified 14/14 tests on Bun 1.3.14 + Windows PowerShell 5.1 with Node absent from the entire test runner PATH; all 12 repository checks passed. Reintroducing `node` makes both isolated-PATH cases fail. New-head CI is pending.

src/tools/exec-command-notification.test.ts
jibramnobhill: Addressed in 9473264083135287b13587dcaf742bf3d160ff46. Both fixtures now invoke the quoted `process.execPath`. The existing Windows failure case in each suite runs with PATH restricted to its temporary fixture directory, so Node installed elsewhere on the host cannot hide this dependency.

Verified 14/14 tests on Bun 1.3.14 + Windows PowerShell 5.1 with Node absent from the entire test runner PATH; all 12 repository checks passed. Reintroducing `node` makes both isolated-PATH cases fail. New-head CI is pending.

PR #4375, opened 2026-09-11
feat(doctor): investigate agent behavior from conversation evidence

devanshrj:
## Summary

Make `/doctor [symptom]` investigate agent behavior or audit memory in the primary conversation using the context-doctor skill. A user can open a fresh conversation, point doctor at an incident in another thread, and return to the saved findings. Plain `/doctor` combines a bounded memory health check with a recent-behavior review. The invocation supplies the user request and current environment; the skill owns workflow selection and distinguishes the investigation conversation from the target incident. Its small entrypoint holds shared guidance and routes to separate memory-audit and behavior-investigation references; focused requests load the relevant procedure, while plain `/doctor` reads both. Both the TUI and app-server listener use the normal turn pipeline.

- Expose existing evidence through `messages list/transcript --include-errors`, transcript truncation metadata, and `steps trace --agent … --step …`. Trace lookup uses normal CLI authentication, checks agent/step ownership, and reports unavailable API traces or unsupported local traces explicitly. Shared conversation-evidence examples select the target backend explicitly; read-only approval classification recognizes both `--backend value` and `--backend=value` for `local`, `api`, and `cloud`, while preserving strict mode, explicit permission rules, and write/operator checks; separate sections explain local transcript/compaction/provider metadata and API provider traces.
- Teach the skill to retrieve compact inventories before targeted messages, qualify historical-prompt claims, and return concise findings with evidence and verification limits. Replace the old format-specific doctor instructions and remove automatic skill-frontmatter rewriting. Keep ordinary output-redirection examples with writable destination paths chosen for the host. Use available filesystem and command capabilities rather than prescribing an editing tool, temporary-directory command, or JSON utility.
- Preserve direct memory audits for both memory formats: check structure, metadata, indexes/links, contradictions, redundancy, discoverability, and core-memory size. Structural defects do not require a conversation incident or provider trace. Preserve essential in-context instructions when moving supporting detail behind links.

## Verification

- `bun run check`: all 12 checks pass.
- `bun run build`: Bun source builds to the Node-targeted distribution.
- 242 focused tests pass across 11 files (Bun 1.3.14), including chronological listing and transcript exports to stdout, `--out`, and `--output`:

```sh
npm exec --yes --package=bun@1.3.14 -- bun test \
  src/cli/helpers/doctor-command.test.ts \
  src/websocket/listener/commands-doctor.test.ts \
  src/cli/subcommands/{messages,messages-evidence,steps,router}.test.ts \
  src/agent/prompt-assets.test.ts \
  src/tools/skill-tool.test.ts src/headless-client-skills.test.ts \
  src/permissions/read-only-shell.test.ts src/permissions/read-only-letta.test.ts
```

- 320 focused permission and approval tests pass across 11 files, covering the shared TUI/headless classifier, listener approval lifecycle, shell-tool aliases, explicit backend forms, and rejection of writes and unsafe shell operators. The backend regression suite reproduced 13 failures before the fix and passes afterward.

- Mock-provider interactive smokes under Bun 1.3.14 and Node 22.19 verify a fresh primary conversation runs `/doctor`, loads the skill entrypoint, reads both packaged workflow references, executes the evidence CLI, and saves its answer for later retrieval. Additional checks verify explicit local-backend listing/search/export, local raw-record discovery, and the unsupported local-provider-trace response against a disposable store. Reference contents are loaded on demand rather than embedded in the entrypoint. These smokes use scripted provider responses, so they verify tool access and packaging rather than model workflow selection; no live model calls were used.
- Before extracting the procedures into references, independent offline skill trials on deliberately broken memfs-v1 and memfs-v2 fixtures exercised an explicit memory audit and plain `/doctor` without history or provider traces. Final reports identified contradictory core policies, broken discovery paths, and format defects; distinguished skill metadata/index rules from memory rules; and avoided unsupported region choices or unnecessary trimming. All 17 fixture files remained unchanged. These trials used coding agents following the skill, not a full Letta model run, and do not establish general diagnostic accuracy.

## AI Disclosure

AI-assisted draft; full human review and editing are pending.

## AI Tool(s) Used

OpenAI Codex: implementation, tests, skill refinement, independent fixture audits, and PR preparation. Letta Code: user-run doctor investigation used to evaluate the behavior.

## Human Verification

The human-verification attestation remains unchecked; it has not been made on the reviewer's behalf.


src/permissions/read-only-shell.ts
```diff
@@ -177,9 +177,9 @@ const SAFE_MEMORY_COMMANDS = new Set([
 // letta CLI read-only subcommands: group -> allowed actions
 const SAFE_LETTA_COMMANDS: Record<string, Set<string>> = {
   memory: new Set(["status", "help", "backups", "export", "tokens"]),
-  // Legacy alias for `letta memory ...`.
   memfs: new Set(["status", "help", "backups", "export", "tokens"]),
   agents: new Set(["list", "help"]),
+  steps: new Set(["trace", "help"]),
```
amelia-letta: `--backend BACKEND` as the second token breaks this allowlist: `tokens[1]` is `--backend`, not the group, so every evidence command the skill prescribes requires approval instead of auto-allowing. the new steps test only asserts the bare form, which the skill never uses.

devanshrj: Fixed in f4e695fc. Read-only classification now recognizes `--backend value` and `--backend=value` for `local`, `api`, and `cloud`, including the flag positions accepted by the CLI, before checking the existing group/action allowlist. The steps test now uses the skill’s explicit API form. Added regressions for evidence commands, shell-tool aliases, and the shared interactive/headless approval classifier; strict mode, explicit deny/always-ask rules, mutations, and unsafe shell operators remain protected.

Output redirection to a file still follows normal write-approval rules.

Validation: 242 doctor tests, 320 permission/approval tests, all 12 repository checks, and the build pass.

PR #4408, opened 2026-09-13
fix(teleport): keep startup recovery from racing an active handoff [LET-12481]

cpacker:
## Summary

Teleport should continue a turn using the tool results supplied by the source computer. Startup approval recovery could instead start another turn on the destination before `teleport_continue` arrived, causing the handoff to fail with `Destination runtime is already processing`.

Keep automatic first-sync recovery for restarted listeners, but honor `recover_approvals: false` when replaying state for `runtime_start`. ChannelGateway now disables recovery when adopting an active delivery. Ordinary delivery and explicit approval recovery keep their existing behavior.

Before: `runtime_start` → pending source approvals recovered → destination starts a turn → `teleport_continue` rejected.

After: `runtime_start(recover_approvals: false)` → state/tool registration only → `teleport_continue` can start the turn. The first `sync` after a listener restart still recovers pending approvals, including when the readiness probe sends `recover_approvals: false`.

## Verification

The deterministic regression starts with a pending Bash approval and exercises the production recovery helper and turn lifecycle. With the old first-sync condition restored, destination setup takes the turn lease despite recovery being disabled; the test fails because `isProcessing` is true. It passes with this change. Backend approval lookup is stubbed and hydration is held behind a promise so timing is deterministic. Existing startup-recovery tests now send the real `sync` command through the message router.

- 86 gateway, teleport, recovery, and message-router tests pass.
- `bun run check` passes all 12 checks.
- A full managed-computer round trip remains to be verified with the updated caller and packaged runtime. This PR does not claim that integration result.

## Breadcrumbs

- Linear: LET-12481
- Origin: [startup approval continuation](https://github.com/letta-ai/letta-code/pull/4404). Its shared replay helper applied automatic recovery to `runtime_start` as well as `sync`.
- Letta conversation: [`conv-56e12686-1b0b-4cde-91fa-716ba3a839ac`](https://chat.letta.com/chat/agent-b86b0d3a-c425-4c88-86b9-44171f178713?conversation=conv-56e12686-1b0b-4cde-91fa-716ba3a839ac)

## AI Tool(s) Used

Letta Code.


cpacker:
CI follow-up: 6b789f6f corrects the protocol dispatch assertion to expect the new first-sync option. All 13 protocol tests and all 12 repository checks pass locally; commit hooks passed.

The other test failure, `listener turn lifecycle integration > teleport yields after persisting the current tool result`, also reproduces on the unchanged base 438f821a. I ran the same 274-file shared CI batch against that base: 2404 passed and this one test failed with `source must not start another model step`. It passes when run with the changed recovery suite. No unrelated lifecycle source changes are included here.

The previous review job failed during reviewer startup with zero turns. The new head is awaiting its own CI results.

PR #4413, opened 2026-09-13
revert(teleport): remove incomplete per-call recovery suppression [LET-12481]

cpacker:
## Summary

Revert [the incomplete teleport recovery fix](https://github.com/letta-ai/letta-code/pull/4408) so it can be replaced by [the pending-teleport approach](https://github.com/letta-ai/letta-code/pull/4410).

Both PRs address the same problem: SAD/restart recovery can start a competing turn on a teleport destination before the source's continuation arrives. The merged fix suppresses recovery during the setup calls. That is insufficient because a separate status `sync` can arrive before the teleport finishes and start recovery anyway.

The replacement uses the correct rule: while the destination is waiting for the teleport continuation, no sync may start a recovery turn. It records that wait when `runtime_start` carries `teleport_id`, blocks recovery across subsequent syncs, and clears the wait when the continuation arrives. The wait expires after 60 seconds so a later sync can recover if the continuation never arrives.

## Comparing the approaches

With the reverted approach:

```text
runtime_start(recover_approvals: false) → destination idle
independent status sync               → recovery starts a turn
teleport_continue                     → rejected as already processing
```

With the replacement:

```text
runtime_start(teleport_id) → destination waits for the continuation
independent status sync  → recovery does not start a turn
teleport_continue        → the source continuation can start
```

I reproduced the gap using the merged production recovery helper with a pending Bash approval: setup left `isProcessing: false`, then an ordinary first sync changed it to `true` before `teleport_continue`. Backend approval lookup was stubbed and hydration held behind a promise to keep the competing turn observable. The replacement's repeated-sync and expired-wait tests also passed. This difference is established by the code and those tests; it does not require waiting for a live teleport failure.

## Verification and merge order

- This is an exact revert of `f5611fd6bb8c48c7cdaf0531ff0041499ebaf3f2`. All six affected files match their pre-merge blobs.
- 85 focused gateway/recovery/teleport tests and 13 protocol tests passed. All 12 repository checks passed.
- This revert alone does **not** fix teleport. Merge [the replacement](https://github.com/letta-ai/letta-code/pull/4410) next; its caller must supply `teleport_id`. Full packaged-runtime teleport verification remains pending that paired change.

## Breadcrumbs

- Linear: LET-12481
- Origin: reverts [the per-call recovery suppression](https://github.com/letta-ai/letta-code/pull/4408), which incompletely addressed the interaction with [startup recovery](https://github.com/letta-ai/letta-code/pull/4404).
- Replacement: [block recovery while a teleport continuation is expected](https://github.com/letta-ai/letta-code/pull/4410).
- Letta conversation: [`conv-56e12686-1b0b-4cde-91fa-716ba3a839ac`](https://chat.letta.com/chat/agent-b86b0d3a-c425-4c88-86b9-44171f178713?conversation=conv-56e12686-1b0b-4cde-91fa-716ba3a839ac).

## AI Tool(s) Used

Letta Code.


cpacker:
CI classification for 7896ef59: Linux x64 and macOS each failed only `listener turn lifecycle integration > teleport yields after persisting the current tool result`, with `source must not start another model step`. This is the same shared-suite failure reproduced on unchanged base 438f821a before either teleport repair: the 274-file batch had 2404 passes and this one failure. The failing test and its turn-approval implementation are unchanged by this exact revert. Current Linux/macOS batches each have 2406 passes and that one failure.

The review workflow separately failed at startup with zero turns and no agent ID. No code review executed. I am retaining the exact requested revert rather than adding an unrelated test-suite repair.

PR #4419, opened 2026-09-13
fix(api): retry message requests when a gateway returns an HTML page [LET-12476]

cpacker:
## Summary

A message request that receives a gateway error page should enter request retry handling. The SDK currently accepts an HTTP 200 HTML response as a stream, yields no events, and leaves the caller trying to resume a run that the server never created.

`APIBackend.createConversationMessageStream` now checks for `text/event-stream` before returning the stream. An invalid content type aborts that response and throws `APIConnectionError`, which the existing request-retry policy already handles. Genuine SSE streams, including streams that later disconnect, retain their existing resume behavior.

Before: message POST → HTTP 200 HTML → empty stream → repeated OTID lookups with no matching run.

After: message POST → HTTP 200 HTML → connection error → existing request retry with the original input.

Example error: `Connection error: expected text/event-stream, received text/html (HTTP 200).`

## Verification

- A real-SDK regression fixture failed on the unchanged base for HTTP 200 HTML, JSON, and missing content type. All three are rejected with this change and classified as retryable connection errors.
- Valid SSE with a charset parameter preserves message chunks, request input, and acting-user headers. A genuine SSE-body disconnect does not cause another message POST in the API backend.
- 100 focused API-backend, stream-recovery, retry-policy, shutdown-retry, and listener-lease tests passed.
- All 12 repository checks and the package build passed.
- A Node-targeted fixture using the same API backend rejected HTML and accepted the subsequent SSE response. The built CLI also started under Node.
- Packaged-runtime restart E2E validation remains pending; this PR does not claim that end-to-end result yet.

## Breadcrumbs

- LET-12476.
- Origin: the existing [API backend facade](https://github.com/letta-ai/letta-code/pull/2034) delegated directly to the SDK without checking the response media type. No single recent introducing regression was identified.
- Related: [listener stream-resume retry](https://github.com/letta-ai/letta-code/pull/4401). This change keeps invalid responses out of that path.
- Generating Letta conversation: `conv-56e12686-1b0b-4cde-91fa-716ba3a839ac`.

## AI Tool(s) Used

Letta Code.


cpacker:
The review job failed before execution: its result reports duration_ms=0, num_turns=0, and agent_id=null. No review ran. Retried the failed job once; the code checks are tracked separately.

PR #4424, opened 2026-09-14
fix(reflection): preserve sandbox failure semantics

### LINKED ISSUE #4369: Reflection can falsely report no changes and consume history after sandbox temp-dir failure
### AI Disclosure

### AI Tool(s) Used

Letta Code was used to inspect the installed code paths, reproduce and test the local failure and mitigation, search existing issues, and draft this report. A human reproduced the user-visible failure, reviewed the technical summary and public-submission requirements, and approved this report.

### Human Verification

### Describe the bug

On macOS, a sandboxed automatic reflection subagent can be allowed to read and write its Letta and memory directories while still being denied access to the process-wide temporary directory returned by `os.tmpdir()`.

When the first background tool call tries to create Letta's private output directory with `mkdtempSync(join(os.tmpdir(), "letta-background-"))`, it fails before the command executes:

```text
EPERM: operation not permitted, mkdtemp '/var/folders/.../T/letta-background-XXXXXX'
```

The reflection child may then finish normally with a textual report explaining that it could not inspect the transcript or memory. Because the child produced a final result and no result event was marked as an error, the subagent manager returns `success: true`. Automatic reflection then records a successful no-change outcome, displays “Dreamed; no memory changes were needed,” advances its processed-range ledger, and removes history that was never inspected from automatic retry.

This is a false-success and data-retention problem: an infrastructure failure becomes indistinguishable from a genuine completed reflection that correctly found nothing durable to change.

#### Observed behavior

1. Automatic reflection starts normally.
2. The reflection child attempts its first background Bash/tool operation.
3. Letta tries to create a process-private output directory beneath `os.tmpdir()`.
4. The macOS sandbox rejects `mkdtemp(..., "letta-background-")` with `EPERM` before Bash executes.
5. The child cannot read the supplied transcript, inspect the memory repository, or make a memory decision.
6. The child returns a textual report describing the infrastructure failure, but the result event is not flagged with `is_error`.
7. In `src/agent/subagents/manager.ts`, the `state.finalResult !== null` branch derives success only from `!state.finalError`; the descriptive failure report is therefore returned as `success: true`.
8. Automatic reflection records a successful no-change outcome and advances past the unprocessed history.

In one verified occurrence, six consecutive automatic passes covering 87 completed assistant steps were consumed this way before detection. Every tool command in those passes failed before execution.

#### Expected behavior

- Reflection should receive a sandbox-writable background-output directory automatically.
- If reflection cannot read its transcript or memory, execute its tools, or complete its inspection, the pass should be recorded as failed and retryable.
- A failed pass must not be displayed as a genuine no-change reflection.
- The processed-range ledger should advance only after an explicit, verifiable terminal outcome: completed with memory changes or completed successfully with a deliberate no-changes decision.

### Letta Code version

First observed on `0.31.14`; the relevant code shapes were also present in `0.32.1` and `0.32.2`.

### Where did this happen?

CLI

Automatic reflection launched by a local Letta runtime.

### Operating system

macOS `26.6.2` on Apple Silicon

### Steps to reproduce

1. Run automatic reflection on macOS in a sandbox that allows the reflection worktree and memory paths but denies the normal `/var/folders/.../T` temporary directory.
2. Leave `LETTA_SCRATCHPAD` unset.
3. Trigger enough completed assistant steps for automatic reflection.
4. Observe the first background tool attempt fail with `EPERM ... mkdtemp ... letta-background-XXXXXX`.
5. Allow the child to return a final report explaining that it could not complete reflection.
6. Observe that the task envelope and reflection UI nevertheless report success/no changes and that the processed-range ledger advances.

No transcript, screenshots, identifiers, or diagnostic archive are attached to protect conversation privacy. The reproducible error shape is:

```text
EPERM: operation not permitted, mkdtemp '/var/folders/.../T/letta-background-XXXXXX'
```

#### Relevant code paths

The background-output path in `src/tools/impl/process_manager.ts` follows this shape:

```ts
function getBackgroundOutputDir() {
  const scratchpad = process.env.LETTA_SCRATCHPAD;
  if (scratchpad) return scratchpad;
  if (!backgroundOutputDir) {
    backgroundOutputDir = mkdtempSync(join(os.tmpdir(), "letta-background-"));
  }
  return backgroundOutputDir;
}
```

The result classification in `src/agent/subagents/manager.ts` follows this shape:

```ts
if (state.finalResult !== null) {
  return {
    report: state.finalResult,
    success: !state.finalError,
    error: state.finalError || undefined,
  };
}
```

This assumes that a child which emits final text without `is_error` completed its actual task. A reflection child can instead describe a terminal infrastructure failure in ordinary final text after all of its tools failed.

#### Verified local mitigation

The following mitigation worked across a real `0.32.1` to `0.32.2` update:

1. Set `LETTA_SCRATCHPAD=$HOME/.letta/tmp` before starting every Letta runtime so background output stays inside a sandbox-writable Letta directory.
2. Add an external exact-match compatibility guard that refuses to classify unmistakable reflection-infrastructure failure reports as successful.
3. Detect package updates, stale listeners, missing scratchpad inheritance, and unfamiliar code shapes before trusting reflection again.

This is intentionally conservative and local. Downstream code cannot reliably infer task success by matching prose in a model-generated final report.

#### Suggested upstream fix

1. When launching a sandboxed subagent, inject a writable `LETTA_SCRATCHPAD` inside its approved local root instead of relying on the host process's `os.tmpdir()`.
2. Preserve structured evidence of required tool failures and whether the child reached its task's completion contract. Do not reduce completion to “final text exists and `finalError` is empty.”
3. Distinguish explicit reflection outcomes such as `changes_committed`, `verified_no_changes`, and `failed_retryable`; advance the processed-range ledger only for the first two.
4. Show a failed/retryable reflection state when infrastructure prevented inspection instead of “no memory changes were needed.”
5. Add regression coverage for a denied `os.tmpdir()` with a writable injected scratchpad, explanatory final text after all tools fail, a failed task envelope, a retained retryable range, and a genuine successful no-change pass.

#### Related public context

- #2810 documents the earlier shared `/tmp/letta-background` collision and privacy problem.
- #2812 replaced that shared directory with a process-private `os.tmpdir()` + `mkdtempSync()` directory and added `LETTA_SCRATCHPAD` coverage.

The private-directory change is sound for ordinary processes. The remaining compatibility gap is that a sandboxed reflection process can have a narrower writable filesystem than the parent process's normal temporary directory.

#### Privacy

This report contains no private transcript, screenshots, account information, local paths, or agent, conversation, run, or message identifiers.

lyunyi9999-mini:
## Summary

Automatic reflection could report a successful no-change pass after a sandboxed background tool failed to create its output directory. The reflection transcript cursor then advanced even though the child never inspected the transcript or memory, so the work was not retryable.

The failure had two contributing paths:

- memory-subagent children inherited no writable scratch directory and `os.tmpdir()` is denied by the macOS filesystem sandbox;
- the stream parser treated a clean `result` envelope as success even when every recorded tool call ended with an error.

This change sets a child-only default `LETTA_SCRATCHPAD` under the approved transcript root (while preserving an explicit user value), records terminal tool-call statuses from both singular and plural `tool_return_message` shapes, and classifies a reflection as failed when all of its identified tool calls failed. Partial-success runs and incomplete streams retain their existing behavior.

The boundary is intentionally narrow: only reflection subagents use the all-tool-failure completion guard, and a tool call must have a matching explicit `success` or `error` terminal status before it participates in the guard. This prevents ordinary subagents and truncated/unknown streams from being reclassified accidentally.

## Verification

- RED: the new scratchpad and all-tool-failure regression tests failed against the unmodified implementation.
- GREEN: `bun test src/agent/subagent-env-composition.test.ts src/agent/subagents/subagent-stream.test.ts` (31 passed).
- Subagent and sandbox suites: 154 passed.
- `bun test src/agent/subagents/sandbox.test.ts` (9 passed).
- `bun test src/tools/impl/process-manager-output.test.ts` (3 passed, 1 existing skip).
- `bun run check` (12/12 checks passed).
- `bun run build` passed.
- `HOME=/private/tmp node ./letta.js --help` passed.
- `git diff --check` passed.

The upstream `src/cli/reflection-worktree-completion.test.ts` suite currently has six telemetry timing assertions failing on the unmodified `924bc00e` baseline; the same failures reproduce in isolation and are unrelated to this change.

## Risk and scope

- A reflection with at least one successful tool call is not affected.
- A tool failure that is not represented by an identified terminal `tool_return_message` remains on the existing completion path.
- No live Windows sandbox run was available on this macOS development host; cross-platform behavior is covered by the pure environment and sandbox-policy tests and CI.

Fixes #4369

## AI Disclosure

## AI Tool(s) Used

OpenAI Codex

## Human Verification


cpacker:
Thanks for the initial fix. #4527 carries it forward on current `main`, keeps each child scratchpad isolated, removes the generated scratchpad after the child exits, and also rejects reflection final text when no tool call succeeded. I credited this PR’s author in the replacement commit.

PR #4461, opened 2026-09-15
fix(slack): guard stale status writes and relinquishment [LET-12560]

cpacker:
## Summary

Slack activity should follow the current work and disappear when that work finishes. A delayed status request or an old thread's cleanup could overwrite newer activity. The controller now checks the thread and revision before writing, while still allowing an old thread to be cleared after newer work has finished.

The new `relinquish(source)` method stops pending writes from clearing activity transferred to another controller. It needs callers to use it; already-sent HTTP requests cannot be canceled by this change. This is a partial controller repair, not a complete fix for follow-up or cold-start interruptions.

## Before and after

- A refresh waits for credentials, then another title becomes current. Previously the old title could be sent afterward; now that obsolete refresh is discarded.
- Thread A's cleanup waits for credentials while thread B starts and finishes. The initial version of this PR discarded A's clear and left its activity visible. The repair lets A's cleanup finish without clearing active work in the same thread.
- An old thread posts a reply after a newer thread becomes active. Its cleanup now leaves the newer thread's activity intact.

## Verification

Deterministic tests use the real controller with held credential lookups and HTTP completions. The two added delayed-cleanup cases failed on the initial PR head and pass after the repair. A separate probe exercised the real gateway's queued lifecycle hooks alongside direct status activation and reply cleanup. These reproduce request ordering, not Slack-client rendering.

Ran 49 focused tests, `bun run check`, and `bun run build`. Both delayed-cleanup cases also pass through the built Node package export. Native Slack-client rendering remains unverified, so this stays draft.

## Breadcrumbs

- Follow-up: LET-12560.
- Origin: [assistant-status implementation](https://github.com/letta-ai/letta-code/pull/3279); [the shared refresh path](https://github.com/letta-ai/letta-code/pull/4380). No single introducing change is asserted for all status-write races.
- The delayed-cleanup regression was introduced by this PR's initial commit `840d25890d2824e8ade5da4fda06de079c7d9365` and repaired in `e88af954`.
- Review and repair conversation: [conv-e6fb5c1c-5ff3-4f78-abcf-69a63701e665](https://chat.letta.com/chat/agent-b86b0d3a-c425-4c88-86b9-44171f178713?conversation=conv-e6fb5c1c-5ff3-4f78-abcf-69a63701e665).

## AI Disclosure

AI-assisted implementation and independent AI review for internal Letta work. No human review is claimed. Titan authored the initial controller changes; Big Chungus reviewed and repaired the delayed-cleanup regression.

## AI Tool(s) Used

Letta Code for implementation, reproduction, validation, and review. Earlier Codex and Claude Code review attempts failed authentication before producing a review.


src/channels/slack/status-controller.ts
```diff
@@ -360,6 +489,7 @@ export function createSlackStatusController(params: {
     clear(): void {
       for (const timer of keepaliveByConversation.values()) clearTimeout(timer);
       stateByConversation.clear();
+      revisionByConversation.clear();
       sourceByConversation.clear();
```
amelia-letta, line 493: `clear()` resets `revisionByConversation` but not `relinquishedThroughRevision` — after a clear, a stale revocation cutoff can suppress the next owner's writes to the same thread until the global revision counter climbs past the old cutoff. Suggest clearing it alongside the other maps.

```suggestion
      clearedStaleReplyKeys.clear();
      relinquishedThroughRevision.clear();
```

PR #4462, opened 2026-09-15
test(channels): isolate persisted control-request routes

cpacker:
## Summary

Unit tests must not leave channel routes for another suite to load. The control-request tests wrote a Slack route for `agent-1` / `conv-1` to the shared test home. The preload cleared the in-memory route map, but the teleport test read the persisted route and rejected the teleport.

Keep control-request routes in memory with the existing load/save test overrides. Add a direct channel-guard assertion to the teleport isolation test so a future leak names the blocking channel instead of only reporting a missing pending teleport. Production code and test retries are unchanged.

## Verification

- On head `460e662a`, all 20 cross-platform Unit jobs pass. API/package verification remains incomplete: the current CI run includes HTTP 503/520 failures from the live API and a startup exit without a captured cause. These are not claimed fixed by this test-only change.

- Reproduced with `bun test src/channels/registry-controls.test.ts src/websocket/listener/channel-route-isolation.test.ts`: before the fix, 12 passed and the teleport test failed; after it, all 13 passed in ten consecutive runs.
- Ran `node scripts/run-unit-tests.cjs --shard 1/4` before and after. The shared batch changed from 1,822 passing tests and two failures to 1,823 passing tests and one failure. The teleport failure is gone. The missing-shell-launcher test still fails in this sandbox, and the isolated MCP OAuth test cannot access libsecret; both failed before this change. CI remains the cross-platform check.

## Breadcrumbs

- Origin: the persisted Slack fixture exists in [the channel test split](https://github.com/letta-ai/letta-code/pull/3295) and earlier [Slack progress tests](https://github.com/letta-ai/letta-code/pull/3029). This is an older fixture leak, not a new teleport behavior change.
- Follow-up to: [route-global cleanup and the teleport isolation test](https://github.com/letta-ai/letta-code/pull/4429), which clear memory but intentionally do not delete persisted channel files.
- Exposed by: [new memory checkout tests](https://github.com/letta-ai/letta-code/pull/4438). At `51467f82`, the two suites belonged to shards 4 and 2; at `b69f80cb`, both belong to shard 1.
- Root-cause evidence: [main CI run 34924403096](https://github.com/letta-ai/letta-code/actions/runs/34924403096), Unit shard 1 on macOS and Ubuntu, failing `the next test can teleport the same agent and conversation`.
- Letta conversation: [conv-de777acf-405b-4c0b-96ca-e642aed4fb23](https://chat.letta.com/chat/agent-b86b0d3a-c425-4c88-86b9-44171f178713?conversation=conv-de777acf-405b-4c0b-96ca-e642aed4fb23).

## AI Tool(s) Used

Letta Code.


src/channels/registry-controls.test.ts
```diff
@@ -6,14 +6,22 @@ import {
...
 import type {
   ChannelAdapter,
   ChannelControlRequestEvent,
   InboundChannelMessage,
 } from "@/channels/types";
 
 beforeEach(() => {
+  // These tests exercise dispatch, not persistence. Keep their routes in memory
+  // so another suite cannot reload them after the preload clears the route map.
+  __testOverrideLoadRoutes(() => null);
```
amelia-letta, line 23: Every other test file that uses route overrides clears them in afterEach. Consider adding cleanup for consistency.

PR #4466, opened 2026-09-15
fix(agent): notify remote completion from existing Cloud status

cpacker:
Remote Agent tasks should keep working while the calling agent handles other messages, and their result should remain collectible after a temporary network failure. The previous implementation kept a submitting CLI process connected until the listener finished, with a one-hour waiting ceiling. Losing that process could report failure while remote work continued.

Computer-routed launches now submit with `--no-wait`, return an enqueue receipt, and exit. The parent harness watches Cloud's existing Super Run status feed. Transient read failures back off and retry without resubmitting input or imposing a one-hour ceiling. Individual model-run failures during listener recovery do not end tracking.

Cancellation removes queued input or reconnects to the selected listener to abort only the run belonging to that input. The error-envelope parser now reads `error` as well as `result`, and process failures report their exit code/signal rather than only startup stderr.

Before: Agent returns a background task ID; its CLI child stays connected to the remote listener until completion. A failed read can end collection while execution continues.

After: Agent returns a background task ID; its CLI child exits after acceptance. The harness observes the accepted Super Run leaving the active feed, or receives its terminal status, then emits the existing task notification. The existing latest-Super-Run read recovers fast completion when it still identifies this send. It never uses a newer send's outcome as this task's outcome.

Reply collection is best-effort and only reads run IDs observed belonging to this input. If the reply is unavailable, completion still notifies the parent with a `letta messages list` command. It does not claim the work failed or launch another copy. Completion is a Cloud activity signal, not certification that the agent accomplished its requested goal.

No Cloud API changes, migration, or new listener protocol are required. Same-computer launches and ordinary A2A sends are unchanged.

## Testing

Deterministic tests exercise an already-finished send, failed reads, provider recovery, default conversations, a newer send overtaking the latest read, reconnects, and missing reply text. Existing listener cancellation tests verify only the addressed run is aborted. Launch-envelope coverage checks receipt identity, failure text, and completed usage after consolidating the output paths.

Live check: ran the branch's Agent child-launch CLI path with `--computer cloud --no-wait` against deployed Cloud. It returned a queued receipt and exited 0; the remote listener produced the requested smoke-test reply. Running the branch observer with initialized production API helpers after completion returned the completion notice and conversation-read command. The suggested messages command retrieved the expected reply. No Cloud endpoint change or deployment was used. This exercised the child-launch and result-observer paths directly, not an LLM-issued Agent tool call. An earlier explicit-connection attempt failed because that computer was not visible to the test credential, before enqueue.

## Agent-facing description

The Agent tool description now says:

> Computer-routed tasks are submitted asynchronously and tracked through Cloud's existing Super Run status feed. Temporary status-read failures retry in the background; there is no one-hour tracking ceiling. Completion still notifies you if the reply could not be collected; use `letta messages list` to read the conversation rather than launching the task again. Stopping a task cancels its queued input or its executing listener run.

## Breadcrumbs

- LET-12149
- Origin: follow-up to [listener-backed Agent launches](https://github.com/letta-ai/letta-code/pull/4389), using [CLI enqueue receipts](https://github.com/letta-ai/letta-code/pull/4383). The original incident's process-exit cause was not established; this PR does not claim OOM.
- Original implementation: [conv-1871a72d-1153-43c0-bc8a-532b0f2cff26](https://chat.letta.com/chat/agent-b86b0d3a-c425-4c88-86b9-44171f178713?conversation=conv-1871a72d-1153-43c0-bc8a-532b0f2cff26)
- Review and repair: [conv-63258eae-8b44-4400-b6a1-7b9a982b3744](https://chat.letta.com/chat/agent-b86b0d3a-c425-4c88-86b9-44171f178713?conversation=conv-63258eae-8b44-4400-b6a1-7b9a982b3744)


src/agent/subagents/remote-turn-wait.ts
```diff
@@ -0,0 +1,253 @@
...
+        firstEvent: first.value,
+        runIds,
+        signal,
+        pollMs: deps.pollMs ?? DEFAULT_POLL_MS,
+        now: deps.now,
+        retrieveRun: (runId) => read((s) => deps.retrieveRun(runId, s)),
+        listRunMessages: (runId) => read((s) => deps.listRunMessages(runId, s)),
+        latestSuperRun: () =>
+          receipt.conversation_id === "default"
+            ? Promise.resolve(null)
```
amelia-letta: `latestSuperRun` is nulled for `conversation_id === "default"`, but the review notes rely on the completed-super-run check to end the wait — for a default-conversation send whose run mapping never arrives (stream dropped before the mapping), the waiter has no terminal path and polls forever since there is no ceiling.

**Resolved:** addressed by the waiter rewrite in 32ee7dfd, which ends tracking from the active-feed snapshot itself.

amelia-letta: Addressed by the rewrite in 32ee7dfd: the waiter now ends tracking from the active-feed snapshot itself (a send omitted from `active_super_runs` is finished), so a default-conversation send no longer depends on a run mapping arriving, and there is still no ceiling needed.

PR #4472, opened 2026-09-15
fix(a2a): carry verified sender assertions [LET-12546]

carenthomas:
## Summary

A Cloud sandbox needs the current human sender when it discovers or messages another agent. Shared-agent discovery now forwards the sender ID through [PR 4483](https://github.com/letta-ai/letta-code/pull/4483). This change also carries the Cloud-minted assertion that lets Cloud verify that identity.

The assertion follows listener inputs, queued turns, message streams, A2A discovery and delivery, conversation creation and forks, schedules, shell-launched CLI commands, and subagents. Queued messages from different humans are kept in separate turns so one sender's permissions never authorize another sender's input. Neutral notifications can still join a turn with one human's messages.

The listener uses its prepared execution context for sender identity on both initial requests and continuations. Assertions remain optional for compatibility with current servers. This CLI change must be released before the companion Cloud enforcement change tracked under LET-12546; merging it alone does not enable that enforcement.

## Verification

- `bun run check`: all 12 checks passed after integrating current main.
- Focused listener, stream, A2A, shell environment, queue authorship, protocol, and command tests: 85 passed.
- Notification steering, notification identity, and message-router tests: 27 passed.
- Turn lifecycle, send leases, recorded-turn recovery, and prepared-context sender tests: 28 passed.

Live cross-service assertion verification still requires the companion Cloud change. No production deployment was performed.

## Breadcrumbs

- Requested in: LET-12546
- Original Letta conversation: [`conv-27d81a0c-79de-4328-a8be-38f761d82cff`](https://chat.letta.com/chat/agent-cd664b86-4d28-49b7-8ad6-60677eaff9be?conversation=conv-27d81a0c-79de-4328-a8be-38f761d82cff)
- Integration and validation: [`conv-0c2142fd-7b5e-4331-828a-5aca776df81b`](https://chat.letta.com/chat/agent-b86b0d3a-c425-4c88-86b9-44171f178713?conversation=conv-0c2142fd-7b5e-4331-828a-5aca776df81b)
- Origin: [PR 3537](https://github.com/letta-ai/letta-code/pull/3537) introduced shared-agent discovery without sender delegation.
- Follow-up to [PR 4483](https://github.com/letta-ai/letta-code/pull/4483), which forwards the sender ID for shared-agent discovery.
- Follow-up to [PR 4456](https://github.com/letta-ai/letta-code/pull/4456), which forwards the requesting user into Cloud schedule turns.

## AI Tool(s) Used

Letta Code (Amelia and Big Chungus the 4th).


src/websocket/listener/queue.ts
```diff
@@ -63,9 +63,12 @@ export function getQueueItemsScope(items: QueueItem[]): {
 }
 
 function hasSameQueueScope(a: QueueItem, b: QueueItem): boolean {
+  const aActingUserId = a.kind === "message" ? a.actingUserId : undefined;
+  const bActingUserId = b.kind === "message" ? b.actingUserId : undefined;
```
amelia-letta: this treats every non-message item as anonymous even when its queue item carries a sender. same-user task and monitor notifications now split from queued steering, delaying that input and dropping the notification’s delegated sender on its separate turn, which matches the current integration-test failures.

src/tools/impl/shell-env.ts
```diff
@@ -360,6 +364,10 @@ export function getShellEnv(): NodeJS.ProcessEnv {
   if (actingUserId) {
     env[ACTING_USER_ID_ENV] = actingUserId;
   }
+  const actingUserAssertion = getRuntimeActingUserAssertion();
+  if (actingUserAssertion) {
+    env[ACTING_USER_ASSERTION_ENV] = actingUserAssertion;
```
amelia-letta, line 369: the nested headless path still drops this value: `headless.ts` only puts `resolveActingUserId()` in the remote launch scope, and `headless-listener-launch.ts` only forwards that id to enqueue. any `letta -p --computer ...` launched from this shell will send an unverified acting-user header and lose delegated access.

src/websocket/listener/queue.ts
```diff
@@ -63,9 +63,14 @@ export function getQueueItemsScope(items: QueueItem[]): {
 }
 
 function hasSameQueueScope(a: QueueItem, b: QueueItem): boolean {
+  const compatibleSender =
+    a.kind !== "message" ||
```
amelia-letta: If a notification is first, this is true for every later message, so notification → human A → human B is consumed as one batch and `buildQueuedTurnMessage` sends B under A's assertion. Please compare each message against the batch's first message sender instead of only `firstQueuedItem`.

src/websocket/listener/queue.ts
carenthomas: fixed in eb94bb0c. task and monitor completions remain intentionally bearer-authored, while neutral queue items can coalesce with the first human sender and later human senders split into separate batches.

src/websocket/listener/queue.ts
carenthomas: fixed in eb94bb0c. the batch now records the first attributed message sender even when neutral items lead, and stops before any later different sender.

src/tools/impl/shell-env.ts
carenthomas: fixed in eb94bb0c. the remote headless scope now carries `acting_user_assertion`, and listener enqueue forwards it with `actingUserId`; the launch test locks both fields.

PR #4478, opened 2026-09-15
fix(channels): restart gateway after unexpected exits

just-cameron:
## Summary

A ChannelGateway child could exit after startup while the parent listener remained alive, leaving enabled channels unavailable until the user restarted Letta Code or Desktop.

This change supervises post-ready gateway exits. It launches one replacement at a time with bounded exponential backoff, waits for each replacement to report ready, and resets the retry budget only after a stable minute. In-flight service commands fail once when their child exits and new commands remain unavailable until the replacement is ready.

The same lifecycle is written to the existing local session log and emitted as aggregate `channel_gateway_lifecycle` telemetry. Telemetry includes the lifecycle event, retry attempt, timing, exit code or signal, restore mode, channel types, platform, and version. It excludes raw errors, account identifiers, credentials, and message content.

This PR covers post-ready gateway loss. It does not address the separate cold-start path where channel restoration never begins.

## Before and after

Before:

1. `ChannelGateway` reports `CHANNEL_GATEWAY_READY`.
2. The child exits.
3. The supervisor rejects in-flight commands and reports the exit.
4. The listener stays alive in Desktop restore mode, but channels remain unavailable until the user restarts Desktop.

After:

1. `ChannelGateway` reports `CHANNEL_GATEWAY_READY`.
2. The child exits and in-flight commands fail once.
3. The supervisor schedules replacements after 1s, 2s, 4s, 8s, and 16s.
4. A replacement must report `CHANNEL_GATEWAY_READY` within 30 seconds before service commands resume.
5. Sixty stable seconds reset the retry budget. Five failed replacements produce a terminal exhaustion event instead of an unbounded crash loop.

## Review notes

Explicit `--channels` listeners now exit only after the replacement budget is exhausted, rather than after the first child exit. Desktop restore mode remains alive and records the exhausted state in its session log and telemetry.

## Verification

The red-first regression test starts a real Node child that reports ready and exits with code 7. On unmodified `origin/main`, the test fails after observing that no second child starts. With this change, the same test observes the replacement process and passes.

Additional deterministic process-boundary tests cover exponential backoff, retry exhaustion, stable-period reset, replacement readiness timeout, exactly-once in-flight command failure, lifecycle-event order, and shutdown during backoff.

- [x] `bun test src/channels/gateway-supervisor.test.ts` (7 pass)
- [x] `LETTA_CODE_TELEM=1 bun test src/telemetry/flush-batching.test.ts` (7 pass)
- [x] Listener integration tests (17 pass)
- [x] Gateway test suite repeated five times (35/35 pass)
- [x] `bun run check` (12/12 checks pass)
- [x] `bun run build`

## AI Disclosure

## AI Tool(s) Used

Letta Code performed source tracing, implementation, regression-test construction, validation, and pull-request drafting.

## Human Verification

## Breadcrumbs

- Requested in: LET-11906
- Letta conversation: [`conv-a39ce41e-181c-4f3e-9649-93e1ec9a5207`](https://chat.letta.com/chat/agent-ec05a4e0-f50c-47fd-8e7b-960f081ab42e?conversation=conv-a39ce41e-181c-4f3e-9649-93e1ec9a5207)
- Origin: The ChannelGateway process boundary was introduced in https://github.com/letta-ai/letta-code/pull/3633. Its supervisor reported post-ready exits but did not relaunch the child; history did not identify a later regression.
- Related: Follow-up to the false-disconnect correction in https://github.com/letta-ai/letta-code/pull/4291


src/cli/subcommands/listen.tsx
```diff
@@ -578,10 +578,32 @@ export async function runListenSubcommand(argv: string[]): Promise<number> {
...
           },
+          onLifecycleEvent: (event) => {
+            telemetry.trackChannelGatewayLifecycle({
+              lifecycle_event: event.kind,
+              restart_attempt: event.restartAttempt,
+              max_restart_attempts: event.maxRestartAttempts,
+              restore_mode: restoreEnabledChannels
+                ? "enabled_accounts"
+                : "explicit_channels",
+              channel_types: channelNames,
```
amelia-letta: In Desktop’s `enabled_accounts` mode, `channelNames` is always `[]`, so the new lifecycle telemetry never records the restored channel types for the primary recovery path.

just-cameron:
Addressed Amelia’s restored-channel telemetry comment in 52260e76c6b4dff9314f9b4ec45d2888f6cffa15. Restore mode now enumerates enabled account channel IDs for `channel_types`; explicit `--channels` behavior is unchanged.

Local validation: `bun run check` 12/12, gateway/listener suites 24 pass, telemetry suite 8 pass, and TypeScript clean.

src/channels/gateway-telemetry-types.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export function resolveChannelGatewayTelemetryTypes(options: {
+  restoreEnabledChannels: boolean;
+  channelNames: readonly string[];
+  restoreAgentScope?: ChannelRestoreAgentScope | null;
+}): string[] {
+  if (!options.restoreEnabledChannels) {
+    return [...options.channelNames];
+  }
+  return listEnabledChannelIds({
```
amelia-letta, line 19: `listEnabledChannelIds()` reads the parent listener’s process-local account cache, but account create/update commands run in the ChannelGateway child. Once the first lifecycle event loads this cache, later child-side changes won’t be reflected, so subsequent restart telemetry can still emit stale `channel_types`; this needs a disk refresh or child-owned snapshot instead.

just-cameron: Fixed in 36433592. `resolveChannelGatewayTelemetryTypes` now reloads each channel's accounts from disk before listing. The test now models the child changing disk behind the listener cache, and fails without the reload.

amelia-letta:
@cpacker Should replacement gateway processes reuse an environment snapshot captured when the supervisor starts, rather than re-reading mutable `process.env` on each retry, so recovery preserves the original execution context?

PR #4492, opened 2026-09-16
fix(cli): keep CJK IME commits intact by coalescing same-tick stdin reads

### LINKED ISSUE #4493: CLI input drops characters and leaves the caret one position before the end when committing Japanese IME text
## AI Disclosure

- [x] This issue was written with AI assistance and reviewed and edited by a human

## AI Tool(s) Used

- opencode (AI coding agent; model `deepseek-v4.1-flash`) — analyzed the captured debug logs and the CLI input code and drafted this report.

## Human Verification

## Describe the bug

When typing Japanese with an IME and confirming a candidate (一括変換), the CLI input is corrupted: the caret ends up **one position before the end**, and for longer strings **part of the committed text is dropped or committed in separate pieces**. Typing again then lands in the wrong place, so the final text is wrong.

`LETTA_DISABLE_KITTY=1` does not change the behavior. Copy/paste of the same text is fine, so only the IME commit path is affected.

## Steps to reproduce

1. Start `letta`.
2. Type `にほんご`, convert to `日本語`, press Enter to confirm.
   - The input shows `日本語` but the caret sits **before `語`**.
3. Type a longer phrase in one conversion, e.g. `日本語の入力中に変換がおかしくなるのが問題だ`.
   - The commit is delivered in pieces (`日本` / `語の` / `入力中に...`), the input appears to auto-confirm early, and continuing to confirm inserts the remaining chunks at the wrong position.

## Expected behavior

The full committed string is inserted at the caret and the caret is placed at the end of the committed text.

## Actual behavior

- Caret is left one character before the end after a multi-character commit.
- Longer commits lose or misplace chunks.

## Root cause

`vendor/ink/build/components/App.js` `handleReadable` treats **each stdin read as its own input event**. A single IME commit is delivered as several reads in the same tick (e.g. `日本` then `語`), so the focused text input receives a burst of separate events. Under that burst:
- the value-sync effect in `PasteAwareTextInput` runs for an older commit after the handler already advanced the caret, clamps `caretOffsetRef` back to the stale value's length and returns a bogus `nudgeCursorOffset` (caret ends one short); and
- the vendored `ink-text-input` handler computes the next state from a render snapshot, so a following event can drop the previous chunk.

## Letta Code version

0.32.11

## Where did this happen?

CLI (iTerm2)

## Operating system

macOS 26.5.2 (arm64), iTerm2 3.6.11, macOS Japanese IME (also with Google 日本語入力)

## Relevant logs, screenshots, or recordings

<details>
<summary>Debug log excerpt</summary>

Captured with:

```
$ LETTA_DEBUG_KEYS=1 LETTA_DEBUG_INPUT=1 letta 2>/tmp/ime.log
```

Each logical commit is delivered as several reads (one `[debug:raw-input]` per read), i.e. several input events in the same tick:

```
# 1) one logical commit: 日本語   (two reads)
[debug:raw-input] len=2 hex: 0x65e5 0x672c                                   # 日本
[debug:raw-input] len=1 hex: 0x8a9e                                          # 語

# 2) one logical commit: 日本語の入力   (two reads)
[debug:raw-input] len=1 hex: 0x65e5                                          # 日
[debug:raw-input] len=5 hex: 0x672c 0x8a9e 0x306e 0x5165 0x529b              # 本語の入力

# 3) one logical commit: 日本語の入力中に変換がおかしくなるのが問題だ   (three reads)
[debug:raw-input] len=2 hex: 0x65e5 0x672c                                   # 日本
[debug:raw-input] len=2 hex: 0x8a9e 0x306e                                   # 語の
[debug:raw-input] len=18 hex: 0x5165 0x529b 0x4e2d 0x306b 0x5909 0x63db 0x304c 0x304a 0x304b 0x3057 0x304f 0x306a 0x308b 0x306e 0x304c 0x554f 0x984c 0x3060   # 入力中に変換がおかしくなるのが問題だ
```

Each read is then dispatched as its own input event (`[debug:ink-keypress]`, printed once per active `useInput` listener):

```
[debug:ink-keypress] raw=0x65e5 0x672c name="" seq="日本" key={escape:false,tab:false,shift:false,ctrl:false,meta:false}
[debug:ink-keypress] raw=0x8a9e      name="" seq="語"  key={escape:false,tab:false,shift:false,ctrl:false,meta:false}
```

</details>

## Additional context

Related but different: #2644 (CLOSED) is about the IME candidate window misalignment, not text/caret corruption.

shimpeiws:
## Summary
Fixes Japanese (and other CJK) IME input corruption: after a candidate is confirmed (一括変換), the caret was left one position before the end of the committed text, and longer commits dropped or misplaced chunks.

`LETTA_DISABLE_KITTY=1` did not change it, and copy/paste was unaffected.

This is an upstream Ink bug, not a Letta-specific one — see [Upstream](#upstream).

## Root cause
A single IME commit is delivered by macOS terminals as several stdin reads in the same tick (e.g. `日本` then `語`). `App.handleReadable` (vendored Ink) dispatched each read as its own input event, so the focused text input handled a burst of separate events:
- the value-sync effect in `PasteAwareTextInput` ran late for an older commit, clamped `caretOffsetRef` to the stale value's length and returned a bogus `nudgeCursorOffset` (caret one short); and
- the vendored `ink-text-input` handler computed the next state from a render snapshot, so a following event could drop the previous chunk.

Debug logs showing a single commit split across reads are in #4493.

## Upstream

The same failure is tracked upstream as [vadimdemedes/ink#759 — "Input Lag, Characters Drop and Cursor position issue in languages using IME(CJK)"](https://github.com/vadimdemedes/ink/issues/759) (open since 2025-08-30). It has stalled for lack of a minimal reproduction — tats-u asked for one on 2025-11-19. The issue body independently proposes this shape of fix: "temporarily buffer the input until a final, composed character is received".

This PR provides that missing minimal reproduction (the split `日本` → `語` reads plus debug logs in #4493), the root cause, and a small fix. Because it lands in vendored Ink, treat it as a **stopgap until we bump Ink to a release that fixes #759** — it must be re-applied when re-vendoring.

## Fix
Buffer plain-text reads briefly (16 ms) in the vendored Ink `App.handleReadable` and dispatch them as a single input event. Control sequences and paste flush the buffer first so ordering is preserved.

### The 16 ms buffer is intentional — scope and cost

Reviewers will notice every printable read now passes through a 16 ms window. That is deliberate, and its cost is bounded:

- **Control keys are never delayed.** Any read containing a C0 control character — Enter, Tab, Esc, arrows, Ctrl-\*, Backspace — flushes the buffer synchronously and dispatches immediately, so control sequences keep their exact position relative to buffered text (covered by the `control keys between text preserve order` test).
- **Paste is unaffected.** Bracketed paste and the non-bracketed paste fallback both flush the buffer before emitting, so paste ordering is unchanged.
- **Plain-text latency.** For an isolated burst the added latency is at most one 16 ms window (~1 frame at 60 Hz). While text keeps arriving faster than 16 ms apart, dispatch waits for the next gap; this reuses the existing `FALLBACK_NORMAL_MS = 16` window already used for non-bracketed paste detection.
- **Trade-off.** Text arriving faster than ~16 ms apart is coalesced into a single input event, the same delivery model as paste — so consumers already receive multi-character inserts today.

### Note on the vendored-Ink red-zone checklist

`AGENTS.md` ("Vendored Ink Patches (Red Zone)") describes `scripts/postinstall-patches.js` as applying patches "by exact string match" with a "throw-on-missing guard". That is stale: the script copies `vendor/**` over `node_modules` wholesale with `copyFileSync` (`scripts/postinstall-patches.js:108-127`) — no string matching, and only a warn-on-failure catch (its only `writeFileSync` is the unrelated Bun-shebang step). The invariant this change actually has to preserve is that the copied filenames/paths are unchanged, and they are (`vendor/ink/build/components/App.js` → `ink/build/components/App.js`). Happy to send a separate docs-only PR to fix the AGENTS.md wording.

## Verification
- `bun run check` — 12/12 checks pass.
- Vendored file is live: `vendor/ink/build/components/App.js` and the installed `node_modules/ink/build/components/App.js` are byte-identical (sha1 `6fbfe27b088614649c0b34739208118e2efb2e1e`), i.e. the postinstall copy has run and this is the code `bun run dev` executes.
- `src/cli/components/PasteAwareTextInput.test.tsx` — 4 tests pass. The two split-commit tests fail on the original code and pass with the fix.
- `bun test src/cli` — only the same pre-existing failures as on `main` (reflection / model catalog / provider selector); no new failures.
- Manual: reproduced and verified on macOS 26.5.2 + iTerm2 3.6.11 with the macOS Japanese IME (`日本語`, `日本語の入力`, `日本語の入力中に変換がおかしくなるのが問題だ`).

## AI Disclosure

## AI Tool(s) Used

- opencode (AI coding agent; model `deepseek-v4.1-flash`) — investigated the bug, reproduced it with an instrumented harness, implemented the fix and regression tests, and drafted this description.
- OpenAI Codex CLI — reviewed the diff and suggested the control-ordering test and a comment clarification.

## Human Verification

Fixes #4493


shimpeiws:
Upstream context on this one: it isn't specific to Letta Code — it's an ink input-handling issue, and it's already tracked upstream. I've posted a reproduction and root-cause analysis to vadimdemedes/ink#759, which had been open since 2025-08 without a minimal repro.

Summary of that comment:

- A single IME commit arrives on `stdin` as several reads in the same tick (e.g. `日本` then `語`).
- `handleReadable` emits one input event per `read()` chunk, so the consumer sees a burst of events before React re-renders — that's what corrupts the value and the caret.
- The same shape is present on current ink (7.x); `splitBackspaceBytes` is the only place chunk granularity is handled today.
- The repro needs no IME: queue two reads on a fake `Readable`, and the value ends up as the last chunk.

So this PR is a stopgap on the vendored copy until the fix lands upstream. If you'd rather wait for upstream than carry a patch in `vendor/ink/` (red zone), that's a fair call — and I'm happy to send it as a PR against ink 7.x instead.

shimpeiws:
Superseded by #4517, which fixes #4493 at the consumer side (PasteAwareTextInput effect guard + ink-text-input caret clamp) rather than buffering reads in vendored Ink `App.js`.

Verified working with Japanese IME on main `8994497c`. Closing this PR.

PR #4494, opened 2026-09-16
fix(slack): show activity for queued follow-ups [LET-12664]

4shub:
## Summary
- extend queued channel lifecycle events with the listener-owned `started` / `queued` submit disposition when known
- propagate that disposition through `ChannelGateway` and preserve it through registry adapter regrouping
- start native Slack thinking status for confirmed queued DM and established-thread follow-ups carrying inbound message IDs
- restore quiet-before-first-concrete-tool behavior for ordinary warm `started` inputs and early/legacy undefined-disposition DM or established-thread events
- preserve explicit `showStartupStatus` true/false overrides, active concrete-title refresh, historical flat-opener startup, and known-root retry suppression

## Root cause and review correction
Big Chungus correctly identified that lifecycle `type: "queued"` means accepted/observed input and fires for both immediately `started` and truly `queued` submissions. The earlier revision incorrectly treated the lifecycle type itself as proof of queue placement, which made ordinary warm turns noisy.

The listener response is authoritative. `ChannelGateway` now attaches `response.disposition` to the accepted queued lifecycle event without suppressing events for started inputs, and registry dispatch preserves that metadata while regrouping sources. Early ingress and legacy producers leave disposition undefined.

Slack only infers startup activity for a non-flat DM/channel source when `disposition === "queued"`. Undefined disposition retains historical flat-opener behavior only. Explicit startup flags win via nullish semantics, and no-message-ID sources remain quiet unless explicitly opted in.

## Tests
- [x] `bun test ./src/channels/gateway-core.test.ts` (31 passed)
- [x] `bun test ./src/channels/registry-controls.test.ts` (12 passed)
- [x] `bun test ./src/channels/slack/status-lifecycle.test.ts` (22 passed)
- [x] `bun test ./src/channels/slack/status-controller.test.ts` (14 passed)
- [x] `bun run check` (12/12 passed)
- [ ] Exact-built-commit pairing-DM visual probe after listener restart (startup status, warm started silence, concrete title preservation, pinning, message clear, reactions, cancellation, fatal error, and web footnote)

Focused coverage includes gateway `started`/`queued` propagation, registry preservation and early undefined ingress, warm started established-thread silence, started DM silence, confirmed queued established/DM startup, reply auto-clear recovery, explicit true/false, no-message-ID silence, active-title preservation, and root retry semantics.

## Evidence
- Before only: original `IMG_0533` reproduction is in the [reported Slack thread](https://letta-ai.slack.com/archives/C09CDK83A81/p1789576875838559). This is not after-proof.
- After proof is blocked on the exact-built-commit pairing-DM visual probe, so this PR remains draft.

## Links and context
- Linear: https://linear.app/letta/issue/LET-12664/show-slack-activity-for-queued-follow-ups-after-intermediate-replies
- Implementation conversation: https://chat.letta.com/chat/agent-ec93c28c-134a-4357-9388-4e7c6ed56bae?conversation=conv-f0c665c4-9ffb-4d0b-a3a4-d0d07ddd6775
- Repro Slack thread: https://letta-ai.slack.com/archives/C09CDK83A81/p1789576875838559
- Supersedes stale #3612 with equivalent behavior implemented against the current shared lifecycle path.
- LET-12407 timeout survival is related but distinct from this queued-follow-up visibility fix.

## Reconnect caveat
Root-shaped retry suppression uses the local adapter agent-thread tracker, which is session-local and is not retained across listener reconnects. Listener disposition and explicit override semantics remain authoritative; only remembered-root suppression has this reconnect limitation.

## Blocker
The required exact-built-commit pairing-DM visual probe could not be run in this environment. Keep this PR draft until that probe is completed and after-proof is attached.


4shub:
CI blocker from my review is resolved: run 35126576453 attempt 2 shows `API / ubuntu-24.04 / 3` passing and the full run green (verified via the attempts API — the aggregate "success" alone wasn't going to be enough, since attempt 1 of that job is still recorded as failed). Same code, clean attempt → flake, as reported.

Remaining blocker is the pairing-DM visual probe on the exact built commit. Draft status is correct until after-proof is attached.

PR #4507, opened 2026-09-17
fix(tools): require Monitor deadlines

amelia-letta:
## Summary

Claude Code 2.1.271 removed session-long Monitor watches: every command and WebSocket monitor now has a five-minute default deadline and a thirty-minute maximum. Letta Code still exposed `persistent: true`, which let monitors run until `TaskStop` or session exit, and accepted deadlines up to one hour.

This change removes the `persistent` tool input, caps `timeout_ms` at thirty minutes, and always starts command and WebSocket monitors with a deadline. `TaskStop` still cancels either source before its deadline. The existing background-process protocol field remains present as `false` for compatibility.

## Verification

- `bun test src/tools/impl/monitor.test.ts src/tools/impl/monitor-event-stream.test.ts src/cli/tui-interrupt-queue-lifecycle.test.tsx src/headless-interrupt-recovery.test.ts` (30 pass)
- `bun test src/tools/impl/stop-monitor.test.ts src/websocket/listener/control-inputs-interrupt.test.ts` (8 pass)
- `bun run check` (12/12 pass)

## Evidence

- Claude Code package: `@anthropic-ai/claude-code@2.1.274` (previous audited: `2.1.270`)
- Release: https://github.com/anthropics/claude-code/releases/tag/v2.1.274
- Monitor contract: https://code.claude.com/docs/en/tools-reference#monitor-tool
- Runtime inventory added or removed no advertised tools. The only `system/init` difference was the package version, and the Read line-prefix and Task metadata/deletion probes were unchanged.
- `SubagentHandback`, `omitClaudeMd`, and per-command `allowed_domains` are tied to Claude's auto-mode classifier and Claude agent-definition surfaces, not Letta Code's Agent tool contract. The remaining release and docs changes cover Claude UI, IDE, hosted cloud, gateway, MCP, plugin, and model-routing behavior rather than local harness mirrors.

## Breadcrumbs

- Requested in: #3926
- Letta conversation: [`conv-75187a75-4cad-4fb5-9654-758c1263ba39`](https://chat.letta.com/chat/agent-cd664b86-4d28-49b7-8ad6-60677eaff9be?conversation=conv-75187a75-4cad-4fb5-9654-758c1263ba39)
- Origin: Monitor's session-long `persistent` option was introduced with the initial event monitor in #3751.
- Workflow: https://github.com/letta-ai/letta-code/actions/runs/35174886217

## AI Disclosure

This draft was created by the automated Claude watcher and has not yet received human verification.

## AI Tool(s) Used

[...]

PR #4509, opened 2026-09-17
fix(worktree): neutralize repository checkout filters

amelia-letta:
## Summary

`EnterWorktree` creates a new checkout with `git worktree add`. Git runs several commands defined inside the repository's `.git` directory while it populates that checkout, and an agent can write those files, so a planted command ran outside the normal shell permission path.

`addWorktreeSafely` now creates the worktree with each of those paths switched off for that one command:

- **Checkout filters** (`filter.<name>.clean|smudge|process|required`) from repository-local config are overridden with command-scoped `-c` values, so files check out as their stored bytes.
- **Worktree-scoped filters.** With `extensions.worktreeConfig` on, Git copies `config.worktree` into the new worktree before checkout, and `git config --local` never reads that file. The scan covers the `--worktree` scope too when the extension is enabled. It checks the extension first, because `git config --worktree` dies in a multi-worktree repository without it.
- **Hooks.** `git worktree add` runs `post-checkout`. `core.hooksPath` points at a fresh empty temporary directory for the command. An empty directory is used instead of `/dev/null` because Git for Windows does not resolve that path.
- **`core.fsmonitor`.** A command value runs when the new index is refreshed, so it is set to `false` for the command.

Creation fails before Git runs when the configuration cannot be neutralized safely: a conditional include (`includeIf`), a repository-local Git LFS transfer program (`lfs.customtransfer.<name>.path`, `lfs.standalonetransferagent`), or a filter name containing `=` or a newline. Each dangerous key shape is defined once and shared by the `git config --get-regexp` scan and the classifiers, and a key the scan reports that no classifier recognizes also fails closed.

Tradeoffs: a repository whose LFS filter was installed with `git lfs install --local` gets pointer files in the new worktree, and a trusted global `post-checkout` hook does not run during creation. `EnterWorktree` provisions the worktree and wires the primary checkout's hooks afterwards, so later commits still run hooks.

Not covered here: nothing requires approval for file-tool writes under `.git/`, and auto-approved read-only commands such as `git status` also run a planted `core.fsmonitor` command. This change closes the `EnterWorktree` route only.

## Verification

- `bun test src/tools/impl/worktree-git.test.ts src/tools/enter-worktree.test.ts src/tools/exit-worktree.test.ts` (49 pass)
- CI at `d8ef3e8e`: 65/65 checks pass, including all eight Windows unit shards.

The tests use real Git repositories. A required smudge filter that exits unsuccessfully (repository-local and worktree-scoped) and a failing `post-checkout` hook each fail a plain `git worktree add`; the protected path creates the worktree and checks out the original bytes. An fsmonitor script that leaves a marker file must not run (skipped on Windows, it needs a POSIX script). A table of rejection cases covers both LFS settings, the worktree-scoped LFS setting, conditional includes, and unsafe filter names. One test adds a second worktree with the extension off, which is where an unconditional `--worktree` scan would die.

The worktree-scope, hook and fsmonitor tests were written first and failed against the filter-only implementation. Reverting those fixes locally fails the same four tests again. The Windows shard initially failed four exact-bytes assertions because runners enable `core.autocrlf`; the fixture now pins it off, reproduced locally by forcing `core.autocrlf=true`.

## Evidence

- Claude Code package: `@anthropic-ai/claude-code@2.1.274`
- Release: https://github.com/anthropics/claude-code/releases/tag/v2.1.274
- Worktree contract: https://code.claude.com/docs/en/worktrees
- The docs now explicitly refuse worktree creation when repository-local `lfs.customtransfer.<name>.path` or `lfs.standalonetransferagent` points Git LFS at a program. The same section documents neutralizing repository-local checkout filters and failing closed for conditional includes or filter names that cannot be overridden.
- The docs cover filters, conditional includes and the LFS settings only. They do not mention hooks, `core.fsmonitor`, or `config.worktree`; those three were found by probing `git worktree add` locally (git 2.39.3) and go beyond upstream parity.
- Runtime evidence was unchanged from the previous 2.1.274 snapshot: no advertised tool, help, init, event, auto-mode, or probe changes. The Read line-prefix and Task metadata/deletion probes still pass unchanged.
- The other focused docs changes cover Claude's plugin marketplace and MCP client selection, which have no local Letta Code mirror. The release's submodule-removal hardening applies to Claude's automatic agent-worktree cleanup; local `ExitWorktree` removes only after an explicit request, and Git already refuses a normal removal when the worktree contains a checked-out submodule.

## Breadcrumbs

- Requested in: #3926
- Letta conversation: [`conv-3733dba8-0ec9-4ba1-b5d8-2ab4eb541363`](https://chat.letta.com/chat/agent-cd664b86-4d28-49b7-8ad6-60677eaff9be?conversation=conv-3733dba8-0ec9-4ba1-b5d8-2ab4eb541363)
- Origin: worktree creation was introduced in #3046 and has always invoked `git worktree add` with repository-local config; no later regression introduced the gap.
- Related: #4507 handles the separate Monitor contract change from the preceding 2.1.274 candidate.
- Workflow: https://github.com/letta-ai/letta-code/actions/runs/35182120501

## AI Disclosure

This draft was created by the automated Claude watcher and has not yet received human verification.

## AI Tool(s) Used

Letta Code (Amelia) reviewed the exact release, official documentation snapshot, isolated runtime observations, local implementation, and tests, then authored this change.

Claude Code reviewed the change, found and fixed the worktree-scoped config, hook and fsmonitor gaps, refactored the key classification and tests, and fixed the Windows CI failure.

## Human Verification

Pending human review.

Claude-watch: claude@2.1.274:docs:5ad76f7151214be0bd9dd0050417ac36db452fd9955b03f81d40b517d7a90ee2:runtime:70723a9f57bd31c168c73a70b454a3df3c865f82420d22d486c5a1802e12730c


src/tools/impl/worktree-git.test.ts
```diff
@@ -1,14 +1,58 @@
...
+      GIT_COMMITTER_NAME: "Letta Test",
+      GIT_COMMITTER_EMAIL: "test@example.com",
+    },
+  }).trim();
+}
+
+async function createRepo(tempDirs: string[]): Promise<string> {
+  const repo = await mkdtemp(path.join(tmpdir(), "letta-worktree-safe-add-"));
+  tempDirs.push(repo);
+  git(["init", "-b", "main"], repo);
```
amelia-letta, line 32: `core.autocrlf` is enabled on Windows runners, so every exact `"content\n"` assertion below receives CRLF and this shard fails deterministically.

```suggestion
git(["init", "-b", "main"], repo);
git(["config", "core.autocrlf", "false"], repo);
```

devanshrj: Confirmed from the Windows shard log (whitespace-only diff on all four `"content\n"` assertions) and reproduced locally by forcing `core.autocrlf=true`. Applied in d8ef3e8e; the fixture now pins `core.autocrlf=false`.

Issue #4436, opened 2026-09-14
Windows: Bash tool should handle git stderr gracefully


jibramnobhill:
## Problem

On Windows, git commands emit informational messages to stderr (LF→CRLF warnings, branch switches, progress output). PowerShell treats any stderr from native commands as errors, which causes exit code 1 even when the git operation succeeded.

This forces agents to manually parse stdout vs stderr to determine actual success, wasting tokens and creating constant second-guessing.

## Established Solutions (from community)

1. `$env:GIT_REDIRECT_STDERR = '2>&1'` — redirect git's stderr to stdout at the source
2. Temporarily set `$ErrorActionPreference = 'Continue'` around git calls
3. Check `$LASTEXITCODE` instead of relying on exit code

## Proposed Improvement

The Bash tool on Windows could automatically handle this pattern for git commands, either:
- Suppress known benign git stderr patterns (LF→CRLF warnings, pre-commit hook warnings)
- Or document the expected pattern agents should use

## Evidence

Multiple projects have hit this:
- github/spec-kit#2253
- StackOverflow threads dating back to 2014
- posh-git documentation

This is a general PowerShell + git friction point, but Letta Code could make it easier for agents working on Windows.

PR #4511, opened 2026-09-17
fix(cli): warn Windows agents not to redirect native stderr in Bash tool commands

jibramnobhill:
## Summary

Refs #4436. Adds one Windows Shell Notes bullet so agents stop redirecting native stderr (`2>&1`, `2>$null`) in Bash tool commands on Windows, and asserts the bullet for every Windows shell family.

**Root cause (verified on Windows PowerShell 5.1.26100.9444, LettaJib):** the Bash tool runs `powershell -NoProfile -Command <cmd>` when `pwsh` is not installed and trusts PowerShell's own exit code. On Windows PowerShell 5.1 (and pwsh < 7.2, see PowerShell/PowerShell#13393 / #13395), redirecting a native command's stderr routes it through the error stream, flips `$?` to false, and `-Command` exits 1 while `$LASTEXITCODE` stays 0. Unredirected stderr is not affected: the tool already captures it separately and reports exit 0.

| Command (PS 5.1, `git add a.txt` emitting the LF?CRLF warning) | Exit |
|---|---|
| `git add a.txt` | 0 |
| `cmd /c "echo warn 1>&2 & exit 0"` | 0 |
| `git add a.txt 2>&1` | **1** |
| `git add a.txt 2>&1 \| Select-String x` / `\| Out-Null` | **1** |
| `git add a.txt 2>$null` | **1** |
| `$ErrorActionPreference='SilentlyContinue'; git add a.txt 2>&1` | **1** |
| `git add a.txt 2>&1; exit $LASTEXITCODE` | 0 |
| `$env:GIT_REDIRECT_STDERR='2>&1'; git add a.txt` | 0 (warning in stdout) |

## What this PR does not do

- No launcher/exit-code change. Trusting `$LASTEXITCODE` for Application final commands via the existing opt-in `POWERSHELL_EXIT_CODE_SUFFIX` fixes `git add 2>&1` but not `2>&1 | Select-String` (final command is a cmdlet) and doubles the command line (AST prefix + base64). Scoped separately.
- Does not change the `NativeCommandError` verbose formatting of redirected stderr in PS 5.1 (cosmetic, same root cause).

## Test plan

- [x] `bun test src/cli/helpers/session-context-windows.test.ts src/cli/helpers/session-context.test.ts` - 7 pass
- [x] biome, `tsc --noEmit`, source-file-size, pre-commit `bun run check` (via husky) green
- [ ] CI

Provenance: guidance text authored by the agent from the verified repro above; the repro table was produced with Node `spawnSync('powershell', ['-NoProfile','-Command', .])` against a temp repo with `core.autocrlf=true`.

## Breadcrumbs

- Origin: public issue #4436; the shell-specific example introduced in this PR was corrected in 0037fff5 following [review feedback](https://github.com/letta-ai/letta-code/pull/4511#discussion_r4041101149).


jibramnobhill:
@overlord-letta review requested on head `4a3fe2ad`. Scope: one Windows Shell Notes bullet in `buildWindowsShellNotes` (`src/cli/helpers/session-context.ts`) telling agents not to redirect native stderr (`2>&1`, `2>$null`) on Windows, plus a test asserting it for every Windows shell family. No launcher/exit-code change; the deterministic PS 5.1 repro table and the deferred launcher option are in the PR body. CI 59 pass / 0 fail / 2 skip on this head.

src/cli/helpers/session-context.ts
```diff
@@ -124,6 +124,7 @@ export function buildWindowsShellNotes(shellContext?: ShellContext): string {
 ${shellLine}
 - HEREDOC syntax (e.g., \`$(cat <<'EOF'...EOF)\`) does NOT work on Windows
 - For multiline strings (git commits, PR bodies), use simple quoted strings instead
+- Do NOT redirect native stderr (\`2>&1\`, \`2>$null\`) in Bash tool commands: on Windows PowerShell 5.1 that turns benign stderr (e.g. git's LF/CRLF warnings) into exit code 1. The Bash tool already captures stderr separately; for git, set \`$env:GIT_REDIRECT_STDERR="2>&1"\` instead
```
amelia-letta: This note is also emitted for cmd and bash, but `$env:...` is PowerShell-only and will fail if the agent copies it there.

jibramnobhill: Addressed in 0037fff5. The note now names the GIT_REDIRECT_STDERR environment variable and value without prescribing PowerShell-only assignment syntax, and explicitly scopes the redirection warning to Windows PowerShell 5.1. Coverage now includes Bash as well as PowerShell, cmd, and unknown shells; the updated test fails on the prior wording. Seven focused tests and all 12 repository checks pass; normal commit hooks also passed. No launcher or exit-code handling changed.

src/cli/helpers/session-context.ts
```diff
@@ -124,6 +124,7 @@ export function buildWindowsShellNotes(shellContext?: ShellContext): string {
 ${shellLine}
 - HEREDOC syntax (e.g., \`$(cat <<'EOF'...EOF)\`) does NOT work on Windows
 - For multiline strings (git commits, PR bodies), use simple quoted strings instead
+- Windows PowerShell 5.1: Do NOT redirect native stderr (\`2>&1\`, \`2>$null\`) in Bash tool commands; that turns benign stderr (e.g. git's LF/CRLF warnings) into exit code 1. The Bash tool already captures stderr separately; for git, set the \`GIT_REDIRECT_STDERR\` environment variable to \`2>&1\` using your shell's syntax instead
```
amelia-letta: `using your shell’s syntax` is misleading here: the Bash tool ignores the detected parent shell and independently prefers PowerShell (`windowsLaunchers`), so cmd/bash assignment syntax can be sent to PowerShell, while unquoted `2>&1` is parsed as redirection. Since this workaround only applies to Windows PowerShell 5.1, keep the exact PowerShell assignment here instead.

jibramnobhill: Verified: detectShellContext reads parent-environment hints, while buildShellLaunchers independently selects Windows launchers (pwsh/PowerShell before cmd). My previous wording conflated them. In 8351f24e I removed the optional GIT_REDIRECT_STDERR workaround altogether rather than restoring an example that could be copied into the wrong shell. The note now just says to leave native stderr unredirected because Bash already captures it separately. This preserves the PowerShell 5.1 warning without requiring an environment assignment. All four shell families are covered; the test fails on the prior wording, then 7 focused tests, all 12 repository checks and normal hooks pass. No launcher changes.

jibramnobhill:
@overlord-letta rereview requested on `8351f24e`. All 65 current-head checks passed, including Windows Unit/API shards and packaging. The two-file change is agent guidance only: keep the PowerShell 5.1 stderr warning, omit the optional Git environment workaround, and cover PowerShell/cmd/Bash/unknown families. Amelia's two threads are addressed and resolved. Seven focused tests (red-first), all 12 local checks, and normal hooks passed. No shell-launcher or exit-code behavior change.

PR #4512, opened 2026-09-17
test(memory-constraints): assert glob override path-segment semantics [LET-12756]

jibramnobhill:
## tl;dr

**Type:** Test/CI

**Goal:** `.memfs.config.json` `fileCharacterLimits` overrides are resolved by `validateMemoryTreeConstraints` (exported from `letta-code/memory-constraints`, also embedded in the MemFS pre-commit validator). Its glob translation is the policy-resolution rule: `*` and `?` stay inside one path segment, `**/` also matches zero directories, and regex metacharacters in a pattern are literal.

**What was missing:** The hook suites (`memory-git-v2-precommit.test.ts`, `memory-constraints.test.ts`, 63 tests) only use patterns where those rules make no difference. Mutating each rule survived the whole suite:

| Mutation of `validateMemoryTreeConstraints` | Existing 30 glob-relevant tests | With this PR |
|---|---|---|
| `*` ? `.*` (crosses `/`) | pass | fail |
| `(?:.*/)?` ? `.*/` (`**/` requires a directory) | pass | fail |
| no metacharacter escaping | pass | fail |
| `?` ? `.` (crosses `/`) | pass | fail |

A user-authored pattern such as `*.md` (root files only) would silently cap every nested file under the first mutation and nothing would notice.

**What changed:** One `test.each` in `src/memory-constraints.test.ts` drives the public export with a fake reader (3-character files, 1-character default cap, one uncapped override) and asserts exactly which paths stay capped for `*.md`, `reference/**/*.md`, `**/MEMORY.md`, `v?.md`, `a.b.md`. The fake `readFile` throws, so the case also asserts `countCharacters` is preferred when supplied. No production change.

## Review notes

**Merge when:** required checks are green on `91ca89a3`; the table above is reproducible by editing the four regex fragments in `src/memory-constraints.ts`.

**Start here:** `src/memory-constraints.test.ts` `cappedPaths` helper and the `test.each` rows; compare against the glob translator loop in `validateMemoryTreeConstraints`.

**Risk:** none to production. The expected-capped lists encode the documented "first match wins / null limit uncaps" semantics; a future intentional change to glob semantics must update the rows.

## Testing

**Automated (Windows 11, Bun 1.4.1, base `435e2cfb`):**

- `bun test src/memory-constraints.test.ts` (18 pass / 0 fail)
- `bun test src/memory-constraints.test.ts src/agent/memory-git-v2-precommit.test.ts src/agent/memory-constraints-audit.test.ts src/memory-frontmatter.test.ts` (63 pass / 0 fail before the change)
- Mutation table above: each of the four mutations flips the new case red and only it.

**Not tested:** Linux/macOS runs (CI); the embedded validator path in the pre-commit hook is covered by the existing hook suites and is unchanged.

## Breadcrumbs

- Linear: LET-12756
- Origin: glob translation added with the memory constraints policy export in #4140; no defect, coverage gap only.

Generating conversation: https://chat.letta.com/chat/agent-03c143ce-b161-4a7b-93e8-6a0d570cf5f4?conversation=conv-f9b37e97-a06b-43a9-80ac-5b84c4537e43


jibramnobhill:
@overlord-letta - requesting review at `91ca89a30f` (first current-head CI green, 60/60). Scope: test-only, `src/memory-constraints.test.ts` - one `test.each` on the public `validateMemoryTreeConstraints` export with a fake reader, asserting glob override path-segment semantics (`*` and `?` do not cross `/`, `**/` may match zero dirs, `.` is literal). Motivation: four mutations of the glob translator survived the existing 30 glob-relevant tests; each now fails this case. No production code.

PR #4513, opened 2026-09-17
test(mcp): assert camelCase schema words are searchable as separate terms [LET-12757]

jibramnobhill:
## tl;dr

**Type:** Test/CI

**Goal:** `letta mcp search` on a local agent ranks tools with `searchLocalMcpTools` (`src/cli/subcommands/mcp-search.ts`). `terms()` splits camelCase identifiers so a schema property like `variableName` is searchable by its trailing word `name`.

**What was missing:** The existing ranking tests (`mcp-search.test.ts`, 7 tests) only use queries that whole-token prefix matching also satisfies. Of four ranking mutations, three are caught (field-weight order, ?3-char prefix matching, local fused-score priority) but deleting the camelCase split survives the suite.

**What changed:** One `searchLocalMcpTools` case: query `name` against the existing fixture returns exactly `mcp__everything__print_env` (schema-only hit, score `0.25` = weight 1 / 4 terms). Without the split the token is `variablename`, no prefix relation holds in either direction, and the result is empty. No production change.

## Review notes

**Merge when:** required checks are green on `e09277f9`.

**Start here:** the new case next to `rejects vector mode .` in `src/cli/subcommands/mcp-search.test.ts`; compare with `terms()` in `mcp-search.ts`.

**Risk:** none to production. The asserted score depends on the fixture's schema having four terms; changing the fixture must update `0.25`.

## Testing

**Automated (Windows 11, Bun 1.4.1, base `435e2cfb`):**

- `bun test src/cli/subcommands/mcp-search.test.ts` (8 pass / 0 fail)
- All nine MCP discovery suites (`unified-mcp`, `mcp-search`, `mcp-tool-names`, `mcp-tool-listing`, `mcp-server-details`, `mcp-io`, `mcp-hosted-stdio`, `reminders/mcp-servers-info`, `mcp`): 60 pass / 0 fail before the change
- Mutation: removing `.replace(/([a-z0-9])([A-Z])/g, "$1 $2")` ? 7 pass / 1 fail (this case).

## Breadcrumbs

- Linear: LET-12757
- Origin: local search scoring introduced with `letta mcp search` in #4155; no defect, coverage gap only.

Generating conversation: https://chat.letta.com/chat/agent-03c143ce-b161-4a7b-93e8-6a0d570cf5f4?conversation=conv-f9b37e97-a06b-43a9-80ac-5b84c4537e43


jibramnobhill:
@overlord-letta - requesting review at `e09277f994` (current-head CI green; the single red was the live `Cloud teleport . active handoff` integration test on `API / ubuntu-24.04-arm / 2`, flaking on main and unrelated PRs in the same window, green on one failed-jobs rerun). Scope: test-only, `src/cli/subcommands/mcp-search.test.ts` - one `searchLocalMcpTools` case asserting camelCase schema words are searchable as separate terms (query `name` reaches `print_env` via `variableName`). Motivation: the camelCase term split was the one of four `mcp-search.ts` mutations the existing 60 tests did not catch. No production code.

PR #4517, opened 2026-09-17
fix(cli): keep multi-read input commits intact in PasteAwareTextInput

jibramnobhill:
## tl;dr

**Type:** Bug fix

**Requested in:** #4493 (Japanese IME commit drops characters / caret one short; reporter on macOS 26.5.2, iTerm2 3.6.11, CLI 0.32.11).

**Goal:** Text that reaches the composer as one logical commit must end up in the input intact, with the caret after the last character — regardless of how many stdin reads the terminal used to deliver it.

**What was missing:** An IME (Input Method Editor — the system component that turns romaji keystrokes into `日本語`) confirms a conversion as one commit, but the terminal may write it to the PTY in several pieces. `letta` reads each piece with its own `stdin.read()` inside the same `'readable'` tick, and vendored Ink emits one `input` event per read. Two or three `useInput` events therefore run **before React flushes the passive effects of the first render**. Two effects then applied an *older* value on top of newer state:

1. `PasteAwareTextInput` value-sync `useEffect([value])` re-issued `setDisplayValue(oldValue)` and clamped the caret to `oldValue.length`.
2. Vendored `ink-text-input`'s caret-clamp effect compared `previousState.cursorOffset` (already advanced by the later event) with its stale closure `originalValue` and moved the caret back to `oldValue.length`.

**What changed:**

- `src/cli/components/PasteAwareTextInput.tsx`: the value-sync effect returns early when the render already displays the incoming `value` (the echo of our own `onChange`); only parent-driven changes (clear, history navigation, restore) sync display/actual/caret. `actualValue` resolves placeholders in its initializer so the mount-time behavior is unchanged.
- `vendor/ink-text-input/build/index.js`: the caret-clamp effect clamps against the latest value the input handler reported (`latestValueRef`) instead of the effect closure's `originalValue`.
- `src/cli/components/PasteAwareTextInput.test.tsx` (new): renders the real component under vendored Ink with a fake raw-mode stdin whose `read()` yields N chunks in one tick, and asserts the parent-observed value, the reported caret, **and the caret actually drawn in the last frame** (inverse-video span).

Not changed: vendored Ink `App.js` still emits one `input` event per `stdin.read()`. Coalescing same-tick reads would also cover this but changes every input path (paste heuristics, key sequences); deferred deliberately.

## Before and after

Reads delivered within one `'readable'` tick (the reporter's `LETTA_DEBUG_INPUT=1` log pattern), reproduced deterministically without an IME:

| reads in one tick | `main` value | `main` caret | this PR value | this PR caret |
|---|---|---|---|---|
| `日本` + `語` | `日本語` | **2** | `日本語` | 3 |
| `日本` + `語の` + `入力中に変換がおかしくなるのが問題だ` | `日本入力中に…問題だ` (**`語の` dropped**) | **2** | all chunks, in order | 20 |
| `ab` + `c` | `abc` | **2** | `abc` | 3 |
| `日本` then `語` in separate ticks (control) | `日本語` | 3 | `日本語` | 3 |

**Before (two reads):** event 1 `日本` → handler caret 2, `onChange("日本")`, render 1 commits. Event 2 `語` runs before render 1's effects flush → handler caret 3, `onChange("日本語")`. React then flushes render 1's effects with `value="日本"`: composer clamps caret to `min(3, 2) = 2`; text input clamps `cursorOffset 3 > 2-1` → 2. The later effect for `日本語` sees `min(2, 3) = 2` and never corrects. User sees `日本|語`.

**After:** the composer effect for render 1 sees `value === displayValue` and returns; the text-input effect clamps against `latestValueRef = "日本語"` and leaves 3. User sees `日本語|`.

## Review notes

**Merge when:** the new test is green on all CI platforms and you agree the guard condition `value === displayValue` correctly identifies "own echo" — every `onChange` site in `PasteAwareTextInput` sets `displayValue` to the same string it emits, so a render where they differ is by construction a parent-driven change.

**Start here:** `PasteAwareTextInput.tsx` value-sync effect (guard + deps), then `vendor/ink-text-input/build/index.js` `latestValueRef`.

**Risk:** a parent that transforms the value it receives in `onChange` (emits `abc`, parent stores `ab`) still syncs, because that render has `displayValue !== value`. The other direction — parent setting the value back to exactly what is displayed — is a no-op both before and after. Paste-placeholder, newline-insert and history paths were run through the existing component tests (`bun test src/cli/components`, 102 pass).

## Testing

**Automated:**

- `bun test src/cli/components/PasteAwareTextInput.test.tsx` — 4 pass / 12 assertions (this head). On `main` (same test file): 3 fail / 1 pass, with the exact values in the table above.
- Mutation A — composer guard reverted, text-input guard kept: 3 fail (value drop + caret 2).
- Mutation B — text-input guard reverted, composer guard kept: 3 fail (drawn caret 2 in both two-read cases; three-read chunks reordered `日本入力中に…問題だ語の`). Each guard is independently necessary.
- `bun test src/cli/components` — 102 pass / 0 fail. `bun test src/cli/text-input-cursor-rendering.test.ts` — pass.
- `bun run check` — 12/12 (cycles, boundaries, biome, typecheck, …).
- Vendored change verified applied to `node_modules` by `scripts/postinstall-patches.js` (isolated `bun install --frozen-lockfile` in the worktree, byte-identical).

**Not tested:** a live Japanese IME under iTerm2/macOS or Windows Terminal/ConPTY (no IME automation available; a pipe-driven child is not a TTY, so raw mode cannot be exercised end to end). The stdin read pattern in the test is taken from the reporter's debug log; whether a given terminal splits commits this way is terminal behavior we do not control. This does not block merge: the defect is a React effect-ordering bug that reproduces on Windows with ASCII, and the fix is platform-neutral.

## Breadcrumbs

- Addresses #4493
- Origin: no single introducing PR found. Both effects predate visible history — present since the initial import (`0d66c756`, 2026-02-24); the text-input clamp is upstream `ink-text-input` 5.0.1 behavior carried into the vendored copy.
- Generating conversation: https://chat.letta.com/chat/agent-03c143ce-b161-4a7b-93e8-6a0d570cf5f4?conversation=conv-269d703c-eab4-42a7-a105-8a44e8d0adcc

## AI Disclosure

## AI Tool(s) Used

Letta Code (agent Hadron)

## Human Verification

Pending — to be completed by the human reviewer/owner before ready-for-review.


jibramnobhill:
@overlord-letta review requested on head `f3f7b3f7` (CI 59 success / 1 skipped on all four platforms).

Scope: two passive-effect guards for input commits split across several stdin reads in one tick (#4493) - the `PasteAwareTextInput` value-sync effect skips own-echo renders, and vendored `ink-text-input` clamps the caret against the latest handler-reported value instead of its stale closure. New `PasteAwareTextInput.test.tsx` drives the real component under Ink with a fake multi-read stdin and asserts value, reported caret, and the caret drawn in the frame; each guard reverted independently fails 3/4 cases (details in the body). Draft stays draft until your verdict.

Issue #2879, opened 2026-06-13
Retire the non-MemFS block path: decouple persona/human seeding from memory_blocks, drop --no-memfs


devanshrj:
## Context

The incremental memory-block teardown (#2868, #2870, #2873, #2874, #2875) has removed the genuinely dead block code: the legacy sleeptime agent, the LET-7353 skills-block migration, subagent block seeding, the `--init-blocks`/`--memory-blocks`/`--block-value` flags, `ISOLATED_BLOCK_LABELS`, and the global/project block-label split. Those were safe because each was already inert (empty arrays, no callers).

What remains can't be peeled off the same way — it's all live for any agent created or run with `--no-memfs`, and it needs design thinking before removal.

## The finding

The remaining "block" code is **not** a deletable parallel legacy system. Persona/human seeding is shared bootstrap that also initializes MemFS:

- `getDefaultMemoryBlocks()` (`src/agent/memory.ts`, ~50 LOC) reads the embedded `persona.mdx`/`human.mdx` → `CreateBlock[]`.
- `createAgent` puts those into the create request's `memory_blocks` field (`src/agent/create.ts:390`) for **every** agent, MemFS or not.
- The local backend renders those same blocks into the **initial files of the local memory git repo** (`src/backend/local/local-backend.ts:123` `initialMemoryFilesFromCreateBody` → `:474` `initializeLocalMemoryRepo`).
- `src/agent/personality.ts` overrides persona/human values per preset; `src/agent/defaults.ts` sets memo/incognito.

So persona/human content has to survive — it's where the MemFS files come from. This is a **refactor**, not a deletion.

## Proposed work (gated on making MemFS mandatory)

1. **Decide to drop `--no-memfs`** (product call). Everything below depends on it.
2. **Decouple seeding** — route persona/human (and personality presets) straight into MemFS file initialization instead of through the `memory_blocks` create field.
3. **Delete the now-unreachable non-MemFS surface:**
   - `letta_no_memfs.md` system-prompt variant (`src/agent/prompt-assets.ts:10`).
   - Server-side block-mutation tools (`memory`, `memory_apply_patch`, `memory_insert`, `memory_replace`, `memory_rethink` — `src/tools/toolset.ts:44`), currently detached for MemFS agents via `detachMemoryTools` but left attached for `--no-memfs`.
[...]

PR #4526, opened 2026-09-17
refactor(tools): remove legacy server-side memory tool handling

devanshrj:
## Summary

Remove Letta Code's handling of the legacy server-side memory tools (`memory`, `memory_apply_patch`, `memory_insert`, `memory_replace`, `memory_rethink`). These are Letta-server block-mutation tools from before MemFS; the repo never implemented them, it only knew their names in order to swap, detach, and label them. This is the "server-side block-mutation tools" item from #2879, unblocked by MemFS becoming mandatory in #3202. 13 files, +18 / −310 against `main`.

The two commits are deliberately separable because they carry different risk:

**1. `refactor(tools): remove dead server memory tool swap` — no behavior change for MemFS agents.**
`ensureCorrectMemoryTool` swapped a server-attached `memory_apply_patch` for server `memory` on every toolset/model switch (TUI `/model` and `/toolset`, listener `update_model`). Both of its extra parameters were already unused "compatibility" parameters, and it returned early unless the agent had one of those two server tools attached — which MemFS agents don't, so in practice it cost an `agents.retrieve(include: agent.tools)` round trip to find nothing. Removing it leaves `agentId` unused in `forceToolsetSwitch` / `switchToolsetForModel`, so that parameter is dropped and the call sites (TUI handlers and the new `toolset-construction.test.ts`) updated.

**2. `refactor(memfs): stop detaching legacy server memory tools` — this one is a real behavior change.**
`detachMemoryTools` is *not* dead code. It runs inside `applyMemfsFlags` whenever MemFS is being enabled for an agent, and both startup paths still auto-enable MemFS for existing Cloud agents that don't have it yet. So today, a pre-MemFS agent with server block tools attached gets them stripped the first time a current build opens it. After this commit such an agent keeps its server `memory` tool while also receiving the client-side MemFS `memory` tool by the same name. The runtime resolves that in favor of the client tool: `buildInvocationTools` (letta-cloud `libs/service-core-runtime/src/lib/core-runtime/tools/clientTools.ts:51-55`, pinned by the "still lets client tools shadow same-name loaded tools" test) drops any loaded server tool whose name matches a client tool, and client tool names are folded into the approval tool names (`steps/step.ts:3448-3455`) so the call is handed back to the client. In a Letta Code session the leftover server tool is therefore inert. What does change: with the `none` toolset (no client tools sent), or from non-Letta-Code clients such as the ADE/SDK, an un-migrated agent would now expose the server `memory` tool, whereas before it was stripped the first time Letta Code opened the agent. Agents created by Letta Code are unaffected (default base tools have long been just `web_search` + `fetch_webpage`). If that residual matters, drop this commit and land only the first.

Also in commit 2:
- `MEMORY_TOOL_NAMES` moves out of `tools/toolset.ts`. `memory` and `memory_apply_patch` are still live as bundled **client** tools, so `cli/helpers/tool-name-mapping.ts` now owns a two-name set for `isMemoryTool` / `getDisplayToolName`. Old transcripts containing `memory_insert` / `memory_replace` / `memory_rethink` calls will render with their raw tool name instead of "Memory"; `MemoryDiffRenderer` never had a rendering for those argument shapes anyway.
- The headless CI scenario stops requesting the server `memory` base tool, since nothing would detach it afterwards. `conversation_search` stays — the scenario prompt uses it.
- `toolset-memfs-detach.test.ts` and its `isolated-unit-tests.json` registration are deleted; the `syncing-memory-filesystem` skill and `applyMemfsFlags` doc comment no longer list the detach step.

Intentionally left alone: the `memory_apply_patch|memory` alternatives in `isToolsNotFoundError` (`agent/create.ts`). `--base-tools memory` is still a valid user flag, so base-tool bootstrap recovery for that name is not legacy-only.

`main` is merged in (`fa9e8010`): #4503/#4506 rewrote both switch functions around `loadTools(undefined, { resolvedToolset })`, so the resolution keeps main's structure with the memory-tool call removed.

## Verification

On head `fa9e8010` (after merging `main`), Bun 1.3.1 / Node for the built artifact:

- `bun run check`: all 12 checks pass on the merged tree (file-size baseline for `use-configuration-handlers.ts` ratcheted 1420 → 1418).
- `bun run build` (pre-merge head `3c6d50f3`): Node artifact builds; `node letta.js --version` → `0.32.13`.
- Focused suites on the merged tree, 179 passed / 0 failed: toolset construction (`toolset-construction`, new in #4503), MemFS enable (`memory-filesystem`, `personality-default-files`), TUI display and approval (`tool-name-mapping`, `format-args-display`, `approval-switch`, `memory-diff-renderer`), headless (`headless-ephemeral-toolset`, `headless-backend-lifecycle`), listener (`listen-model-update`, `commands/model-toolset`, `client-toolset-protocol`), and toolset (`client-toolset`, `toolset-caching`, `toolset-defaults`, `toolset-provider-resolution`, `toolset-client-tool-rule-cleanup`, `settings-manager-toolset`).
- Full `scripts/run-unit-tests.cjs` (pre-merge head `3c6d50f3`): 3148 passed, 30 failed, 2 errors. All failures are in 14 files that spawn the CLI and hit `Letta Code cannot run on Bun 1.3.1. Bun 1.3.2 or newer is required` (or use the 1.3.2 `beforeAll` signature). Re-running those 14 files on the untouched base commit `f8d2b757` gives identical per-file pass/fail counts, so they come from the local Bun version, not this branch. CI should be treated as the authority for those files.
- Not done: no live Cloud run against a legacy agent that still has a server `memory` tool attached. The collision behavior above is from reading the runtime source, not from exercising it.

## AI Disclosure

Human review, the policy acknowledgment, and the verification phrase below are pending while this PR is a draft.

## AI Tool(s) Used

- Claude Code: code investigation, implementation, automated verification, and PR preparation.


src/agent/memory-filesystem.ts
```diff
@@ -553,13 +552,7 @@ export async function applyMemfsFlags(
...
   const isEnabled = enabling || localMemfsEnabled;
 
-  // 3. Detach old API-based memory tools when enabling.
-  if (enabling) {
-    const { detachMemoryTools } = await import("@/tools/toolset");
-    await detachMemoryTools(agentId);
-  }
-
-  // 4. Add git tag + clone/pull repo.
+  // 3. Add git tag + clone/pull repo.
```
amelia-letta, line 555: This leaves pre-MemFS agents with both the legacy server `memory` tool and the MemFS client tool under the same name, so memory calls may hit stale block storage instead of MemFS.

devanshrj: Checked this against the runtime in letta-cloud, and the collision resolves in favor of the client tool:

- `libs/service-core-runtime/src/lib/core-runtime/tools/clientTools.ts:51-55` — `buildInvocationTools` drops every loaded server tool whose name matches a `client_tools` entry before the client tools are appended. Pinned by `clientTools.test.ts:77` ("still lets client tools shadow same-name loaded tools").
- `steps/step.ts:2161-2171` — the tool list sent to the model is filtered from `invocationTools`, so a legacy server `memory` tool never reaches the model while Letta Code is sending its MemFS `memory` client tool.
- `steps/step.ts:3448-3455` — client tool names are folded into the approval tool names, so a `memory` call is handed back to the client rather than executed server-side.

So in a Letta Code session the server block tool is inert cruft on an un-migrated agent, not a write path. Two narrower cases do change: the `none` toolset sends no client tools, so on such an agent the model would see the server `memory` tool; and non-Letta-Code clients (ADE/SDK) talking to that agent see it too. Before this PR both were cleared the first time Letta Code opened the agent. Noted in the description; if that residual matters, the second commit is the one to drop.

PR #4527, opened 2026-09-17
fix(reflection): preserve sandbox failure semantics

### LINKED ISSUE #4369: Reflection can falsely report no changes and consume history after sandbox temp-dir failure

cpacker:
## Summary

A sandboxed reflection child can write its memory worktree and transcript directory but not macOS's normal temporary directory. Background tools used `os.tmpdir()` anyway, so the first tool call could fail before execution. The child could then return explanatory text that the subagent manager accepted as success, causing automatic reflection to report no changes and advance past history it never inspected.

This change gives each memory subagent a private background-output directory under the writable transcript root. The parent removes that directory after the child exits. Reflection completion also requires at least one identified tool call to return `success`, so final text after zero successful tools remains failed and retryable.

An explicit `LETTA_SCRATCHPAD` remains unchanged. The headless auto-reflection integration fixture now performs a real deterministic `Bash` read of `TRANSCRIPT_PATH` before reporting success, so it exercises the same completion contract as production reflection.

Fixes #4369.

## Verification

- `bun test src/agent/subagent-env-composition.test.ts src/agent/subagents/subagent-stream.test.ts src/agent/subagents/sandbox.test.ts src/tools/impl/process-manager-output.test.ts` — 50 passed
- `bun test src/agent/subagents/subagent-stream.test.ts src/headless-bidirectional-reflection.test.ts` — 15 passed
- No live macOS Seatbelt run was available from this sandbox. The changed path composition and sandbox policy are covered by focused tests, and the macOS CI jobs run on each published head.

## Breadcrumbs

- Requested in: #4369
- Letta conversation: [`conv-e6fcae3b-f337-4066-8809-89f936a850cc`](https://chat.letta.com/chat/agent-19babcc0-e958-41b0-81d8-00107f71deb7?conversation=conv-e6fcae3b-f337-4066-8809-89f936a850cc)
- Origin: no single introducing pull request. #2812 moved background logs into process-private directories under `os.tmpdir()`, then #2853 restricted memory-subagent writes to Letta-owned paths.
- Replaces: #4424, rebased onto current `main` with per-child scratchpad isolation, cleanup, strict completion validation, and a transcript-reading integration fixture

## AI Disclosure

## AI Tool(s) Used

Letta Code

## Human Verification

Internal Letta agent-authored change. Human review requested.


src/agent/subagents/subagent-stream.ts
```diff
@@ -257,6 +285,24 @@ export function processStreamEvent(
   }
 }
 
+/**
+ * Whether every identified tool call reached an error terminal state.
+ * Incomplete streams remain unknown instead of being reclassified as failures.
+ */
+export function hasOnlyFailedToolCalls(state: ExecutionState): boolean {
+  if (state.displayedToolCalls.size === 0) return false;
```
amelia-letta: Allowing a zero-call run here makes a reflection that never opens `$TRANSCRIPT_PATH` count as successful, so its transcript is marked consumed without being inspected. The failing integration fixture uses `DeterministicPongExecutor`, which always returns `pong` without tools, so please fix that fixture instead of treating zero-call reflections as valid.

just-cameron (approved):

PR #4530, opened 2026-09-18
fix(headless): authorize remote startup as sender [LET-12850]

carenthomas:
## Summary

Remote `--computer` launches resolve or create their Cloud agent and conversation before execution moves to the selected computer. Those startup requests were authenticated only as the sandbox principal, so an existing target available to the sender could appear not found.

This change carries the inherited acting user through remote startup agent reads, conversation reads, and conversation creation. Direct CLI launches remain principal-only, and attributed agent reads bypass the agent-ID-only in-flight cache by using request-scoped options.

## Verification

- `bun test src/headless-backend-lifecycle.test.ts src/headless-startup-backend.test.ts` (9 passed)

## Risk and rollback

The acting-user header is added only when the launch is already routed through a remote computer and an acting user exists. Reverting this PR restores the prior principal-only startup behavior.

## Breadcrumbs

- Linear: `LET-12850`
- Related behavior: https://github.com/letta-ai/letta-code/pull/4334

## AI Disclosure

## AI Tool(s) Used

Letta Code

## Human Verification

Pending human review.


src/headless-startup-backend.ts
```diff
@@ -0,0 +1,51 @@
+import {
+  actingUserRequestOptions,
+  resolveActingUserId,
+} from "@/agent/acting-user";
+import type { Backend } from "@/backend";
+
+type HeadlessStartupBackend = Pick<
+  Backend,
+  "retrieveAgent" | "retrieveConversation" | "createConversation"
```
amelia-letta, line 9: `Agent({ computer })` without an existing agent takes the `--new-agent` path, but `createAgent()` still uses the global backend with no acting-user header. Fresh remote agents remain attributed to the sandbox principal.

amelia-letta:
@cpacker for nested `--computer` launches, is `LETTA_ACTING_USER_ID` the intended immutable sender identity for these pre-handoff agent/conversation reads and creation, rather than keeping them sandbox-principal-only until the listener receives `acting_user_id`?

PR #4531, opened 2026-09-18
feat(tools): Workflow tool as a thin wrapper around Agent SDK query()

sarahwooders:
## Summary

A `Workflow` tool that runs a plain-JavaScript orchestration script in the background. The script calls `agent(prompt)` to spawn subagents and uses `pipeline()` / `parallel()` to decide what runs concurrently. Each `agent()` call is one Agent SDK `query()` in an agent-free ephemeral conversation (`agent_id: null`, `parent_agent_id` = the invoking agent, `is_subagent: true`).

This is the minimal cut of #4179, which stays open as the reference for the features removed here.

| | #4179 | this PR |
| --- | --- | --- |
| Source lines (excl. tests) | ~4,000 | ~1,430 |
| Test lines | ~2,700 | ~1,000 |

## What is kept

- The DSL: `meta` literal, `agent`, `pipeline` (no barrier between stages), `parallel` (barrier), `phase`, `log`, `args`. Scripts run in a `node:vm` context.
- Agent-free, parent-linked worker conversations with read-only tools by default (`Read`, `Grep`, `Glob`).
- Per-subagent guards: 10-minute timeout (`opts.timeoutMs`), 60 tool calls, stop after 3 identical consecutive tool calls.
- Concurrency cap (`maxConcurrent`, default 16) and a 1000-agent lifetime cap.
- Model handling: inherit the invoking conversation's model, or pass a handle/alias (tool `model` or `opts.model`) resolved through the model catalog. Unknown handles fail at launch (tool input) or resolve that call to `null` (per call), with a pointer to `letta model list`.
- Background launch, `TaskOutput` progress log, `TaskStop`, and a completion `<task-notification>` through the same bridge as background Bash/Monitor.
- A per-run directory under `~/.letta/workflows/executions/<id>/` with `script.js`, `args.json`, and an append-only `journal.jsonl` (one line per completed agent).

## What is removed (vs #4179)

- **Resume / replay** and everything that existed to make it correct: cache keys, occurrence counters, `AsyncLocalStorage` branch paths, and the `Date.now()` / `Math.random()` blocks.
- **Remote computers** (`computer` selectors, per-computer client pool, `cwd` on remote).
- **Structured output**: no `StructuredOutput` tool, no hand-written JSON Schema validator, no retry. Replaced by `agent(prompt, { json: true })`, which strips a code fence and `JSON.parse`s the reply (~15 lines). Nothing validates the shape.
- **USD budget**, `workflow()` nesting, `sleep()`.
- **Worker thread** and its second `Bun.build` entry. Scripts run in-process; a script stuck in a CPU loop after an `await` cannot be stopped by `TaskStop`. Self-contained to add back if that bites.
- **Live UI**: execution registry, status panels, `/workflows`, `InputRich` / listener / protocol changes.

## Review notes

- `tools/manager.ts`: `Workflow` joins `STREAMING_SHELL_TOOLS` (so `parentScope` is injected) and a small `SCOPED_BACKGROUND_TOOLS` set replaces the inline `Monitor` check. Two doc comments were compacted so the file stays under its size baseline (lowered 2798 → 2795).
- Agent SDK dependency: `@letta-ai/letta-agent-sdk ^0.8.11` (lockfile resolves 0.8.12), loaded lazily by `sdk-loader.ts` as in #4179 because the SDK depends on a published `@letta-ai/letta-code`.
- Requires the API backend; the local store has no agent-free conversations.

## Testing

- `bun run check`: 12/12.
- `bun test src/tools/ src/cli/helpers/`: 1,057 pass (50 in the new engine/spawner/tool/meta tests).
- Verified `loadAgentSdk().createLocalClient()` loads the real SDK from the worktree under Bun.
- Not done: a live run against the API. The tool tests inject a fake spawner; the SDK query path is covered by unit tests with a fake client only.

## Follow-ups that would shrink this further

Adding `schema`, `timeoutMs`, `maxToolCalls`, and token usage to the SDK's `query()` would let `sdk-spawner.ts` drop to a few dozen lines and give every SDK user the same guards.


src/tools/workflow/workflow-engine.ts
```diff
@@ -0,0 +1,251 @@
...
+    }
+    currentPhase = title;
+    emit({ kind: "phase", title });
+  }
+
+  function log(message: unknown): void {
+    emit({ kind: "log", message: String(message) });
+  }
+
+  const context = vm.createContext({
```
amelia-letta, line 210: `agent` is a host-realm function, so workflow code can escape this VM with `agent.constructor("return process")()`; I confirmed it returns the host `process`. That bypasses the documented no-filesystem/network boundary and the subagent tool allowlist, while the approval display hides the script body.

amelia-letta:
@cpacker Does the SDK guarantee that a delayed `createLocalClient().query()` carrying only `parentAgentId` preserves the initiating listener user’s authority, or should Workflow explicitly capture and propagate `actingUserId` when the run is created?

src/tools/workflow/workflow-engine.ts
```diff
@@ -0,0 +1,254 @@
...
+    try {
+      if (signal.aborted) throw new Error("Workflow aborted.");
+      emit({ kind: "agent", callIndex, label, phase, status: "running" });
+      agentsSpawned++;
+      const outcome = await spawner(
+        { prompt, options: opts, callIndex },
+        signal,
+      );
+      if (signal.aborted) throw new Error("Workflow aborted.");
+      totalTokens += outcome.totalTokens ?? 0;
```
amelia-letta, line 117: `TaskStop` sets this signal before the spawner returns its accumulated usage, so the preceding throw skips this line and the terminal progress event. `/workflows` therefore reports zero or undercounted tokens for interrupted agents even though `sdk-spawner` preserved them.

src/tools/workflow/sdk-spawner.ts
```diff
@@ -0,0 +1,430 @@
...
+ * the real identity.
+ */
+function recordArgumentFragment(
+  call: TrackedToolCall,
+  input: unknown,
+  rawArguments: string | undefined,
+): void {
+  const fragment = argumentFragment(input, rawArguments);
+  if (fragment !== undefined) {
+    const parsed = parseArgumentObject(fragment);
```
amelia-letta: `rawArguments` values are deltas, so a fragment that happens to parse as `{}` is not the completed argument object. For `{"command":"run-0","options":` + `{}` + `}`, this resets every distinct call's identity to `{}` and the third call is stopped as an identical repeat.

src/tools/workflow/workflow-engine.ts
```diff
@@ -0,0 +1,260 @@
...
+    options.maxConcurrent ?? DEFAULT_MAX_CONCURRENT,
+  );
+
+  let currentPhase: string | null = null;
+  let callCounter = 0;
+  let agentsSpawned = 0;
+  let totalTokens = 0;
+
+  function agent(prompt: unknown, callOptions?: unknown): Promise<unknown> {
+    const pending = callAgent(prompt, callOptions);
```
amelia-letta, line 87: `agent()` promises are not tracked, so `agent("slow"); return "done"` completes the workflow and disposes the SDK client while that subagent is still running. The completion notification and journal can therefore claim success before every launched worker has finished.

src/tools/workflow/journal.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/** Create the run directory and persist the script (and args) it will run. */
+export function createExecutionDir(
+  executionsDir: string,
+  executionId: string,
+  script: string,
+  args: unknown,
+): { executionDir: string; scriptPath: string; journalPath: string } {
+  const executionDir = join(executionsDir, executionId);
+  mkdirSync(executionDir, { recursive: true });
```
amelia-letta: These directories and files use the process umask, so with a normal `022` they are `0755`/`0644` even though they persist full scripts, args, prompts, and responses. On a shared machine, other local users can read workflow data.

src/tools/workflow/workflow-engine.ts
sarahwooders: Agreed, and the boundary claim was wrong: node:vm is a scope, not a sandbox, and any host value that crosses into the context (the hooks, the promises they return, `args`) leaks its realm. Wrapping every value is a treadmill, so 3d6512db does not try. Instead: the docs (tool description, authoring skill, engine header) now say the script runs with the CLI's own privileges, like Bash, and that reading/searching/writing belongs in subagents where the allowlist applies; and the approval prompt now prints a `script` argument verbatim (up to 40 lines, with a remaining-line count) instead of a JSON-escaped string cut at 500 chars, so the user approves the source that actually runs.

src/tools/workflow/workflow-engine.ts
sarahwooders: Fixed in 3d6512db. The abort check moved after accounting: when the spawner returns for an interrupted agent, its tokens are added to the run total, the journal line is written, and the terminal progress event (with `totalTokens`) is emitted before the call throws. `emit` is now gated on run settlement rather than `signal.aborted`, and the run does not settle until every in-flight agent has reported, so `/workflows` and the registry see the interrupted agent's usage. Regression test: "an interrupted subagent still reports its outcome and tokens".

src/tools/workflow/sdk-spawner.ts
sarahwooders: Fixed by #4582 (merged into this branch as 915ff3a3): raw deltas are always concatenated and never treated as a standalone snapshot, with a regression replay for `{"command":"run-0","options":` + `{}` + `}`.

src/tools/workflow/workflow-engine.ts
sarahwooders: Fixed in 3d6512db. Every `agent()` promise is tracked in an in-flight set, and `executeWorkflow` drains that set before returning or rethrowing, so `agent('slow'); return 'done'` waits for the worker (its outcome is journaled and counted) before the completion notification is queued and the SDK client is disposed. Regression test: "waits for un-awaited agent() calls before settling".

src/tools/workflow/sdk-spawner.ts
```diff
@@ -0,0 +1,408 @@
...
+ */
+function recordArgumentFragment(
+  call: TrackedToolCall,
+  input: unknown,
+  rawArguments: string | undefined,
+): void {
+  const fragment = argumentFragment(input, rawArguments);
+  if (fragment !== undefined) {
+    call.raw = call.raw === undefined ? fragment : call.raw + fragment;
+    call.complete = undefined;
```
amelia-letta, line 127: This leaves the completed JSON as raw bytes, so semantically identical calls with different whitespace or escapes get different identities and can alternate until the 1,000-call cap or timeout without tripping the three-call guard. Parse and canonicalize `call.raw` in `onResult`, after all deltas have been accumulated.

src/tools/workflow/journal.ts
sarahwooders: Fixed in 3d6512db: the run directory is created with mode 0700 and script.js, args.json, and journal.jsonl with 0600, regardless of umask. Test: "creates run files owner-only regardless of umask" (skipped on win32).

sarahwooders:
On the `actingUserId` question: today the Workflow tool captures `actingUserId` at launch only for the completion notification's scope. The SDK query is created later with `parentAgentId` and nothing about the acting user; `createLocalClient()` spawns its own app-server under the CLI's credentials, so the worker conversation is created with the CLI's authority, not the initiating listener user's. `AgentFreeQueryOptions` has no field to carry it, so propagating would need an SDK change first. Leaving this open for @cpacker; if the answer is that it should propagate, the tool-side change is small once the SDK accepts it.

src/cli/components/InlineGenericApproval.tsx
```diff
@@ -23,12 +23,37 @@ type Props = {
...
+function formatScriptArgs(parsed: Record<string, unknown>): string {
+  const { script, ...rest } = parsed;
+  const lines = String(script).split("\n");
+  const shown = lines.slice(0, SCRIPT_PREVIEW_MAX_LINES);
+  if (lines.length > shown.length) {
+    shown.push(`… (${lines.length - shown.length} more lines)`);
+  }
+  const header =
+    Object.keys(rest).length > 0 ? `${JSON.stringify(rest, null, 2)}\n\n` : "";
+  return `${header}${shown.join("\n")}`;
```
amelia-letta, line 41: Rendering the decoded script verbatim also emits any ANSI/control sequences in it. Since this preview is the approval boundary for host-privileged code, a script can clear or rewrite the terminal and spoof what the user approves; strip or visibly escape control sequences before rendering.

src/cli/components/InlineGenericApproval.tsx
```diff
@@ -23,12 +23,37 @@ type Props = {
 // Horizontal line character for Claude Code style
 const SOLID_LINE = "─";
 
+const SCRIPT_PREVIEW_MAX_LINES = 40;
+
+/**
+ * A `script` argument is source the user is approving to run; show it
```
amelia-letta, line 29: Only the first 40 lines are shown even though the remainder executes with CLI privileges, so host-side behavior after line 40 is hidden from the user approving it. The remaining-line count does not preserve this as an approval boundary.

src/tools/impl/workflow.ts
```diff
@@ -0,0 +1,448 @@
...
+    "If the result above is empty or unexpected, read that file BEFORE diagnosing — do not assume agents returned non-empty results.",
+  ].join("\n");
+}
+
+export async function workflow(args: WorkflowArgs): Promise<WorkflowResult> {
+  let script = typeof args.script === "string" ? args.script : undefined;
+  if (typeof args.scriptPath === "string" && args.scriptPath) {
+    try {
+      script = readFileSync(
+        resolve(getCurrentWorkingDirectory(), args.scriptPath),
```
amelia-letta, line 258: `scriptPath` bypasses the source preview entirely: approval shows only the filename, then this reads and executes unseen contents with CLI privileges. Preview the exact bytes that will run or do not treat this path as source approval.

src/tools/impl/workflow.ts
```diff
@@ -0,0 +1,448 @@
...
+
+/**
+ * Scripts can return anything (BigInt, circular graphs, functions); the
+ * notification must never throw over it.
+ */
+function jsonSafe(value: unknown): unknown {
+  try {
+    return JSON.parse(JSON.stringify(value) ?? "null");
+  } catch {
+    return String(value);
```
amelia-letta, line 228: `String(value)` can also throw, for example for a cyclic null-prototype object containing a BigInt, so `finish()` exits before terminal bookkeeping and the task stays `running` forever. Make this fallback non-throwing so every run reaches a terminal state.

PR #4537, opened 2026-09-18
fix(tools): align overflow output with Claude

amelia-letta:
## Summary

Letta Code saves oversized model-facing output to files so the model can read the rest when needed. Successful Bash results currently keep a 30,000-character middle excerpt, failed Bash results use the same file-backed path, and hook feedback has no equivalent cap.

Claude Code's documented contract is more selective: successful Bash overflow and hook strings keep up to the first 2,000 characters plus the saved file path, each hook string is measured independently against a 10,000-character limit, and failed Bash output returns a 10,000-character head-and-tail excerpt without a file path. This change applies those boundaries to the corresponding Letta Code paths, with one deliberate departure: failed Bash overflow keeps the saved file path.

Claude Code's documented failure result is the excerpt alone. Letta Code holds the full output and deletes the foreground output file before returning, so without an overflow file the rest of a large failing build or test log would be unreachable, which is exactly when the model needs it. The path costs about 100 characters.

The hook cap covers every string a hook can put in front of the model, not only the ones Claude Code's documentation lists (plain stdout and `additionalContext`): block and error stderr are capped too. A `UserPromptSubmit` block goes straight into a system reminder with nothing clamping it downstream, so a noisy failing hook could otherwise flood the context. SessionStart output is capped in `runSessionStartHooks`, which rebuilds its feedback from raw stdout rather than using the executor's result.

The 2,000-character preview only applies when the overflow file was actually written. With `LETTA_TOOL_OVERFLOW_TO_FILE=false` or a failed write, output falls back to the normal excerpt at the full limit instead of a short prefix with no path to the remainder.

## Before and after

```text
Before

successful Bash output over 30,000 characters
  -> 30,000-character middle excerpt + saved file path

failed Bash output over 30,000 characters
  -> 30,000-character middle excerpt + saved file path

model-facing hook feedback
  -> full string inline

After

successful Bash output over 30,000 characters
  -> first 2,000 characters + saved file path

failed Bash output over 10,000 characters
  -> 10,000-character head-and-tail excerpt + saved file path

each model-facing hook string over 10,000 characters
(stdout, additionalContext, SessionStart output, block and error stderr)
  -> first 2,000 characters + saved file path

any of the above when no overflow file could be written
  -> normal excerpt at the full limit, no saved file path
```

## Out of scope

`shell_command` and `TaskOutput` keep the existing 30,000-character middle excerpt plus saved file path. `shell_command` is not a Claude-shaped tool, and the cited output-limit documentation does not cover `TaskOutput`, so both are intentionally left out of this parity change.

## Verification

- `bun test src/tools/truncation.test.ts src/tools/truncation-overflow.test.ts src/tools/overflow-integration.test.ts src/tools/tool-truncation.test.ts src/tools/bash.test.ts src/hooks/executor.test.ts src/hooks/integration.test.ts src/hooks/session-start-feedback.test.ts` (139 pass)

## Evidence

- Claude Code package: `@anthropic-ai/claude-code@2.1.276`, unchanged from the previous audited snapshot
- Release: https://github.com/anthropics/claude-code/releases/tag/v2.1.276
- Hook output limits: https://code.claude.com/docs/en/hooks#json-output
- Bash output limits: https://code.claude.com/docs/en/tools-reference#output-limits
- The release's `advisor_20260301` proxy fix has no local mirror because Letta Code does not depend on the Claude Code package or emit that request tag.
- The exact runtime digest is unchanged. Advertised tools, help, doctor output, `system/init`, auto-mode defaults, event types, and capabilities did not change. The Read line-prefix probe and Task metadata and deletion probe still pass with the same asserted behavior.

## Breadcrumbs

- Requested in: #3926
- Letta conversation: [`conv-89673cbb-132a-4463-b831-4086fc6e9a74`](https://chat.letta.com/chat/agent-cd664b86-4d28-49b7-8ad6-60677eaff9be?conversation=conv-89673cbb-132a-4463-b831-4086fc6e9a74)
- Origin: Bash truncation was introduced in [`4118d018`](https://github.com/letta-ai/letta-code/commit/4118d018), and Claude-compatible hooks were introduced in #607 with `additionalContext` support added in #826.
- Workflow: https://github.com/letta-ai/letta-code/actions/runs/35368357020

## AI Disclosure

This draft was created by the automated Claude watcher and has not yet received human verification.

## AI Tool(s) Used

Letta Code (Amelia) reviewed the exact release, official documentation snapshot, isolated runtime observations, local implementation, and tests, then authored this change.

Follow-up commits were written with Claude Code (`db754a81` rewords the new Bash description so it does not name Grep, which no toolset preset includes; `22e4cdf9` caps SessionStart output and hook stderr in response to review, and removes test cleanup made unnecessary by the disposable test home; `39e937a5` stops the executor from saving SessionStart feedback that `runSessionStartHooks` discards): `57443c7c` (saved file on failed Bash overflow, preview gated on the overflow file, shared test helper) and `02c70462` (Bash tool description now explains the saved overflow file, a code comment marking the failure-path file as a deliberate departure, and overflow-directory cleanup in the tool-truncation tests).

Claude-watch: claude@2.1.276:docs:fd41239dcc978fb0971dc9f8d10746e34a82b46a4cac178c4eb8897a0628f9bb:runtime:c3298b053328e6a1164b7c620af9937e4d81ed201bf7e18442ee5bd8b2e39d34


src/hooks/executor.ts
```diff
@@ -19,6 +20,13 @@ import {
 /** Default timeout for hook execution (60 seconds) */
 const DEFAULT_TIMEOUT_MS = 60000;
 
+function truncateHookFeedback(text: string, workingDirectory: string): string {
```
amelia-letta: Blocking and error stderr still bypass this helper and enter feedback unbounded below. A noisy failing hook can still flood the model context despite the new 10k contract.

src/hooks/executor.ts
```diff
@@ -387,7 +395,9 @@ export async function executeHooks(
           input.event_type === "UserPromptSubmit" ||
           input.event_type === "SessionStart"
         ) {
-          feedback.push(result.stdout.trim());
+          feedback.push(
```
amelia-letta: `runSessionStartHooks` discards this truncated feedback and rebuilds it from raw `result.results` stdout, so oversized SessionStart output still enters the first model message unbounded.

src/hooks/executor.ts
```diff
@@ -19,6 +20,13 @@ import {
 /** Default timeout for hook execution (60 seconds) */
 const DEFAULT_TIMEOUT_MS = 60000;
 
+function truncateHookFeedback(text: string, workingDirectory: string): string {
+  return truncateByChars(text, LIMITS.HOOK_OUTPUT_CHARS, "Hook", {
```
amelia-letta, line 31: This file-backed path receives no secret map, but hook subprocesses inherit the parent environment. A hook that emits an inherited credential now persists the raw value in the overflow file, unlike the shell overflow paths that scrub before writing.

src/hooks/executor.ts
devanshrj: Fixed in 22e4cdf9. The cap is now applied once where each executor assembles its result, rather than at individual push sites, so block stderr and `Hook error:` strings are covered along with stdout and `additionalContext`. This matters most for `UserPromptSubmit` blocks, which go straight into a system reminder with nothing clamping them downstream. Added a test for oversized stderr from a blocking hook.

src/hooks/executor.ts
devanshrj: Good catch, this was a real bug. Fixed in 22e4cdf9: `runSessionStartHooks` now caps the stdout it rebuilds. I also stopped `executeHooks` from collecting SessionStart stdout, since its only caller discards it — otherwise oversized output would be capped twice and leave a second, orphaned overflow file. The new test goes through `runSessionStartHooks` and asserts exactly one overflow file is written. The earlier test called `executeHooks` directly, which is why it passed while the production path was uncapped.

src/hooks/executor.ts
devanshrj: Not changing this one; the comparison with the shell paths doesn't hold. The Bash overflow scrub only covers `invocationSecrets` — agent secrets injected for that specific tool call (`manager.ts` passes them as `secretEnv`) — not credentials inherited from the parent environment. A Bash command that prints an inherited credential persists it to its overflow file in exactly the same way. Hooks are spawned with plain `process.env` and receive no injected secrets, so there is no secret map to pass here. The same string also already reaches the model inline, unscrubbed. Scrubbing inherited-environment credentials would be a separate change that needs to cover every overflow writer, not just hooks.

src/hooks/executor.ts
```diff
@@ -417,7 +430,9 @@ export async function executeHooks(
   return {
     blocked,
     errored,
-    feedback,
+    feedback: feedback.map((text) =>
```
amelia-letta: This also truncates SessionStart block/error stderr, but `runSessionStartHooks` discards `result.feedback`; a noisy failing SessionStart hook now writes an orphaned overflow file whose path is never returned on every session. Skip capping discarded SessionStart feedback, not just its stdout.

devanshrj: Right, I exempted SessionStart stdout from the executor's cap but missed the stderr half. Fixed in 39e937a5: `executeHooks` now skips the cap entirely for SessionStart, since `runSessionStartHooks` discards that feedback and caps the stdout it rebuilds. Added a test through `runSessionStartHooks` with a failing hook that writes oversized stderr, asserting no overflow directory is created.

PR #4520, opened 2026-09-17
fix(memfs): repair conflicts in hidden conversations

devanshrj:
## Summary

When post-turn MemFS sync encounters a Git conflict, start a hidden conversation with the same agent to resolve it. Keep the repair prompt and completion messages out of the active conversation.

Use the existing background runner with a memory-scoped toolset, a 20-turn limit, and no startup sync, reflection, or recursive repair. Coordinate repairs across processes, verify Git state before retrying the push, and back off five minutes when a repair remains unresolved. One-shot headless runs wait for their repair before exiting.

Attached shared-memory repositories retain their existing reminder behavior.

## Verification

- `bun run check` — all 12 checks passed.
- `bun run build` — passed.
- 163 focused tests passed across reminders, memory Git, repair coordination, subagents, listener cleanup, and headless repair; the isolated listener shared-memory sync test also passed.
- `LETTA_TEST_BUILT_CLI=<worktree>/letta.js bun test src/headless-memory-repair.test.ts` — all three tests passed against the built Node CLI. The same scenarios passed under Bun 1.3.2.

Tests exercise real Git conflicts and a deterministic backend, including an actual primary-to-hidden-conversation launch. No live-model conflict-resolution run was performed.

## AI Disclosure

AI-assisted implementation. Human review and policy attestation are pending while this PR is a draft.

## AI Tool(s) Used

OpenAI Codex researched the implementation, wrote the changes and tests, and ran validation.

## Human Verification

Pending human review before marking ready.

PR #4539, opened 2026-09-18
feat(memory): delegate edits and conflict repair to background subagents

devanshrj:
## Summary

Memory edits and Git conflicts can interrupt the primary agent. Delegate both to the existing `memory` subagent, rewritten to run silently in the background as a fresh agent. The primary supplies a self-contained handoff; the worker receives the exact checkout path and read-only transcript access only as a fallback. Completion and failure stay in the background task log.

Automatic MemFS Git-conflict repair uses the same worker, covering the repair use case from #4520. Primary memory tools are removed from default toolsets and prompts delegate memory writes. Reflection remains separate and shares only checkout coordination. One-shot headless emits the primary result before draining background memory work.

Fresh workers remain stateless, so shell tools retain the parent checkout instead of selecting a nonexistent child memory directory. Transcript filenames are unique across process restarts. Worker identity is logged at startup, its report before sync, and failed sync is recorded as failure. Primary handoffs quote factual corrections and exceptions to avoid broadening their scope.

## Verification

- `bun run check`: all 12 checks passed; build passed.
- Focused memory/background/shell-environment suite: 119 passed across 11 files. Final follow-up suite: 27 passed across four files. Counts overlap.
- Source headless launch/sender/diagnostics: six passed across three files. Built-Node headless memory tests: three passed, including actual fresh-child launch with a deterministic local backend.
- Interactive reflection worktree completion and listener shared-memory cleanup: 13 passed across two files.
- Regression coverage includes distinct read-only snapshots for repeated launches, no inline history, no history export for Git repair, immediate silent delegation, startup identity logging, terminal export failures, and failed sync preserving the worker result and releasing the checkout.

Bun 1.3.14; built CLI on Node 25.1.0.

## Live validation

The final build passed **15/15 actual primary-to-worker trials**: simple edit plus unrelated question, coding plus memory, multi-turn review preferences, corrected project facts with a local-test exception, and real Git merge conflicts; three repetitions each. The primary model wrote every edit handoff and called the real Agent tool. Repair assignments came from the normal harness hook. No evaluator-written handoffs or inline transcript replay.

All 15 saved the required facts, completed background work, committed cleanly and synced. All three primary coding fixtures passed their tests. Across the 15 trials: **zero transcript reads, zero unrelated worker answers or coding work, and zero drain timeouts**. Manual audit checked the actual handoffs and files, including narrow exceptions and stale-fact replacement.

A separate real process restart resumed the same primary conversation and transcript root: the new worker corrected Bun to pnpm, created a distinct snapshot, preserved the previous read-only snapshot, committed and synced. No filename collision or direct primary memory edit.

The first fix-validation round completed/synced 15/15 but preserved facts accurately in 14/15: one primary broadened “local tests” to “testing/development.” That failure prompted the verbatim-correction instruction and the separate complete final rerun above. Earlier worker-only experiments did not test automatic handoff generation and are not used as production-path evidence here.

Cloud / Claude Sonnet 4.6 / macOS sandbox / headless / reflection off / unrestricted permissions, with four independent trials concurrent in the final matrix. These are short histories, not exhaustive model or UI validation. Git transport retry policy is unchanged; earlier remote stalls were not conclusively diagnosed. Failures now remain inspectable instead of being mislabeled as successes.

Full local artifacts, exact prompts and per-trial evidence: `/tmp/memory-handoff-final-4539/REPORT.md`; earlier fix round and restart: `/tmp/memory-handoff-fixed-4539/`.

## AI Disclosure

AI-assisted implementation. Human review and policy attestation are pending while this PR is a draft.

## AI Tool(s) Used

OpenAI Codex — implementation, tests, validation, and draft PR preparation.

## Human Verification

Pending human review before marking ready for review.


src/tools/impl/task.ts
```diff
@@ -890,14 +890,18 @@ export async function task(args: TaskArgs): Promise<string> {
     existingAgentId: effectiveAgentId,
     existingConversationId: effectiveConversationId,
     maxTurns: args.max_turns,
-    forkedContext: config.fork,
+    forkedContext: subagent_type !== "memory" && config.fork,
     parentScope: resolvedParentScope,
     environment:
```
amelia-letta, line 927: `computer` still routes memory workers to another machine, but `runMemoryWorker` locks/syncs the checkout on this machine and passes that local path to the child. Reject routed memory tasks (or move the whole lifecycle to the target), otherwise they can fail or edit a different checkout.

src/headless.ts
```diff
@@ -3276,6 +3268,7 @@ export async function handleHeadlessCommand(
   // Report all milestones at the end for latency audit
   markMilestone("HEADLESS_COMPLETE");
   reportAllMilestones();
+  await finishBackgroundMemoryTasks(agent.id, conversationId);
```
amelia-letta: This drain only covers successful one-shot completion. Earlier `exitHeadless` paths and `exitBidirectional` can leave a detached memory child editing after the parent PID that owns the checkout lock has exited, so a new process may reclaim the lock and mutate concurrently; drain or cancel memory tasks from both shutdown helpers.

src/agent/subagents/memory-handoff.ts
```diff
@@ -0,0 +1,71 @@
...
+    }
+    const directory = join(
+      getTranscriptRoot(),
+      params.agentId,
+      "memory-handoffs",
+    );
+    await mkdir(directory, { recursive: true, mode: 0o700 });
+    // Task counters restart with the process; never overwrite an earlier read-only snapshot.
+    transcriptPath = join(directory, `${randomUUID()}.json`);
+    await writeFile(transcriptPath, JSON.stringify(messages, null, 2), {
```
amelia-letta, line 69: Each memory task now writes the full paginated conversation to a UUID file, and nothing removes these snapshots. Repeated updates create unbounded, potentially quadratic growth under `~/.letta/transcripts`; clean up after the worker finishes or add bounded retention.

amelia-letta:
@sarahwooders Is it intentional for the primary to act on identity or behavior changes immediately even though the fresh memory worker may fail silently, or should those changes remain gated on a successful memory write?

src/agent/prompts/letta_root_memfs.md
```diff
@@ -55,33 +55,31 @@ The MemFS is a git-backed projection of your memory. Changes affect your future
...
+When delegating, provide a self-contained description of what to remember, correct, delete, or reorganize, including relevant facts, corrections, and exceptions. Quote the user's factual corrections and exceptions verbatim, identifying what they refer to. Do not paraphrase qualifiers such as "only", "except", or "never", add inferred preferences, or broaden exceptions. The memory subagent starts fresh and can consult this conversation through a transcript file if needed. Delegation means the update is in progress, not already saved.
 
-`$AGENT_NAME` is normally populated when the runtime knows the current agent name, but direct shell environments can still miss it. Use a non-empty author name fallback when committing directly.
+### Editing memory directly
+
+When memory is the main task, inspect Git status and read the current files before editing. If a memory worker you launched is still editing the same checkout, use TaskOutput to wait for it to finish before making direct edits, then reread the files. This wait is appropriate only because memory is now the main task.
```
amelia-letta: Direct primary edits here bypass `withMemoryOperation`; waiting only for a worker this conversation launched does not cover automatic repair, reflection, or a worker from another conversation/process using the same checkout. Those can mutate the index concurrently, so primary memory work needs to participate in the checkout lock rather than relying on prompt coordination.

src/tools/memory-tool-operation.ts
```diff
@@ -0,0 +1,114 @@
...
+    const root = ordered[index];
+    return root
+      ? withMemoryOperation(
+          root,
+          () => execute(index + 1),
+          args.signal instanceof AbortSignal ? args.signal : undefined,
+        )
+      : run(args);
+  };
+  return execute(0);
```
amelia-letta: Releasing the lease after each tool call leaves the common `Write`/`Edit` then `git add && git commit` flow unlocked between the edit and commit. A queued worker can acquire the dirty checkout and either fail or absorb the primary's changes, so direct updates need ownership across the whole edit/commit transaction.

src/agent/memory-operation.ts
```diff
@@ -0,0 +1,183 @@
...
+    { wait: true, signal },
+    true,
+  );
+  if (!acquired) throw new Error("Failed to acquire memory checkout");
+  return currentOperation.run(acquired.lease, async () => {
+    try {
+      signal?.throwIfAborted();
+      return await operation();
+    } finally {
+      await Promise.all(acquired.lease.processes);
```
amelia-letta: Awaiting every tracked shell completion means a memory-scoped `exec_command`/`Bash` that yields or uses `run_in_background` no longer returns its session until the process exits. A long-running process can hang the tool call indefinitely while holding the checkout lock.

src/tools/impl/shell-runner.ts
```diff
@@ -510,6 +511,7 @@ export function startShellProcess(
     abortHandler();
   }
 
+  trackMemoryOperationProcess(completion, terminateProcess);
```
amelia-letta: Shell env is built from `process.env` and never receives `getMemoryOperationEnv()` for the newly acquired lease. If this shell invokes a memory-aware Letta child, the child waits on its parent's lock while this tracking makes the parent wait for the child.

src/agent/subagents/memory-worker.ts
```diff
@@ -0,0 +1,100 @@
...
+      },
+    );
+    if (result.status === "pushed") deps.onMemoryPushed?.();
+    return result;
+  };
+  return withMemoryOperation(
+    params.memoryDir,
+    async () => {
+      // Another worker may have repaired the checkout before this one acquired it.
+      if (params.repairOnly && (await sync()).status !== "conflict") {
```
amelia-letta, line 42: `sync()` can return `pushed` here when another actor resolved the conflict before this queued repair acquired the lock. This early return emits `memory_updated` but skips the prompt recompile below, so the parent conversation keeps the old memory despite the successful push.

src/tools/impl/memory-task-lifecycle.ts
```diff
@@ -0,0 +1,147 @@
...
+    if (pending.length === 0) return;
+    // One task's failure must not skip the teardown of the others.
+    await Promise.allSettled(pending);
+    for (const completion of pending) finished.add(completion);
+  }
+}
+
+/** Interactive exits do not wait for memory work: the child, its snapshot and its lock are torn down. */
+export async function cancelBackgroundMemoryTasks(): Promise<void> {
+  await finishBackgroundMemoryTasks(undefined, undefined, { cancel: true });
```
amelia-letta, line 137: Aborting the child does not roll back edits it made before cancellation; `runMemoryWorker` then sees `dirty`, releases the lock, and leaves a partially edited or staged checkout for the next session or worker. Interactive exit needs to drain the worker or restore its in-flight changes before releasing ownership.

src/agent/memory-conflict-repair.ts
```diff
@@ -0,0 +1,93 @@
...
+  const signature = await describeMemoryConflict(memoryDir, gitDir);
+  let previous: string | undefined;
+  try {
+    previous = (
+      JSON.parse(await readFile(path, "utf8")) as { signature?: string }
+    ).signature;
+  } catch {
+    /* No usable earlier attempt. */
+  }
+  if (previous === signature) return false;
```
amelia-letta, line 74: This treats a queued or running repair as already failed. Another conversation can reach post-turn sync after this marker is written but before the repair acquires the checkout lock, then surface a manual conflict even though automatic repair is still pending; track in-flight repair separately or only mark the signature after the worker finishes.

src/agent/memory-operation.ts
```diff
@@ -0,0 +1,158 @@
...
+    } catch {
+      // An unreadable owner file must not wedge every memory operation.
+      return null;
+    }
+  };
+  const isHeld = async (): Promise<boolean> => {
+    const owner = await readOwner();
+    if (!owner || !isAlive(owner.pid)) return false;
+    try {
+      return Date.now() - (await stat(path)).mtimeMs < STALE_MS;
```
amelia-letta, line 97: A live holder can miss this 60s window during process suspension or an event-loop stall, then resume after another process has reclaimed the file; without fencing, both operations will mutate the checkout concurrently. Stale recovery needs to distinguish PID reuse/death from a paused holder, or make the old holder stop once its token is replaced.

devanshrj:
Superseded by the split stack, all merged: #4625 (fork helpers), #4626 (checkout lock), #4627 (background memory worker), #4628 (conflict repair), #4634 (prompt policy).

PR #4541, opened 2026-09-18
feat(cli): add conversation Tray commands [LET-12857]

4shub:
## tl;dr

Tray gives a Cloud agent a small workspace that the user can see for the current conversation. Letta Code previously had no command or skill for that workspace. This PR adds JSON-only `letta tray add`, `list`, `update`, and `delete` commands for versioned markdownlets, plus a bundled skill that tells agents to update an existing item instead of creating duplicates.

The draft remains blocked on a live round trip against the matching deployed Cloud API and the human review required by this repository's AI policy.

## Review notes

The command resolves an agent/conversation pair atomically from explicit flags, harness environment variables, or the last session; it never combines IDs from different sources. One cached route-capability probe is shared by skill discovery and direct Skill loading, so managed localhost proxies expose Tray while self-hosted and local backends hide it. Mutation commands preserve Cloud resource errors and do not fetch the Tray first. The skill uses one-line, cross-shell commands with JSON payload files.

## Testing

The localhost-proxy regression test starts real Bun HTTP servers: a 200 route must expose `managing-tray` through both discovery and direct Skill loading, while a 404 route must hide and reject it. The same cached probe serves both paths. Focused command coverage also exercises payload files, complete scope examples, payload boundaries, unmodified Cloud 404 responses, and no mutation preflight.

A live add/list/update/delete round trip is not yet possible because the matching Cloud API has not been deployed. That blocks ready-for-review.

## Breadcrumbs

- Requested in: LET-12857
- Letta conversation: [`conv-d76f2efe-e159-4610-bbbe-1132b727c076`](https://chat.letta.com/chat/agent-ec93c28c-134a-4357-9388-4e7c6ed56bae?conversation=conv-d76f2efe-e159-4610-bbbe-1132b727c076)

## AI Disclosure

## AI Tool(s) Used

Letta Code researched the existing CLI patterns, implemented the command/API client/skill, addressed an independent agent review, and ran the listed validation. Human review is pending while this pull request remains a draft.

## Human Verification

Pending human review. The exact verification phrase is intentionally omitted until a human has reviewed and understands the change.


src/backend/api/tray-support.ts
```diff
@@ -0,0 +1,55 @@
...
+
+async function probeTrayRoute(agentId: string): Promise<boolean> {
+  const { baseUrl, apiKey } = await getApiRequestConfig();
+  const key = `${baseUrl}|${agentId}`;
+  const cached = supportCache.get(key);
+  if (cached && cached.expiresAt > Date.now()) return cached.value;
+
+  const value = (async () => {
+    try {
+      const response = await apiFetch(
```
amelia-letta, line 22: This probe is awaited on every turn, expires after 30 seconds, and has no timeout, so a slow or stalled localhost or self-hosted endpoint can delay or hang message submission even when Tray is never used.

PR #4549, opened 2026-09-18
feat(workflow): track run status and token usage

sarahwooders:
## Summary

Follow-up to #4531 (stacked on `sarah/workflow-lite`). Adds status tracking and token accounting to the Workflow tool without touching the DSL.

- **Token usage per subagent.** The spawner sums the Letta `usage_statistics` payloads the SDK forwards as `stream_event` messages (the SDK's result message carries cost but not tokens). The total rides on the outcome, the journal line, the progress event, and the run result. An early stop (timeout, runaway guard, abort) keeps what was already counted.
- **Execution registry** (`src/tools/workflow/execution-registry.ts`). In-process record per run: status, phases, agents (label, phase, status, duration, tokens), the last 50 log lines, and the token total. Fed by the engine's progress events; finished runs are retained for 5 minutes. Agents still queued/running when a run is stopped are marked `error · interrupted`.
- **`/workflows` command.** Plain-text tree per run, grouped by phase:

  ```
  ○ review — Review changed files
    1/3 agents done · 9s · 133.6k tokens
    Review (1/2)
      ✓ review:a.ts · 4s · 120k tokens
      ▶ review:b.ts
    Verify (0/1)
      ✗ verify:a.ts · 13.6k tokens · timed out
    » starting
    task workflow_1 · log /path/to/out.log
  ```
- **Completion notification.** Summary is now `Workflow "…" completed · 36s · 4 agents · 167.2k tokens` (or `failed after 12s`), with `total_tokens` in the usage block and `totalTokens` in the JSON payload.

## Deliberately not included

- No Ink status panel or `InputRich` changes (the flicker-prone part of #4179); `/workflows` and the existing background-task row are the surfaces.
- No listener/protocol snapshot changes.
- No USD cost. The SDK reports `totalCostUsd` on the result message but it goes missing on early stops; tokens are always countable, so they're the tracked number.

## Testing

- `bun run check`: 12/12.
- 59 tests across the workflow engine/spawner/registry/stats/display/tool files; 1,115 pass under `src/tools/`, `src/cli/helpers/`, `src/cli/commands/`.
- Not live-verified: the token sum depends on the SDK forwarding `usage_statistics` as `stream_event` with `total_tokens`, which #4179 verified against the real API but this branch has not re-run.


src/tools/workflow/workflow-engine.ts
```diff
@@ -113,6 +114,7 @@ export async function executeWorkflow(
         signal,
       );
       if (signal.aborted) throw new Error("Workflow aborted.");
+      totalTokens += outcome.totalTokens ?? 0;
```
amelia-letta, line 117: TaskStop aborts the signal before the spawner returns, so the preceding check throws before this captured token count is added or emitted. Every stopped run therefore reports 0 tokens, despite the early-stop accounting added here.

amelia-letta:
@cpacker Should the new process-global workflow registry be scoped to the execution context captured at launch, so `/workflows` cannot expose another conversation’s workflow metadata and output paths in a shared process?

PR #4538, opened 2026-09-18
fix(listener): register without a name prompt and open the sign-in page [LET-12841]

christinatong01:
## Summary

**Type:** Bug fix

**Goal:** `letta server` is the command the new chat.letta onboarding hands to a user to connect their computer ("Run one command" card). It should register the machine with as little friction as possible: paste, sign in, done.

**What was missing:** On a fresh machine with no saved credentials the command stopped twice before doing anything useful:

1. `PromptEnvName` rendered an interactive "Enter computer name (or press Enter for hostname)" prompt. The help text already says hostname is the default, and `--debug` already skipped the prompt.
2. `runListenerOAuthLogin` printed the device-auth URL and waited. The TUI login (`LettaLoginView`) opens the verification page in the browser; the listener path did not, so the user had to copy the URL out of the terminal.

**What changed:**

- `src/cli/subcommands/listen.tsx`: when no saved name and no `--computer-name`/`--env-name`, use `hostname()` and print `Registering this computer as "<name>" (use --computer-name to change it).` The `PromptEnvName` Ink component is removed; the name is still persisted via `settingsManager.setListenerEnvName` as before.
- `src/cli/subcommands/listen.tsx` (second commit): the first-run path renders a one-shot `FirstRunWelcome` banner (`AnimatedLogo` static frame + "Welcome to Letta" + the registering line) via Ink and unmounts it, leaving it in scrollback. Later runs (saved name) skip it.
- `src/websocket/listener/auth.ts`: `runListenerOAuthLogin` calls a new `openInBrowser` dep after printing the URL and code. Best-effort via dynamic `import("open")`, same pattern as `LettaLoginView`; failures are swallowed because the URL is still printed. Injected through `ListenerOAuthDeps` so tests never open a real browser.

## Before and after

**Before** (fresh `HOME`, `letta server`):

```
Enter computer name (or press Enter for hostname):
<Enter>
Log file: ~/.letta/logs/remote/....log
No API key found. Starting OAuth login...
To authenticate, visit: https://app.letta.com/oauth/device?user_code=XXXX-XXXX
Your code: XXXX-XXXX
Waiting for authorization...
```

**After:**

```
Registering this computer as "my-macbook" (use --computer-name to change it).
Log file: ~/.letta/logs/remote/....log
No API key found. Starting OAuth login...
To authenticate, visit: https://app.letta.com/oauth/device?user_code=XXXX-XXXX
Your code: XXXX-XXXX
Waiting for authorization...
```
[...]

PR #4554, opened 2026-09-18
fix(listener): point a first-run computer back at the browser once registered [LET-12841]

christinatong01:
## tl;dr

`letta server` is the command the chat.letta onboarding wizard hands a new user to connect their computer ("Run one command" card). After #4538 the first run registers without a prompt, shows a welcome banner, and opens the browser for device auth — but once registered, the `ListenerStatusUI` footer still told the user to "visit any agent and click the cloud button at the bottom left of the messenger input", and repeated the computer name the banner had just printed. A wizard user has no agent yet and is mid-flow in a browser tab that detects the listener on its own (`listConnections` poll), so that copy sent them looking for a button that does not apply to them. On a first run the footer now points them back at the browser and the duplicate name line is dropped; existing computers keep the current copy.

Stacked on #4538 (base is `christina/listener-onboarding-friction`) because it reuses the `showedFirstRunWelcome` flag introduced there.

## Before and after

First run, after registration completes (below the banner and sign-in output):

Before:

```
The name of your computer is: my-macbook

Awaiting instructions

Connect to this computer by visiting any agent and clicking the "cloud" button
at the bottom left of the messenger input and swapping your computer to my-macbook
```

After:

```
Awaiting instructions

Connected. Head back to Letta in your browser — it will detect this
computer automatically. Keep this window open.
```

Later runs (saved name, no banner) are unchanged: `isFirstRun` defaults to `false`, so the name line and the computer-picker instructions render as before.

## Testing

Ran the built `letta.js server` in a real terminal on a managed Linux sandbox desktop with an empty `HOME` and a valid API key, so it took the first-run path and reached the status UI — see Visual proof. Also ran it a second time with the same `HOME` (saved name) and confirmed the existing name line and picker copy render as before. No automated test is meaningful here: the change is two conditional copy branches on a boolean prop with no decision logic beyond the flag #4538 already sets.

## Visual proof

First-run status after registration

## Breadcrumbs

- Requested in: review of #4538 first-run output, LET-12841
- Letta conversation: [`conv-0c977403-32c4-437c-9340-03bcfeb8e062`](https://chat.letta.com/chat/agent-83173023-adfa-4ceb-903e-e12778950378?conversation=conv-0c977403-32c4-437c-9340-03bcfeb8e062)
- Linear: LET-12841
- Origin: the footer copy dates from the original `ListenerStatusUI` and was correct for its only caller at the time (a user connecting a computer to an existing agent). No single introducing pull request; the mismatch appears once #4538 makes `letta server` the onboarding entry point for users with no agent.
- Related: Follow-up to #4538


src/cli/subcommands/listen.tsx
```diff
@@ -865,6 +863,7 @@ export async function runListenSubcommand(argv: string[]): Promise<number> {
         <ListenerStatusUI
           connectionId={connectionId}
           envName={connectionName}
+          isFirstRun={showedFirstRunWelcome}
```
amelia-letta, line 866: `showedFirstRunWelcome` is true for every fresh config, including a user who runs `letta server` directly with an existing API key, so no onboarding tab may exist. That path now tells them to return to a browser that was never opened; this needs an onboarding-specific signal rather than using first registration as a proxy.

PR #4569, opened 2026-09-20
fix(slack): render the gateway model footer on harness-posted messages

overlord-letta:
## tl;dr

Slack messages posted by the harness itself — turn failure notices and direct sends — rendered a plain "View on web" footnote, while agent replies sent through the Cloud Slack integration render `<model> · Configure · View` for the same agent. The two footers appeared side by side in the same channel. This makes the harness footnote builder emit the same rich format when the channel gateway knows the runtime's model handle (which it already tracks per runtime), falling back to "View on web" when the model is unknown or the agent is local.

## Review notes

The Cloud Slack integration sends through this same sender and swaps the trailing context block for its own footer before posting. This change keeps a single trailing context block on every footnoted message, so that override keeps working unchanged.

## Testing

- `bun run check` — all 12 checks pass (biome, tsc, unit suites).
- New unit coverage: `src/channels/slack/presentation.test.ts` asserts the rich footer byte-for-byte against the format the Cloud Slack integration produces (provider-prefix stripping, mrkdwn escaping, fallback, local-agent omission); adapter and status-controller tests assert the rich footer on a direct send and on a fatal turn notice.
- Not exercised: the pairing-DM visual probe required by `src/channels/slack/AGENTS.md` for footnote changes. This sandbox has no Slack bot credentials, so the rendered footer has not been eyeballed in a live workspace on this commit.

## Breadcrumbs

- Requested in: internal bug-bash thread reporting that two agent messages in one Slack channel showed different footers
- Letta conversation: [conv-83c1f385](https://chat.letta.com/chat/agent-19babcc0-e958-41b0-81d8-00107f71deb7?conversation=conv-83c1f385-1123-4736-abee-010175d71f5f)
- Origin: not a regression. The harness footnote in `src/channels/slack/presentation.ts` and the Cloud integration's reply footer were written independently; no single introducing PR.

## AI Disclosure

## AI Tool(s) Used

Letta Code (Titan agent)

## Human Verification

Pending human review.


src/channels/gateway-local.ts
```diff
@@ -65,6 +66,23 @@ export interface LocalChannelGatewayHandle {
   ): Promise<ServiceCommandResponse>;
 }
 
+function enrichLifecycleEventWithModelHandle(
+  gateway: ChannelGateway,
+  event: ChannelTurnLifecycleEvent,
+): ChannelTurnLifecycleEvent {
+  const withModel = (source: ChannelTurnSource): ChannelTurnSource => ({
+    ...source,
+    modelHandle:
```
amelia-letta, line 75: Both new `getModelStatus` reads use a cache that is only refreshed at initial registration or through the channel `/model` handler, so Desktop or mod model changes leave later footers showing the old model.

amelia-letta:
@cpacker Should the fatal-error footer use the model snapshot captured when the turn started, rather than `gateway.getModelStatus()` at lifecycle completion? This determines whether a mid-turn model change can relabel the completed turn.

overlord-letta:
Closing in favor of a letta-cloud-side fix. Tracing the reported case further showed the plain-footer error notice was posted by the Cloud Slack gateway's own lifecycle-error path, which skips the footer option it applies to replies — so the fix belongs there, and duplicating the footer format in this repo would add the drift the request was meant to remove. This approach can be reopened if the self-hosted/Desktop direct-post case (letta-code's own adapter posting for cloud agents) needs the same treatment.

